PR #2, opened 2016-11-12
Throw OperationCanceledExceptions instead of ObjectDisposedExceptions

AArnott:
When an object is disposed of and the request is canceled, prefer throwing `OperationCanceledException` as callers tend to expect that.


AArnott:
@IlyaBiryukov: yes. Roslyn reported that they were seeing unexpected `ObjectDisposedException` being thrown. While that was their bug, it reminded me that in CPS when we're developing complex systems, we want this exception to mean there's a bug. And cancelling a request and *then* disposing the instance whose request hasn't been responded to was preferred to throw `OperationCanceledException` than `ObjectDisposedException` to help us recognize that this was an expected exception and not the sign of a bug.

PR #18, opened 2017-03-14
Changed header message handler to be more resilient to extra spaces or lack of spaces.

tinaschrepfer:


tinaschrepfer:
@AArnott That's odd.  When I run the tests on my machine it all passed, including the one I just added.  Is it possible that the PR build is using some stale bits?

AArnott:
Most likely your test depends on being checked out with CRLF line endings, which AppVeyor for some reason won't do. So you should recode your test to explicitly encode its line endings in your multiline strings.

PR #19, opened 2017-03-15
Add support for sending named parameter objects

tinaschrepfer:
Common language server specifies that the parameter passed in the message must be a single object, not an array of one object.  Talked to Ilya about the design to support this in StreamJsonRpc and not be a breaking change.  This PR now supports sending the parameter over as an object.


tinaschrepfer:
Hold off on this review.  After talking to Dirk from VS Code, it looks like all calls using the common language server protocol needs to have only one parameter and the parameter passed through as object.  I'm going to make that change instead of this one, which determines how the parameter should be passed on a per-call basis.

src/StreamJsonRpc.Desktop/MultilingualResources/StreamJsonRpc.Desktop.cs.xlf
```diff
@@ -148,6 +148,10 @@
           <source>This operation is not allowed before listening for messages has started.</source>
           <target state="translated">Před zahájením naslouchání zprávám není tato operace povolená.</target>
         </trans-unit>
+        <trans-unit id="ParameterNotObject" translate="yes" xml:space="preserve">
+          <source>Parameter is not in the form of a single object</source>
+          <target state="new">Parameter is not in the form of a single object</target>
```
tinaschrepfer: I'm not sure what I need to do to localize the new resource I added.  It's in English everywhere.  :(

AArnott: I guess you have the MAT extension installed in your VS, is that right? I would have expected it to make this change. Translation comes later. What you have should be fine, so long as it's what MAT did.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -80,12 +85,13 @@ internal OutstandingCallData(object taskCompletionSource, Action<JsonRpcMessage>
         /// </summary>
         /// <param name="stream">A bidirectional stream to send and receive RPC messages on.</param>
         /// <param name="target">An optional target object to invoke when incoming RPC requests arrive.</param>
+        /// <param name="passParameterAsObject">Value which indicates whether there can be only one parameter and the parameter must be passed as an object</param>
         /// <returns>The initialized and listening <see cref="JsonRpc"/> object.</returns>
-        public static JsonRpc Attach(Stream stream, object target = null)
+        public static JsonRpc Attach(Stream stream, object target = null, bool passParameterAsObject = false)
```
AArnott: I've created #20 to guard against accidental breaks like this.

AArnott: Rather than `passParameterAsObject`, can we use the name "useNamedParameters"?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -439,7 +449,22 @@ protected virtual void Dispose(bool disposing)
             cancellationToken.ThrowIfCancellationRequested();
 
             arguments = arguments ?? EmptyObjectArray;
-            JsonRpcMessage request = JsonRpcMessage.CreateRequest(id, targetName, arguments, this.JsonSerializer);
+
+            JsonRpcMessage request;
+            if (passParameterAsObject)
```
AArnott: Please prefix all instance member access with `this.`

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1,16 +1,15 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
+﻿using Microsoft;
+using Microsoft.VisualStudio.Threading;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using StreamJsonRpc;
+using System;
```
AArnott: Please keep System usings on top.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -526,6 +525,49 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
...
+        Assert.Equal(message.Parameters.Type, JTokenType.Array);
+    }
+
+    [Fact]
+    public void InvokeAsync_ProcessParameterSentAsObject()
+    {
+        var sending = new MemoryStream();
+
+        var rpc = JsonRpc.Attach(sendingStream: sending, receivingStream: null, target: null, passParameterAsObject: true);
+        rpc.NotifyAsync("test", new Foo() { Bar = "test" }).Wait();
```
AArnott: Can you please add a test that uses an anonymous type? (e.g. `new { foo = 3, bar = "3" }`)

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -526,6 +525,49 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
...
+
+        sending.Position = 0;
+        var handler = new HeaderDelimitedMessageHandler(null, sending);
+        var json = handler.ReadAsync(default(CancellationToken)).GetAwaiter().GetResult();
+
+        var message = JsonRpcMessage.FromJson(json, new JsonSerializerSettings());
+        Assert.Equal(message.Parameters.Type, JTokenType.Object);
```
AArnott: Please add assertions for the expected fields on that object.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: FYI I completed that PR and re-built your PR in appveyor. It now registers as a build failure due to the breaking API changes.

tinaschrepfer:
@AArnott , what should I do to add the new methods to the public API so the appveyor build passes?

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
```diff
@@ -120,6 +127,13 @@ public object[] GetParameters(ParameterInfo[] parameterInfos, JsonSerializer jso
                 return this.Parameters.ToObject<object[]>();
             }
 
+            if (parameterInfos.Length == 1 && parameterInfos[0].ParameterType == typeof(JToken))
+            {
+                // We want to ignore methods that only have one parameter and the parameter type is JToken.
+                // This is reserved to be called when the parameter is passed as an object.
+                return null;
```
AArnott: @ilyabiryukov Is this a breaking change? Do methods today have the option of accepting a parameter as a `JToken`? If not, then I guess this change might be OK, but we may want to revisit in the future since accepting a JToken as a parameter even when the parameters came in as an array sounds like a reasonable scenario to support. So as long as we could support that going forward without a breaking change, I'm OK with this.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -145,7 +145,7 @@ public JsonRpc(Stream sendingStream, Stream receivingStream, object target = nul
         public JsonRpc(DelimitedMessageHandler messageHandler, object target = null)
         {
             Requires.NotNull(messageHandler, nameof(messageHandler));
-
+
```
AArnott: Could you revert the trailing whitespace change, please?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -320,6 +346,37 @@ public Task InvokeWithCancellationAsync(string targetName, IReadOnlyList<object>
...
+        /// </exception>
+        /// <exception cref="RemoteMethodNotFoundException">
+        /// Result task fails with this exception if the <paramref name="targetName"/> method is not found on the target object on the server.
+        /// </exception>
+        /// <exception cref="RemoteTargetNotSetException">
+        /// Result task fails with this exception if the server has no target object.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">If <paramref name="targetName"/> is null.</exception>
+        /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
+        public Task<Result> InvokeWithParameterPassedAsObjectWithCancellationAsync<Result>(string targetName, object argument = null, CancellationToken cancellationToken = default(CancellationToken))
```
AArnott: I don't think we need the two methods (one with CancellationToken and one without). Let's just have one (that takes an optional CancellationToken), and the shorter method name.

With the parameters array methods, we had to have it because the args came in as a params array, so we couldn't accept a CancellationToken parameter after that. But here, that's not a restriction.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -362,8 +418,34 @@ public Task<Result> InvokeWithCancellationAsync<Result>(string targetName, IRead
         /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
         public Task NotifyAsync(string targetName, params object[] arguments)
         {
+            // If somebody calls InvokeInternal<T>(id, "method", null), the null is not passed as an item in the array.
+            // Instead, the compiler thinks that the null is the array itself and it'll pass null directly.
+            // To account for this case, we check for null below.
+            arguments = arguments ?? new object[] { null };
```
AArnott: This caters to the way C# in particular works. We're hacking in functionality to reverse something the compiler did, which in general would make it more difficult to predict what will happen when you call a params method. So I'm not a fan of 'fixing up' here. Rather, if someone wants to pass in null as a single argument, they can always use the syntax `new object[] { null }`.
Or if we want to make it Just Work by default, the proper way is to add an overload of the NotifyAsync method that accepts exactly one argument. That way, when `null` is the one argument, C# will pick that overload instead of the params one.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -292,6 +292,32 @@ public Task<Result> InvokeAsync<Result>(string targetName, params object[] argum
...
+        /// </exception>
+        /// <exception cref="RemoteMethodNotFoundException">
+        /// Result task fails with this exception if the <paramref name="targetName"/> method is not found on the target object on the server.
+        /// </exception>
+        /// <exception cref="RemoteTargetNotSetException">
+        /// Result task fails with this exception if the server has no target object.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">If <paramref name="targetName"/> is null.</exception>
+        /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
+        public Task<Result> InvokeWithParameterPassedAsObjectAsync<Result>(string targetName, object argument = null)
```
AArnott: The json-rpc spec refers to the two parameter modes as "by-name" and "by-position". I suggest we shorten this method name to `InvokeWithNamedParametersAsync` to better reflect the spec's naming.

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
```diff
@@ -67,7 +67,14 @@ public static JsonRpcMessage CreateRequest(int? id, string @method, IReadOnlyLis
 
         public static JsonRpcMessage CreateRequestWithNamedParameters(int? id, string @method, object namedParameters, JsonSerializer parameterSerializer)
         {
-            return new JsonRpcMessage(method, JToken.FromObject(namedParameters, parameterSerializer), id);
+            if (namedParameters == null)
```
AArnott: Is this legal? The json-rpc 2.0 spec reads: 

> by-name: params MUST be an Object

No allowance is made for a null value here. So it seems to me that we should pass `{}` when no parameters are specified, or (perhaps better) throw an exception so folks realize they must pass in an empty object, since that's what the remote end will get.

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
```diff
@@ -8,6 +8,7 @@
 using StreamJsonRpc;
 using Xunit;
 using Xunit.Abstractions;
+using Newtonsoft.Json.Linq;
```
AArnott: sort usings

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
```diff
@@ -44,6 +45,82 @@ public JsonRpcClientInteropTests(ITestOutputHelper logger)
     }
 
     [Fact]
+    public async Task NotifyAsync_ParameterObjectSentAsArray()
+    {
+        Task notifyTask = this.clientRpc.NotifyAsync("test", new  { Bar = "test" });
+        JObject request = await this.ReceiveAsync();
+        Assert.Equal<JTokenType>(JTokenType.Array, request["params"].Type);
+        Assert.Equal<string>("test", ((JArray)request["params"])[0]["Bar"].ToString());
```
AArnott: You don't really need to specify all the generic type arguments for these `Assert.Equal` calls, do you?

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
```diff
@@ -44,6 +45,82 @@ public JsonRpcClientInteropTests(ITestOutputHelper logger)
...
+    }
+
+    [Fact]
+    public async Task NotifyAsync_ParameterNullSentAsArray()
```
AArnott: Your test method name suggests you're passing in a null parameter, but you're not passing in any parameters.

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
```diff
@@ -44,6 +45,82 @@ public JsonRpcClientInteropTests(ITestOutputHelper logger)
...
+        Assert.Equal<JTokenType>(JTokenType.Object, request["params"].Type);
+        Assert.Equal<string>("test", request["params"]["Bar"].ToString());
+    }
+
+    [Fact]
+    public async Task InvokeWithParameterPassedAsObjectAsync_NoParameter()
+    {
+        Task notifyTask = this.clientRpc.InvokeWithParameterPassedAsObjectAsync<object>("test");
+        JObject request = await this.ReceiveAsync();
+        Assert.Null(request["params"]);
```
AArnott: Similar to my comment to the product code, I think this is incorrect. We pass empty arrays for methods with no parameters, and the spec says we should pass in an object in this mode (without exception given for no-arg method calls).

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -563,6 +590,21 @@ public static string ServerMethod(string argument)
...
         }
 
+        public static string TestParameter(JToken token)
+        {
+            return "object";
+        }
+
+        public static string TestParameter(string argument)
+        {
+            return "array";
```
AArnott: I think returning the concatenation of this constant + argument would make for a better assertion in the test that communication succeeded.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -526,6 +525,34 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
...
+    {
+        string result1 = await this.clientRpc.InvokeAsync<string>(nameof(Server.TestParameter), "test");
+        Assert.Equal("array", result1);
+    }
+
+    [Fact]
+    public async Task CanInvokeServerMethodWithNoParameterPassedAsObject()
+    {
+        string result1 = await this.clientRpc.InvokeWithParameterPassedAsObjectAsync<string>(nameof(Server.TestParameter));
+        Assert.Equal("object or array", result1);
```
AArnott: I'm not sure I buy this -- until our support for *receiving* by-name messages includes finding method overloads with matching parameters. Until then, I think the only thing we should be willing to invoke is the one method that accepts `JToken`. So by that reasoning, I would expect `result1 == "object"`

AArnott:
@tinaschrepfer The build should be broken for you locally as well (assuming you've merged your topic branch with the tip of your target branch and re-run init.ps1). 
Then you see the build warnings/errors in the Error List and can use the lightbulb to resolve it using Roslyn's code fix provider to automatically update the PublicAPI.Unshipped.txt file.

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
IlyaBiryukov: That is a breaking change. We shouldn't ignore methods with only one JToken parameters because they can be called passing the parameter value in array. We should be able to call such methods with existing code.

src/StreamJsonRpc.Shared/JsonRpc.cs
IlyaBiryukov: How about ```InvokeWithParameterObjectAsync``` ?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -362,8 +418,34 @@ public Task<Result> InvokeWithCancellationAsync<Result>(string targetName, IRead
...
+        /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
+        public Task NotifyWithParameterAsObjectAsync(string targetName, object argument = null)
+        {
+            // If argument is null, this indicates that the method does not take any parameters.
+            object[] argumentToPass = argument == null ? null : new object[] { argument };
+
             int? id = null;
-            return this.InvokeCoreAsync<object>(id, targetName, arguments, CancellationToken.None);
+            
+            return this.InvokeCoreAsync<object>(id, targetName, argumentToPass, CancellationToken.None, true);
```
IlyaBiryukov: Please add the parameter name

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -430,16 +543,38 @@ protected virtual void Dispose(bool disposing)
...
 
-            arguments = arguments ?? EmptyObjectArray;
-            JsonRpcMessage request = JsonRpcMessage.CreateRequest(id, targetName, arguments, this.JsonSerializer);
+            JsonRpcMessage request;
+            if (isParameterPassedAsObject)
+            {
+                object argument = arguments;
+                if (argument != null)
+                {
+                    if (arguments.Count != 1 || arguments[0] == null || arguments[0].GetType().IsArray)
```
IlyaBiryukov: How about ```!arguments[0].GetType().IsClass``` ?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -430,16 +543,38 @@ protected virtual void Dispose(bool disposing)
         /// <param name="targetName">Name of the method to invoke.</param>
         /// <param name="arguments">Arguments to pass to the invoked method. If null, no arguments are passed.</param>
         /// <param name="cancellationToken">The token whose cancellation should signal the server to stop processing this request.</param>
+        /// <param name="isParameterPassedAsObject">Value which indicates if parameter should be passed as an object.</param>
```
IlyaBiryukov: how about ```isParameterObject```?

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
AArnott: I guess we may be missing a test for JToken parameters then. Ideally, any breaking change would cause a unit test failure.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Either one works for me.

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
tinaschrepfer: By passing in null, the serializer omits the parameter altogether, which is allowed according to the spec.  From the common language server protocol, it looks like there can be requests where the parameter is "void".  I tested not passing in any params (passing null in streamjsonprc api), or passing an empty array, against the Java language server.  It seemed to have accepted both.  However, wouldn't we want to support calling a method with no "params" at all?

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
tinaschrepfer: So if I leave this out I get an exception in the following case: Server method have 2 methods that both take only one argument.  One method takes JToken, the other takes string.  What ends up happening is both of them are matched when InvokeAsync is called and the code ends up throwing an exception due to multiple methods found.  Is this an existing bug then?

src/StreamJsonRpc.Shared/JsonRpc.cs
tinaschrepfer: So this line was in InvokeAsync to begin with.  I only added it after adding my unit tests and one of them failed, then I noticed that this line was only present in InvokeAsync and not NotifyAsync.  Talked to Ilya and we thought it was just because this line was accidentally left out of the NotifyAsync?  If we want to remove it, it should be done in both places.

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
tinaschrepfer: Talked to Ilya about this and it looks like that's just the contract for using streamsjsonrpc.  The server shouldn't have 2 methods both with one argument and one of them is JToken.  JToken is used as a catch all.  I'll modify my unit test to not break the contract and also have a unit test in place to verify the contract.

src/StreamJsonRpc.Shared/DataContracts/JsonRpcMessage.cs
AArnott: Good point on the spec allowing the parameter to be omitted. I thought you were supplying it with a null value with this change and I didn't know it could be omitted entirely.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Rats. I'd made this comment before when we were developing the library but I guess it didn't get resolved.

Let's not propagate the pattern to `NotifyAsync`. Instead, let's apply the right pattern to NotifyAsync (with the 1 arg overload) and add the 1 arg overload to InvokeAsync as well. We can remove the offending code from InvokeAsync at a point where we feel good about breaking changes (e.g. 2.0).

src/StreamJsonRpc.Shared/JsonRpc.cs
tinaschrepfer: @AArnott Not sure what's going on, but I just can't see what whitespace you're referring to?

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
tinaschrepfer: I guess not.

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
tinaschrepfer: Changed.

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
tinaschrepfer: See previous reply about omitting params when null is set.  This test would verify that case.

tinaschrepfer:
@AArnott I see the warning for public API in error list, but light bulb shows me now options to fix it.  The only thing it shows is "change signature".  Am I missing an extension or something?

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: You don't see the green band of spaces you added to line 148 in the diff?

src/StreamJsonRpc.Desktop/PublicAPI.Shipped.txt
```diff
@@ -67,4 +67,4 @@ static StreamJsonRpc.JsonRpc.Attach(System.IO.Stream sendingStream, System.IO.St
 static StreamJsonRpc.JsonRpc.Attach(System.IO.Stream stream, object target = null) -> StreamJsonRpc.JsonRpc
 virtual StreamJsonRpc.DelimitedMessageHandler.Dispose(bool disposing) -> void
 virtual StreamJsonRpc.JsonRpc.Dispose(bool disposing) -> void
-virtual StreamJsonRpc.JsonRpc.InvokeCoreAsync<ReturnType>(int? id, string targetName, System.Collections.Generic.IReadOnlyList<object> arguments, System.Threading.CancellationToken cancellationToken) -> System.Threading.Tasks.Task<ReturnType>
```
AArnott: Removal of this line signifies a breaking change.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -239,6 +239,31 @@ public void StartListening()
...
+        /// Result task fails with this exception if the <paramref name="targetName"/> method is not found on the target object on the server.
+        /// </exception>
+        /// <exception cref="RemoteTargetNotSetException">
+        /// Result task fails with this exception if the server has no target object.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">If <paramref name="targetName"/> is null.</exception>
+        /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
+        public Task InvokeAsync(string targetName, object argument)
+        {
+            return this.InvokeAsync<object>(targetName, argument);
```
AArnott: Thanks for adding this overload. However this one may be missing a test, as it would throw `StackOverflowException`. You'll need to explicitly surround `argument` with `new object[] { argument }` in order to force the compiler to call the other overload.

src/StreamJsonRpc.Desktop/PublicAPI.Shipped.txt
tinaschrepfer: This is actually an edit, since I had to add the "isParameterObject" parameter (see next change).  Are you saying I shouldn't break this API?

src/StreamJsonRpc.Shared/JsonRpc.cs
tinaschrepfer: @AArnott I saw a mail with your comment, but can't find it here - ". However this one may be missing a test, as it would throw StackOverflowException. You'll need to explicitly surround argument with new object[] { argument } in order to force the compiler to call the other overload." 

This method actually calls the other overload I added InvokeAsync<T>, and that overload does explicitly call with new object[] { argument }, so I think we're okay here.

src/StreamJsonRpc.Desktop/PublicAPI.Shipped.txt
AArnott: An edit is equivalent to a delete. You're changing a public API and that's a breaking change. Code that compiled against the old version in IL indicates that it's invoking a method with *n* arguments and you've removed that overload and put one with *n+1* arguments in its place, which will cause the CLR to throw `MethodNotFoundException` for. So you can't edit existing public APIs.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: I agree. You don't see it here because I deleted it after I realized my mistake. :)

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -145,7 +145,7 @@ public JsonRpc(Stream sendingStream, Stream receivingStream, object target = nul
         public JsonRpc(DelimitedMessageHandler messageHandler, object target = null)
         {
             Requires.NotNull(messageHandler, nameof(messageHandler));
-
+
```
AArnott: Please execute Format Document on this file so that this trailing whitespace (and any other issues) are resolved.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -292,6 +345,35 @@ public Task<Result> InvokeAsync<Result>(string targetName, params object[] argum
...
+        /// <exception cref="RemoteTargetNotSetException">
+        /// Result task fails with this exception if the server has no target object.
+        /// </exception>
+        /// <exception cref="ArgumentNullException">If <paramref name="targetName"/> is null.</exception>
+        /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
+        public Task<Result> InvokeWithParameterObjectAsync<Result>(string targetName, object argument = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            // If argument is null, this indicates that the method does not take any parameters.
+            object[] argumentToPass = argument == null ? null : new object[] { argument };
+            return this.InvokeWithCancellationCoreAsync<Result>(targetName, argumentToPass, cancellationToken, true);
```
AArnott: Can you name the last arg? (`passParameterAsObject: `)

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
```diff
@@ -1,11 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
-using System.Text;
+﻿using StreamJsonRpc;
```
AArnott: Please sort with `System` usings on top.

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
```diff
@@ -44,6 +41,82 @@ public JsonRpcClientInteropTests(ITestOutputHelper logger)
...
+        Task notifyTask = this.clientRpc.InvokeAsync<object>("test");
+        JObject request = await this.ReceiveAsync();
+        Assert.Equal(JTokenType.Array, request["params"].Type);
+        Assert.Equal(0, ((JArray)request["params"]).Count);
+    }
+
+    [Fact]
+    public async Task InvokeWithParameterPassedAsObjectAsync_ParameterObjectSentAsObject()
+    {
+        Task notifyTask = this.clientRpc.InvokeWithParameterObjectAsync<object>("test", new { Bar = "test" });
```
AArnott: tip for next time: use different values for the various strings so that you detect bugs in your assertions where the values get mixed up in the product. e.g. if the method name and parameter got mixed up, they'd still both be "test" and your test might pass.

AArnott:
This looks good. Thank you, @tinaschrepfer for responding to our requests for changes. I do have a few pending comments (most particularly the breaking API change fix that must be reverted) but after that I think we can merge.

src/StreamJsonRpc.Desktop/PublicAPI.Shipped.txt
tinaschrepfer: Okay, I've added another overload method so the existing API isn't broken.

src/StreamJsonRpc.Tests/JsonRpcClientInteropTests.cs
tinaschrepfer: Cool, didn't think about that.  Changed.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -433,13 +585,50 @@ protected virtual void Dispose(bool disposing)
         /// <returns>A task whose result is the deserialized response from the JSON-RPC server.</returns>
         protected virtual async Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string targetName, IReadOnlyList<object> arguments, CancellationToken cancellationToken)
         {
+            return await InvokeCoreAsync<ReturnType>(id, targetName, arguments, cancellationToken, isParameterObject: false).ConfigureAwait(false);
```
AArnott: optimization tip: A thunk method like this doesn't need to be async. If you remove the `async` keyword from the method signature, the `await` keyword from this expression, and the `.ConfigureAwait(false)` from the end of the expression, you'll eliminate the state machine that the compiler and runtime will have to create for an async method.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -345,8 +427,26 @@ public Task InvokeWithCancellationAsync(string targetName, IReadOnlyList<object>
         /// <exception cref="ObjectDisposedException">If this instance of <see cref="JsonRpc"/> has been disposed.</exception>
         public Task<Result> InvokeWithCancellationAsync<Result>(string targetName, IReadOnlyList<object> arguments = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            int id = Interlocked.Increment(ref this.nextId);
-            return InvokeCoreAsync<Result>(id, targetName, arguments, cancellationToken);
+            return InvokeWithCancellationCoreAsync<Result>(targetName, arguments, cancellationToken, isParameterObject: false);
```
AArnott: I'm a little concerned that derived types that override the original `InvokeCoreAsync<T>` method will find that in code that doesn't leverage your new functionality their override is no longer invoked. Can we rearrange these thunks so that the param array style still calls through the original virtual method and only the new param object style goes directly to the new one?  Our own implementation of the old virtual method can still simply forward the call to the new method, but this way an override could still take effect.

tinaschrepfer: Good point.  I've removed the extra InvokeWithCancellationCoreAsync method and just have the individual methods keep track of incrementing ID and calling the right InvokeCoreAsync method.

PR #20, opened 2017-03-16
Add roslyn analyzer to guard public API from breaks

AArnott:
Along with the analyzer, we add the PublicAPI.Shipped.txt file to reflect the API we're committed to.


src/StreamJsonRpc.Desktop/StreamJsonRpc.Desktop.ruleset
```diff
@@ -74,4 +74,8 @@
     <Rule Id="UseAsyncSuffix" Action="Error" />
     <Rule Id="AvoidAsyncSuffix" Action="Error" />
   </Rules>
+  <Rules AnalyzerId="Roslyn.Diagnostics.Analyzers" RuleNamespace="Roslyn.Diagnostics.Analyzers">
+    <Rule Id="RS0017" Action="Error" />
+    <Rule Id="RS0026" Action="Info" />
```
kirankulkarni747: why are overidding these to be not warnings?

AArnott: RS0017 represents API breaks. They should not be so easily ignored as the warnings we already have.
RS0026 is a good idea (avoid multiple overloads that take optional parameters) but it's too late for that. We already violate it, so again, to avoid warnings for existing code, I'm reducing it to info for now.

PR #21, opened 2017-03-20
added threading analyzer

javierdlg:


javierdlg:
That is correct. Andrew had already implemented the analyzer in most of the projects. Only this one was missing.

AArnott:
This is not necessary. project.json inherits dependencies over P2P references. So the line you added I strongly suspect is a no-op. That's why it wasn't in there to begin with.

PR #24, opened 2017-03-28
Add support for providing alternate name for methods

tinaschrepfer:
Common language server protocol's method names contain "/", which aren't valid method names in C#.  Add support for alternate names to enable this scenario.


src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Microsoft;
+
+namespace StreamJsonRpc
+{
+    /// <summary>
+    /// Attribute which allows a method to have an alternate name.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
```
AArnott: Given that `AllowMultiple=false` why is the string resource you added necessary that is prepared to report on multiple of these attributes being applied?

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// Ctor.
+        /// </summary>
+        /// <param name="name">Alternate name of a method.</param>
+        public MethodNameAttribute(string name)
+        {
+            Requires.NotNullOrWhiteSpace(name, nameof(name));
+
+            Name = name;
```
AArnott: Please prefix all instance member use with `this.`

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace StreamJsonRpc
+{
+    /// <summary>
+    /// Attribute which allows a method to have an alternate name.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    public class MethodNameAttribute : Attribute
+    {
+        /// <summary>
+        /// Ctor.
```
AArnott: `Initializes a new instance of the <see cref="MethodNameAttribute" /> class.`

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        /// <param name="name">Alternate name of a method.</param>
```
AArnott: Can you describe what the significance of the alternate name is, both here and in the type's summary?
Specifically the details I was looking for but didn't see was whether this *replaces* the remote name of the method vs. adds another option.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -654,6 +661,38 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
                 }
             }
         }
+        
+        private static Dictionary<string, List<string>> GetMethodToAttributeMap(object target)
+        {
+            var methodToAttributeMap = new Dictionary<string, List<string>>();
```
AArnott: Can you specify a `StringComparer` argument to the dictionary's ctor? Even if it matches what would be the default behavior, it's good to know that it was considered.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -654,6 +661,38 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+            if (target != null)
+            {
+                for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
+                {
+                    var anyMethodWithAttribute = t.DeclaredMethods.Any(m => m.GetCustomAttribute(typeof(MethodNameAttribute)) != null);
+                    if (anyMethodWithAttribute)
```
AArnott: Your check before enumerating seems like it results in scanning the type twice (worst case scenario). I don't see what benefit the up front check is though. Why not drop it and just have the foreach loop?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -654,6 +661,38 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+                    {
+                        foreach (MethodInfo method in t.DeclaredMethods)
+                        {
+                            var attribute = method.GetCustomAttribute(typeof(MethodNameAttribute)) as MethodNameAttribute;
```
AArnott: nit: please [use the `()` cast operator](http://blog.nerdbank.net/2008/06/when-not-to-use-c-keyword.html) instead of `as` unless you expect the cast to fail in some legit circumstances. That way, we'll get more actionable exceptions if it were to ever go wrong.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -654,6 +661,38 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
                 }
             }
         }
+        
+        private static Dictionary<string, List<string>> GetMethodToAttributeMap(object target)
```
AArnott: Can you add xml doc comments to this method, including a thorough description of what the returned dictionary contains?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -654,6 +661,38 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+        {
+            var methodToAttributeMap = new Dictionary<string, List<string>>();
+            if (target != null)
+            {
+                for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
```
AArnott: I like how you've prepared support for methods defined on base types. 💯

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -654,6 +661,38 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+                            {
+                                if (!methodToAttributeMap.ContainsKey(method.Name))
+                                {
+                                    methodToAttributeMap.Add(method.Name, new List<string>());
+                                }
+                                if (!methodToAttributeMap[method.Name].Contains(attribute.Name))
+                                {
+                                    methodToAttributeMap[method.Name].Add(attribute.Name);
+                                }
+                            }
```
AArnott: I'm not sure how critical performance will be for this method. But given we seek to be an perf-invisible interop layer, this block could be rewritten to have minimal dictionary lookups and key hashing:

```csharp
List<string> names;
if (!methodToAttributeMap.TryGetValue(method.Name, out names))
{
  names = new List<string>();
  if (!names.Contains(attribute.Name, StringComparer.Ordinal))
  {
    methodToAttributeMap.Add(attribute.Name, names);
  }
}

names.Add(attribute.Name);
```

I also threw in use of `StringComparer.Ordinal` in the example to state that we intentionally want a case sensitive search (which I *believe* we do).

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
```diff
@@ -28,12 +32,43 @@ internal TargetMethod(JsonRpcMessage request, object target, JsonSerializer json
...
+            bool errorFound = false;
             for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
             {
-                bool matchFound = false;
-                foreach (MethodInfo method in t.GetDeclaredMethods(request.Method))
+                var matchFound = false;
+                var methodName = request.Method;
+                var methodAttribute = request.Method;
+
+                if (methodToAttributeMap != null)
```
AArnott: Do we really need to support the null case? Are there paths where this is legitimately null? If not, I'd prefer we `Requires.NotNull` at the top and lose the if check so we have fewer branches in our code to worry about.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
```diff
@@ -28,12 +32,43 @@ internal TargetMethod(JsonRpcMessage request, object target, JsonSerializer json
...
+                {
+                    foreach (string methodKey in methodToAttributeMap.Keys)
```
AArnott: You're looking up the value for this key a few times in this loop. That's extra dictionary CPU time that you don't need if you enumerate the dictionary instead of its Keys collection.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
```diff
@@ -28,12 +32,43 @@ internal TargetMethod(JsonRpcMessage request, object target, JsonSerializer json
...
+                    {
+                        if (methodToAttributeMap[methodKey].Any(a => string.Equals(a, request.Method, StringComparison.OrdinalIgnoreCase)))
+                        {
+                            if (methodToAttributeMap[methodKey].Count != 1)
+                            {
+                                errorFound = true;
+                                this.errorMessages.Add(string.Format(CultureInfo.CurrentCulture, Resources.ConflictingMethodNameAttribute, methodKey));
```
AArnott: Why are we waiting till a message comes in for a particular method before reporting an error, where folks might not notice it if they didn't happen to test that method? Why not force it to be noticed as soon as the bug is introduced by throwing when first reflecting over the type?

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
```diff
@@ -28,12 +32,43 @@ internal TargetMethod(JsonRpcMessage request, object target, JsonSerializer json
...
-                foreach (MethodInfo method in t.GetDeclaredMethods(request.Method))
+                var matchFound = false;
+                var methodName = request.Method;
+                var methodAttribute = request.Method;
+
+                if (methodToAttributeMap != null)
+                {
+                    foreach (string methodKey in methodToAttributeMap.Keys)
+                    {
+                        if (methodToAttributeMap[methodKey].Any(a => string.Equals(a, request.Method, StringComparison.OrdinalIgnoreCase)))
```
AArnott: It feels like we're using the dictionary backwards, IMO. All access to it is driven by incoming messages which will contain the *value* of the dictionary entry instead of the key. But imagine if we swapped key/value in this dictionary: then we could do a constant time lookup to see if there is a substitution we should apply to the incoming method name.

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
AArnott: Also: what if there are multiple overloads of the same method in the .NET source code, yet they don't all have exactly the same `[MethodName]` attribute argument? What happens then? Can this be documented on this attribute for discoverability?

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
```diff
@@ -173,12 +209,12 @@ private static object[] TryGetParameters(JsonRpcMessage request, MethodSignature
             }
         }
 
-        private bool TryAddMethod(JsonRpcMessage request, Dictionary<MethodSignature, object[]> targetMethods, MethodInfo method, JsonSerializer jsonSerializer)
+        private bool TryAddMethod(JsonRpcMessage request, Dictionary<MethodSignature, object[]> targetMethods, MethodInfo method, JsonSerializer jsonSerializer, string methodAttribute = null)
```
AArnott: The `methodAttribute` parameter name sounds like it's an attribute. But it's only a string -- the value of a property from the attribute. Can we give this a more descriptive name like `requestedMethodName` or `clrMethodName`?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -620,6 +692,31 @@ public static string TestInvalidMethod(JToken test)
 
         public new string RedeclaredBaseMethod() => "child";
 
+        [MethodName("child/InvokeVirtualMethodOverride")]
+        public override string InvokeVirtualMethodOverride() => $"child {nameof(InvokeVirtualMethodOverride)}";
+
+        [MethodName("test/InvokeTestMethod")]
+        public string InvokeTestMethodAttribute() => "test method attribute";
+
+        public override string InvokeVirtualMethodNoOverride() => $"child {nameof(InvokeVirtualMethodNoOverride)}";
```
AArnott: From a usability standpoint, I'm having a hard time deciding whether I would predict the `[MethodName]` attribute to be inherited from the base class or not. whichever way you're going, can you document it on the attribute class?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -559,6 +559,51 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
...
+        Assert.Equal("test method attribute", correctCasingResult);
+
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync<string>("teST/InvokeTestmeTHod"));
+
+        string baseMethodResult = await this.clientRpc.InvokeAsync<string>("base/InvokeMethodWithAttribute");
+        Assert.Equal("base InvokeMethodWithAttribute", baseMethodResult);
+
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync<string>("base/InvokeVirtualMethodOverride"));
+
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync<string>("child/InvokeVirtualMethodOverride"));
```
AArnott: Can you add comments justifying why you expect this to fail? The child method appears to explicitly claim this is the method name, so I don't know why it wouldn't be found except perhaps that it conflicts with the base method.

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
tinaschrepfer: I tested the behavior of this, even if I set AllowMultiple to false the overridden method can still specify its own MethodName attribute with a different name value.  Not sure what the behavior should be for setting AllowMultiple.  Hence the extra check and unit tests.

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
tinaschrepfer: It's meant to replace, so alternate name was a bit misleading.  I'll change that.  If there are overloads with different MethodNameAttribute values, then it would result in an error.  I'll be more clear in the description.  Good case on the overload scenario, I've added a test.

src/StreamJsonRpc.Shared/JsonRpc.cs
IlyaBiryukov: Or, rather
```csharp
List<string> names;
if (!methodToAttributeMap.TryGetValue(method.Name, out names))
{
  names = new List<string>();
  methodToAttributeMap.Add(method.Name, names);
}

if (!names.Contains(attribute.Name, StringComparer.Ordinal))
{
  names.Add(attribute.Name);
}

```

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -67,6 +67,11 @@ internal OutstandingCallData(object taskCompletionSource, Action<JsonRpcMessage>
         /// A delegate for the <see cref="HandleInvocationTaskResult(JToken, Task)"/> method.
         /// </summary>
         private readonly Func<Task, object, JsonRpcMessage> handleInvocationTaskResultDelegate;
+        
+        /// <summary>
+        /// A map of the target object's method names to MethodNameAttribute value.
+        /// </summary>
+        private readonly Dictionary<string, List<string>> methodToAttributeMap;
```
IlyaBiryukov: It's not clear from the name what attribute it is about. Should it be named _methodNames_ instead?

src/StreamJsonRpc.Shared/JsonRpc.cs
tinaschrepfer: You're right, I think this was left over from a previous implementation.  I'll remove it.

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
IlyaBiryukov: Allow multiple = true would allow the same method to have different aliases? Seems to be useful.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
IlyaBiryukov: +1

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
tinaschrepfer: This would make differentiating between base class and derived class super difficult if they have different names for the same method.  Right now I throw an error as soon as I see that a method names to multiple names (MethodNameAttribute values).  If we support multiple = true, the logic to determine whether multiple names for a single method is valid (multiple attributes on the same method in the same class) or invalid (base and derived classes have different values for MethodNameAttribute).  There isn't a real life usage for having to declare multiple names for a method, so I left it as not allowed.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
tinaschrepfer: The problem with reversing the dictionary is that to we still need to traverse the whole map to detect method name conflicts.  We could address this by having 2 dictionaries, but I thought that was a bit overkill since look up here should be pretty quick.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
tinaschrepfer: When we first reflect over it it's in the ctor of JsonRpc, are you saying the construction should fail then?  I thought that might be overkill since maybe only one method has this error, but the rest are fine.

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Microsoft;
+
+namespace StreamJsonRpc
+{
+    /// <summary>
+    /// Attribute which allows a method to have an alternate name.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    public class MethodNameAttribute : Attribute
```
IlyaBiryukov: _MethodNameAttribute_ is kind of generic. How about _JsonRpcMethodAttribute_? 
Later we can add more properties to it; as it wouldn't be limited to just the method name.

src/StreamJsonRpc/MultilingualResources/StreamJsonRpc.pl.xlf
```diff
@@ -151,6 +151,11 @@
           <source>Parameter is not in the form of a single object</source>
           <target state="new">Parameter is not in the form of a single object</target>
         </trans-unit>
+        <trans-unit id="ConflictingMethodNameAttribute" translate="yes" xml:space="preserve">
+          <source>The '{0}' method has multiple MethodNameAttribute defined with different values.</source>
+          <target state="new">The '{0}' method has multiple MethodNameAttribute defined with different values.</target>
```
IlyaBiryukov: Should "MethodNameAttribute" it be paremeterized instead? Then you can use nameof() or typeof() as format argument.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
tinaschrepfer: Added explanation.  The reason why we throw conflict for all cases is because we we do the method invocation, we always invoke the method in the derived method.  If there are conflicts, you can argue that as long as we pass in a method name that's the same as the derived method MethodNameAttribute, then we should invoke that method and not throw an exception.  But what happens if you pass in the base method MethodNameAttribute?  You would expect it to invoke the base method, but it doesn't.  For simplicity (also for the sake of I can't see a reason why you would ever want conflicts), I made it throw as soon as a conflict is detected.

tinaschrepfer:
Pushed out a change which addressed feedback.  There are still a few outstanding issues that I didn't address since there's a pending discussion.  Let me know what you guys think about those.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
AArnott: Conflicts should be detected and reported during JsonRpc construction rather than when a method call comes in.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -663,13 +663,14 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+        /// Creates a dictionary which maps a request method name to its clr method name via JsonRpcMethodAttribute value.
         /// </summary>
         /// <param name="target"></param>
         /// <returns></returns>
-        private static Dictionary<string, List<string>> GetMethodToAttributeMap(object target)
+        private static Dictionary<string, string> GetRequestMethodToClrMethodMap(object target)
         {
-            var methodToAttributeMap = new Dictionary<string, List<string>>(StringComparer.Ordinal);
+            // First we create a dictionary to map clr method names to JsonRpcmethodAttribute value, this simplifies the process of detecting attribute name conflicts.
+            var methodToAttributeMap = new Dictionary<string, string>(StringComparer.Ordinal);
```
AArnott: Consider updating the name of this variable to match what you're storing. Perhaps: `clrMethodToRequestMethodMap`

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -679,23 +680,31 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
-                            }
+                                methodToAttributeMap.Add(method.Name, attribute.Name);
+                            }                            
                         }
                     }
                 }
             }
 
-            return methodToAttributeMap;
+            // If not conflicts are find, then create another map which reverses the key/value mapping.
```
AArnott: grammer nit: "If not conflicts are find" -> "If *no* conflicts are *found*"

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -679,23 +680,31 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+                                }
                             }
-
-                            if (!names.Contains(attribute.Name, StringComparer.Ordinal))
+                            else
                             {
-                                names.Add(attribute.Name);
-                            }
+                                methodToAttributeMap.Add(method.Name, attribute.Name);
+                            }
```
AArnott: Format Document to remove trailing whitespace.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -679,23 +680,31 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+                                    throw new RemoteTargetException(string.Format(CultureInfo.CurrentCulture, Resources.ConflictingMethodNameAttribute, method.Name, nameof(JsonRpcMethodAttribute)));
+                                }
                             }
-
-                            if (!names.Contains(attribute.Name, StringComparer.Ordinal))
+                            else
                             {
-                                names.Add(attribute.Name);
-                            }
+                                methodToAttributeMap.Add(method.Name, attribute.Name);
```
AArnott: It looks like the effect you're going for is that if I define two overloads with the same method name in C#, and only one overload bears the attribute that changes the name, both methods are impacted. That's a bit odd, but with adequate documentation on the attribute I think we can live with it.
But what if instead we invert the dictionary we're creating initially (rather than inverting it at the bottom of your method) and simply store the request method name as the key up front? Then the attribute can apply to just the one overload it was given? Of course, then you'd need to define what to do if two methods of different CLR names collide in their request method name. So IMO the way that should work is your dictionary's value is a `List<MethodInfo>` and you allow the collisions. You also add an entry to the dictionary for every single method on the type rather than only those with attributes. The dictionary then becomes the complete map of: "if you see a request method coming in with *x*, here are all the MethodInfo's to look for a compatible signature". The TargetMethod resolver can already respond to method signature ambiguity (which we can move earlier during reflection in a subsequent change to this library). 

The reason I suggest this change is I think it will lend to a more predictable API use for the consumer, and it's 'future proof' to enhancing our disambiguation logic in a few ways that I'm thinking we may want to.

AArnott: If you end up storing `MethodInfo`, I think you already have a test to make sure the system invokes an override method instead of a virtual one, so that should be useful to verify we don't have issues there.

tinaschrepfer: Discussed offline with @AArnott and agreed that conflicts should never be allowed, and all overloads must declare the attribute if one of them declares it.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -663,48 +663,79 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
             }
 
-            // If not conflicts are find, then create another map which reverses the key/value mapping.
-            var attributeToMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
-            foreach (var method in methodToAttributeMap)
+            // We shouldn't ever hit this unless there's a bug in the above code.  Put in so we can detect bugs easier.
+            if (clrMethodToRequestMethodMap.Count != requestMethodToClrMethodMap.Count)
             {
-                attributeToMethodMap.Add(method.Value, method.Key);
+                throw new InvalidOperationException("There was an error reflecting over the target object.");
```
AArnott: Please replace this if block with this:
```csharp
Assumes.True(clrMethodToRequestMethodMap.Count == requestMethodToClrMethodMap.Count, "There was an error reflecting over the target object.");
```

That throws an internal error exception, which is more appropriate given your comment than an InvalidOperationException.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -586,13 +586,61 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
...
     {
-        var invalidServer = new InvalidOverloadServer();
+        var invalidServer = new ConflictingOverloadServer();
 
         var streams = Nerdbank.FullDuplexStream.CreateStreams();
         var serverStream = streams.Item1;
         var clientStream = streams.Item2;
 
-        Assert.Throws<RemoteTargetException>(() => JsonRpc.Attach(this.serverStream, invalidServer));
+        Assert.Throws<ArgumentException>(() => JsonRpc.Attach(this.serverStream, invalidServer));
```
AArnott: Thanks for testing the failure case.
Can you also test that `new JsonRpc(...)` would also throw?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -604,8 +652,8 @@ public void JsonRpcMethodAttribute_ConflictOverrideMethodsThrowsException()
         var serverStream = streams.Item1;
         var clientStream = streams.Item2;
 
-        Assert.Throws<RemoteTargetException>(() => JsonRpc.Attach(this.serverStream, invalidServer));
-    }
+        Assert.Throws<ArgumentException>(() => JsonRpc.Attach(this.serverStream, invalidServer));
+    }
```
AArnott: Format Document to eliminate trailing whitespace.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -663,48 +663,79 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+            var clrMethodToRequestMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
+
+            // Create a dictionary to map <see cref="JsonRpcMethodAttribute" /> values to clr method names, used to detect that the same attribute value doesn't map to multiple methods.
+            var requestMethodToClrMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
+
             if (target != null)
             {
                 for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
                 {
+                    HashSet<string> clrMethodsWithoutReplacementName = new HashSet<string>(StringComparer.Ordinal);
```
AArnott: What if instead of maintaining this 3rd collection, you simply added all methods to your two dictionaries regardless whether they have attributes applied or not? Then you'll detect the conflicts you mean to, I *think*.
Not a required change. Just an idea.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -655,6 +662,82 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+                            {
+                                throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.OverloadsMissingMethodAttribute, method.Name, nameof(JsonRpcMethodAttribute)));
+                            }
+
+                            string attributeValue;
+                            if (clrMethodToRequestMethodMap.TryGetValue(method.Name, out attributeValue))
+                            {
+                                if (!string.Equals(attribute.Name, attributeValue, StringComparison.Ordinal))
+                                {
+                                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, Resources.ConflictingMethodNameAttribute, method.Name, nameof(JsonRpcMethodAttribute)));
```
AArnott: Tip: Your `throw new ArgumentException(string.Format(CultureInfo.CurrentCulture,` line could be reduced to start with just `Requires.Fail(`. It takes care of the throw, the string.Format and CultureInfo for you.

src/StreamJsonRpc.Desktop/Resources.resx
```diff
@@ -121,6 +121,14 @@
     <value>Both readable and writable are null.</value>
     <comment>'readable' and 'writable' are parameter names and should not be localized.</comment>
   </data>
+  <data name="ConflictingMethodAttributeValue" xml:space="preserve">
+    <value>The {0} Name value of {1} maps to multiple methods.</value>
```
AArnott: I think this could be more clearly actionable to the user if it were reworded to:

> .NET methods "{type1}.{method1}" and "{type2}.{method2}" cannot both map to the same request method name: {requestMEthodName}."

Mentioning the attribute and property name I think is not necessary. They'll figure it out if we point them directly at the problematic methods.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -572,17 +709,110 @@ private static void SendObject(Stream receivingStream, object jsonObject)
...
+        [JsonRpcMethod("test/string")]
+        public string InvokeOverloadConflictingMethodAttribute(string test) => $"conflicting string: {test}";
 
+        public string InvokeOverloadConflictingMethodAttribute(string arg1, int arg2) => $"conflicting string: {arg1} int: {arg2}";
+    }
+
+    /// <summary>
+    /// This class is invalid because two different methods in the same class have the same <see cref="JsonRpcMethodAttribute" /> value.
+    /// </summary>
+    public class SameAttributeUsedOnDifferentMethodsServer : BaseClass
```
AArnott: Can you add another test case:

```csharp
public class Foo
{
    [JsonRpcMethod("Second")]
    public void First() { }

    public void Second() { }
}

```

Obviously, I think we'd expect this to be a failure case.

AArnott: Also, as we have special code to allow equivalence between SomeMethod and SomeMethodAsync, some tests around that and this alternate method name feature sounds appropriate, both success and collision failure cases.

src/StreamJsonRpc.Desktop/Resources.resx
```diff
@@ -121,6 +121,14 @@
...
+    <comment>{0} is the attribute name {1} is the attribute value.</comment>
+  </data>
+  <data name="ConflictingMethodNameAttribute" xml:space="preserve">
+    <value>The '{0}' method has multiple {1} defined with different values.</value>
```
AArnott: I think this could be more clearly actionable to the user if it were reworded to:

> All overloads of the "{type}.{method}" method must share a common value for JsonRpcMethodAttribute.Name.

src/StreamJsonRpc.Desktop/Resources.resx
```diff
@@ -179,6 +187,10 @@
   <data name="NonNegativeIntegerRequired" xml:space="preserve">
     <value>A non-negative integer is required.</value>
   </data>
+  <data name="OverloadsMissingMethodAttribute" xml:space="preserve">
+    <value>The {0} method has {1} defined, but not all of its overloads do.</value>
```
AArnott: I think you can delete the message and reuse the `ConflictingMethodNameAttribute` message after you reword that as I have suggested.

src/StreamJsonRpc.Shared/JsonRpc.cs
tinaschrepfer: Good idea.  I modified the logic slightly so it's cleaner.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
tinaschrepfer: Added a few tests for this.  Also I realized the attribute is really an alternate, not a replacement.  That is, if a method has a JsonRpcMethodAttribute defined, you can still call the same method using the CLR method name or the attribute name value.  We could make it more strict so that it's a full on replacement, so once you've defined a JsonRpcMethodAttribute you can't call the actual CLR method anymore, but that's more logic to guard against that.  I didn't see a need for being strict in this sense so I left it as is.  Let me know what you think about that.

src/StreamJsonRpc.Shared/Reflection/MethodNameAttribute.cs
tinaschrepfer: This is actually alternate, not replace.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
AArnott: Thanks for raising that. That is cause for concern, IMO. I wouldn't have anticipated that from the attribute name or the documentation you had on it before. So at minimum we'd want the docs to be clear on this point. But I think unless someone has studied these updated docs, the expected behavior would be that the original method name no longer works. And fixing that later would be a protocol breaking change. So I think we should fix that now.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -559,6 +559,234 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
         await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync<string>(nameof(Server.TestInvalidMethod)));
     }
 
+    [Fact]
+    public async Task InvokeAsync_InvokesMethodWithAttributeSet()
```
AArnott: I love the tests you're adding. I think we have a substantial set here now that it's probably worth moving them to their own file.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
AArnott: BTW, we should have tests verifying the behavior (whichever way we go) around the original method name being accepted or rejected. Do you already have such tests?

tinaschrepfer: Changed the behavior so it's a replacement.  I had a test for the alternate name case, but changed that now to cover replacement.  It's expected to fail if I call a CLR method name on a method that has the attribute defined.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -69,9 +69,14 @@ internal OutstandingCallData(object taskCompletionSource, Action<JsonRpcMessage>
...
 
         /// <summary>
-        /// A map of the target object's method names to JsonRpcMethodAttribute.Name value.
+        /// A dictionary to map clr method names to <see cref="JsonRpcMethodAttribute" /> values, used to detect that same method shouldn't have multiple attribute values.
         /// </summary>
-        private readonly Dictionary<string, string> requestMethodToClrMethodMap;
+        private Dictionary<string, string> requestMethodToClrMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
+
+        /// <summary>
+        /// A dictionary to map <see cref="JsonRpcMethodAttribute" /> values to clr method names, used to detect that the same attribute value doesn't map to multiple methods.
```
AArnott: Both of these fields claim to be present to detect errors. As that is the only use you call out, are they not also used in the success cases?

src/StreamJsonRpc.Shared/Reflection/JsonRpcMethodAttribute.cs
```diff
@@ -4,7 +4,7 @@
 namespace StreamJsonRpc
 {
     /// <summary>
-    /// Attribute which allows a method to have an alternate name.
+    /// Attribute which allows a method to have a different name.  If applied on a method, the method's CLR name can no longer be used for remote invocation.
```
AArnott: consider: "allows a method to have a different name" -> "changes the name by which this method can be invoked over JSON-RPC." Your clarifying second sentence sounds great.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -69,9 +69,14 @@ internal OutstandingCallData(object taskCompletionSource, Action<JsonRpcMessage>
...
+        /// </summary>
+        private Dictionary<string, string> clrMethodToRequestMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
```
AArnott: Also, we should probably use `ReadOnlyDictionary<string,string>` for these fields as a way to guarantee they don't change later on since we depend on them being thread-safe.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -655,6 +667,69 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
             }
         }
 
+        /// <summary>
+        /// Creates a dictionary which maps a request method name to its clr method name via <see cref="JsonRpcMethodAttribute" /> value.
+        /// </summary>
+        /// <param name="target">Object to reflect over and analyze its methods.</param>
+        /// <returns>Dictionary which maps a request method name to its clr method name.</returns>
+        private void GetRequestMethodToClrMethodMap(object target)
```
AArnott: This method no longer returns anything. Perhaps `Get` isn't the right verb to lead its name with.

src/StreamJsonRpc.Shared/Reflection/JsonRpcMethodAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    /// <summary>
+    /// Attribute which allows a method to have a different name.  If applied on a method, the method's CLR name can no longer be used for remote invocation.
+    /// </summary>
+    /// <remarks>
+    /// This attribute should be used when rpc message method names can be different from the actual CLR method names.
+    /// Useful in cases where rpc message method names contain illegal characters for CLR method names, i.e. "text/OnDocumentChanged".
+    /// 
+    /// If methods are overloaded, each overload must define its own <see cref="JsonRpcMethodAttribute"/>  with all the same values.  
+    /// Conflicts will result in error being thrown during <see cref="JsonRpc"/> construction.
```
AArnott: I'm glad we expect conflicts to be thrown up-front. But why then do the docs on the two dictionary fields we now have on JsonRpc state that their presence is for enforcement? Is that stale?

src/StreamJsonRpc.Shared/Reflection/JsonRpcMethodAttribute.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// 
+    /// If methods are overridden, the base class can define a <see cref="JsonRpcMethodAttribute"/> and derived classes will inherit the attribute.
+    /// If derived class and base class have conflicting <see cref="JsonRpcMethodAttribute"/> values for a method, an error will be thrown during <see cref="JsonRpc"/> construction.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
```
AArnott: Does a test fail if you set `Inherited = false` here?

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -655,6 +667,69 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+        {
+            if (target != null)
+            {
+                for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
+                {
+                    foreach (MethodInfo method in t.DeclaredMethods)
+                    {
+                        var attribute = (JsonRpcMethodAttribute)method.GetCustomAttribute(typeof(JsonRpcMethodAttribute));
+                        var attributeValue = attribute == null ? null : attribute.Name;
```
AArnott: Tip: 

```csharp
var attributeValue = attribute == null ? null : attribute.Name;
```
is equivalent to
```csharp
var attributeValue = attribute?.Name;
```

And as the type is not stated elsewhere on this line, using `string` instead of `var` would be preferable here.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -655,6 +667,69 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+                            {
+                                Requires.Fail(Resources.ConflictingMethodNameAttribute, method.Name, nameof(JsonRpcMethodAttribute), nameof(JsonRpcMethodAttribute.Name));
+                            }
+                        }
+                        else
+                        {
+                            clrMethodToRequestMethodMap.Add(method.Name, attributeValue);
+                        }
+
+                        if (attributeValue != null)
```
AArnott: Why not unconditionally add the requestMethodToClrMethodMap entry? We always need to do that mapping at runtime for incoming requests, and it seems to me like if we guaranteed an entry in this map, we wouldn't need to keep the second dictionary around.

src/StreamJsonRpc.Shared/Reflection/JsonRpcMethodAttribute.cs
AArnott: I tested it. Yes it does. :smile:

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: @tinaschrepfer  I'm trying my hand at fixing this myself.

src/StreamJsonRpc.Shared/Reflection/JsonRpcMethodAttribute.cs
AArnott: Resolved with my code change.

src/StreamJsonRpc.Shared/Reflection/JsonRpcMethodAttribute.cs
AArnott: Fixed

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Fixed. Please review the commits I pushed to your branch.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Fixed.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Fixed. It now returns an object again.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Fixed.

src/StreamJsonRpc.Shared/JsonRpc.cs
AArnott: Fixed

AArnott:
@tinaschrepfer after you approve of my changes, I'll merge.

src/StreamJsonRpc.Shared/JsonRpc.cs
```diff
@@ -655,6 +667,78 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
...
+            var clrMethodToRequestMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
+            var requestMethodToClrMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
+            var candidateAliases = new Dictionary<string, string>(StringComparer.Ordinal);
+
+            for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
+            {
+                foreach (MethodInfo method in t.DeclaredMethods)
+                {
+                    var attribute = (JsonRpcMethodAttribute)method.GetCustomAttribute(typeof(JsonRpcMethodAttribute));
+                    var requestName = attribute?.Name ?? method.Name;
```
tinaschrepfer: Ah, I see what you mean now.  That makes sense!

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
```diff
@@ -12,51 +12,50 @@ namespace StreamJsonRpc
...
             Requires.NotNull(target, nameof(target));
+            Requires.NotNull(jsonSerializer, nameof(jsonSerializer));
+            Requires.NotNull(requestMethodToClrMethodMap, nameof(requestMethodToClrMethodMap));
 
             this.request = request;
             this.target = target;
 
-            var targetMethods = new Dictionary<MethodSignature, object[]>();
-            for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
+            if (requestMethodToClrMethodMap.TryGetValue(request.Method, out string clrMethodName))
```
tinaschrepfer: Neat, I didn't know you can declare the variable directly in TryGetValue.

src/StreamJsonRpc.Shared/JsonRpc.cs
tinaschrepfer: I see what you mean now.  This makes sense!

tinaschrepfer:
@AArnott, not sure how I would approve this review as I'm not finding a button which lets me approve it.  But please feel free to merge.

src/StreamJsonRpc.Shared/Reflection/TargetMethod.cs
AArnott: Any `out` parameter can now declare its receiving variable as of C# 7

PR #25, opened 2017-04-04
Migrate to .NET SDK projects

javierdlg:
This migrates our project to use the new VS2017 .NET SDK project type.

The only source code changes made here were a few #if conditions (from DESKTOP to NET45). Most of the changes then are to build authoring, and merging the Desktop and Shared projects into the portable version of StreamJsonRpc


javierdlg:
@AArnott I am unfamiliar with AppVeyor and I'm unsure as to what to do regarding the failing check.

AArnott:
I'm looking into it

PR #43, opened 2017-06-19
Implement multiple target support for StreamJsonRpc

tinaschrepfer:
Implement multiple target support for StreamJsonRpc

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -558,6 +558,44 @@ public class JsonRpcTests : TestBase
         await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync<string>(nameof(Server.TestInvalidMethod)));
     }
 
+    [Fact]
+    public void AddLocalRpcTarget_ExceptionThrownWhenRpcHasStartedListening()
+    {
+        Assert.Throws<InvalidOperationException>(() => this.clientRpc.AddLocalRpcTarget(new AdditionalServerTargetOne()));
+    }
+
+    [Fact]
+    public async Task AddLocalRpcTarget_AdditionalTargetMethodFound()
+    {
+        var streams = Nerdbank.FullDuplexStream.CreateStreams();
+        var rpc = new JsonRpc(streams.Item1, streams.Item2);
+        rpc.AddLocalRpcTarget(new Server(), new AdditionalServerTargetOne(), new AdditionalServerTargetTwo());
+        rpc.StartListening();
+
+        var serverMethodResult = await rpc.InvokeAsync<string>(nameof(Server.ServerMethod), "test");
+        Assert.Equal("test!", serverMethodResult);
+
+        var plusOneResultInt = await rpc.InvokeAsync<int>(nameof(AdditionalServerTargetOne.PlusOne), 1);
+        Assert.Equal(2, plusOneResultInt);
+
+        var plusOneResultString = await rpc.InvokeAsync<string>(nameof(AdditionalServerTargetTwo.PlusOne), "one");
+        Assert.Equal("one plus one!", plusOneResultString);
+
+        var plusTwoResult = await rpc.InvokeAsync<int>(nameof(AdditionalServerTargetTwo.PlusTwo), 1);
+        Assert.Equal(3, plusTwoResult);
+    }
+
+    [Fact]
+    public async Task AddLocalRpcTarget_NoTargetContainsRequestedMethod()
+    {
+        var streams = Nerdbank.FullDuplexStream.CreateStreams();
+        var rpc = new JsonRpc(streams.Item1, streams.Item2);
+        rpc.AddLocalRpcTarget(new Server(), new AdditionalServerTargetOne(), new AdditionalServerTargetTwo());
+        rpc.StartListening();
+
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => rpc.InvokeAsync("PlusThree", 1));
+    }
+
     protected override void Dispose(bool disposing)
     {
         if (disposing)
@@ -780,6 +818,32 @@ public class JsonRpcTests : TestBase
 
         internal void InternalMethod()
         {
+        }
+    }
+
+    public class AdditionalServerTargetOne
+    {
+        public int PlusOne(int arg)
+        {
+            return arg + 1;
+        }
+    }
+
+    public class AdditionalServerTargetTwo
+    {
+        public int PlusOne(int arg)
+        {
+            return arg + 3;
+        }
+
+        public string PlusOne(string arg)
+        {
+            return arg + " plus one!";
+        }
+
+        public int PlusTwo(int arg)
+        {
+            return arg + 2;
         }
     }
 
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -559,6 +559,14 @@ public class JsonRpcTests : TestBase
     }
 
     [Fact]
+    public async Task InvokeAsync_NullRefExceptionNotThrownIfNoTargetObjectIsAttached()
+    {
+        var streams = Nerdbank.FullDuplexStream.CreateStreams();
+        var rpc = JsonRpc.Attach(streams.Item1, streams.Item2);
+        await Assert.ThrowsAnyAsync<RemoteTargetNotSetException>(() => rpc.InvokeAsync<string>(nameof(Server.ServerMethod), "test"));
+    }
+
+    [Fact]
     public void AddLocalRpcTarget_ExceptionThrownWhenRpcHasStartedListening()
     {
         Assert.Throws<InvalidOperationException>(() => this.clientRpc.AddLocalRpcTarget(new AdditionalServerTargetOne()));
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -559,14 +559,6 @@ public class JsonRpcTests : TestBase
     }
 
     [Fact]
-    public async Task InvokeAsync_NullRefExceptionNotThrownIfNoTargetObjectIsAttached()
-    {
-        var streams = Nerdbank.FullDuplexStream.CreateStreams();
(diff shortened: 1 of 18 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -64,17 +63,15 @@ public class JsonRpc : IDisposableObservable
         /// </summary>
         private readonly Func<Task, object, JsonRpcMessage> handleInvocationTaskResultDelegate;
 
-        /// <summary>
-        /// A dictionary to map clr method names to <see cref="JsonRpcMethodAttribute" /> values.
-        /// </summary>
-        private readonly ReadOnlyDictionary<string, string> requestMethodToClrMethodMap;
+        private readonly List<Tuple<object, ReadOnlyDictionary<string, string>>> targetRequestMethodToClrMethodMap;
```
AArnott: Can you add/update the xml doc comment for this member to clearly describe the contents of this collection?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -64,17 +63,15 @@ public class JsonRpc : IDisposableObservable
...
 
         private readonly CancellationTokenSource disposeCts = new CancellationTokenSource();
 
         private Task readLinesTask;
         private int nextId = 1;
         private bool disposed;
         private bool hasDisconnectedEventBeenRaised;
+        private bool startedListening = false;
```
AArnott: FxCop whines when you initialize fields to their default values. Can you drop the `=false` from this field?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -230,10 +225,37 @@ public static JsonRpc Attach(Stream sendingStream, Stream receivingStream, objec
         }
 
         /// <summary>
+        /// Adds the specified targets as possible objects to invoke when incoming messages are received.
```
AArnott: Either here or in the `<remarks>` section, can you explain the relationship these targets have with each other, and how order is significant?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -119,9 +115,8 @@ public JsonRpc(DelimitedMessageHandler messageHandler, object target = null)
             };
             this.JsonSerializer = new JsonSerializer();
 
-            this.requestMethodToClrMethodMap = this.callbackTarget != null
-                ? new ReadOnlyDictionary<string, string>(GetRequestMethodToClrMethodMap(this.callbackTarget))
-                : EmptyDictionary;
+            this.targetRequestMethodToClrMethodMap = new List<Tuple<object, ReadOnlyDictionary<string, string>>>();
```
AArnott: nit: can you move this to a field initializer?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -230,10 +225,37 @@ public static JsonRpc Attach(Stream sendingStream, Stream receivingStream, objec
         }
 
         /// <summary>
+        /// Adds the specified targets as possible objects to invoke when incoming messages are received.
+        /// </summary>
+        /// <param name="targets">Targets to invoke when incoming messages are received.</param>
+        /// <remarks>This method must be called before JsonRpc starts listening for messages.</remarks>
+        public void AddLocalRpcTarget(params object[] targets)
```
AArnott: Your method takes a plural argument yet your method name suggests singular.
We can rename the method, or take a singular parameter. I'm leaning toward a singular parameter, myself, based on my belief that most callers will have just one target to "add" so we'll avoid a loop and an array allocation by just taking one. And callers would then be responsible to avoid calling it with a null argument.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -230,10 +225,37 @@ public static JsonRpc Attach(Stream sendingStream, Stream receivingStream, objec
...
+        {
+            Requires.NotNull(targets, nameof(targets));
+
+            if (this.startedListening)
+            {
+                throw new InvalidOperationException("Cannot attach additional targets once JsonRpc has started listening for messages.");
```
AArnott: Please move the message to `Strings.resx` so it can be localized.
Also, you can reduce these 4 lines to just one by calling `Verify.Operation`

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -230,10 +225,37 @@ public static JsonRpc Attach(Stream sendingStream, Stream receivingStream, objec
...
+            }
+
+            foreach (var target in targets)
+            {
+                if (target != null)
+                {
+                    this.targetRequestMethodToClrMethodMap.Add(new Tuple<object, ReadOnlyDictionary<string, string>>(
```
AArnott: Tip: you can use `Tuple.Create(` instead of `new` here to avoid having to specify all the generic type arguments since the inferred type arguments will be equivalent.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -558,6 +558,44 @@ await invokeTask.NoThrowAwaitable(); // may or may not throw due to cancellation
         await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync<string>(nameof(Server.TestInvalidMethod)));
     }
 
+    [Fact]
+    public void AddLocalRpcTarget_ExceptionThrownWhenRpcHasStartedListening()
+    {
+        Assert.Throws<InvalidOperationException>(() => this.clientRpc.AddLocalRpcTarget(new AdditionalServerTargetOne()));
```
AArnott: Can we also test a null `object[]` and an array with null elements?

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Done.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Done.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Done.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Done

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Done.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Done.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
tinaschrepfer: Done.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -230,10 +230,26 @@ public static JsonRpc Attach(Stream sendingStream, Stream receivingStream, objec
...
         /// <summary>
+        /// Adds the specified target as possible object to invoke when incoming messages are received.  The target object
+        /// should not inherit from each other and are invoked in the order which they are added.
+        /// </summary>
+        /// <param name="target">Target to invoke when incoming messages are received.</param>
+        /// <remarks>This method must be called before JsonRpc starts listening for messages.</remarks>
+        public void AddLocalRpcTarget(object target)
+        {
+            Requires.NotNull(target, nameof(target));
+            Verify.Operation(!this.startedListening, Resources.BothReadableWritableAreNull);
```
IlyaBiryukov: Copy-pasted? Should be a different error message I think.

AArnott:
@tinaschrepfer just to call out that Ilya's last couple of comments are still pending resolution.
Also: what version of VS do you anticipate consuming these changes in?

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: That's so weird.  I swear when I wrote this it was the right message.  I'll change it.

tinaschrepfer:
@AArnott, comments addressed.  Our feature is targeting Dev16.  However, without these changes in place for StreamJsonRpc, I can't consume the new APIs for custom messages.  If you could push these changes into vscore or something, I can always build against that branch.  Thanks!

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -773,13 +773,13 @@ private static JsonRpcMessage CreateError(JToken id, Exception exception)
                 foreach (var targetMap in this.targetRequestMethodToClrMethodMap)
                 {
                     targetMethod = new TargetMethod(request, targetMap.Item1, jsonSerializer, targetMap.Item2);
-                    if (targetMethod.IsFound)
+                    if (targetMethod != null && targetMethod.IsFound)
```
AArnott: We don't need a `targetMethod != null` check here because you just assigned it with `new`, which cannot return null.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -773,13 +773,13 @@ private static JsonRpcMessage CreateError(JToken id, Exception exception)
...
                     {
                         break;
                     }
                 }
 
-                if (!targetMethod.IsFound)
+                if (targetMethod == null || !targetMethod.IsFound)
```
AArnott: Thanks for this change. Can we add a test to prove we never throw NRE when `targetMethod==null`, since evidently we are missing one?

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Whoops, wasn't thinking.  Removed.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: I actually don't think targetMethod can ever be null.  The only scenario where that's possible is if targetRequestMethodToClrMethodMap is empty.  But we already check for that beforehand and return a different error.  In any case, I added a test to test that a remote target exception is thrown if we call InvokeAsync with no target object.

AArnott: doh! I guess we already had such a test. I'll remove the extra and merge your PR. Thanks!

PR #46, opened 2017-07-18
Fix Live Unit Testing for this repo

AArnott:


src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -465,6 +465,7 @@ public void SetEncodingToNullThrows()
     }
 
     [Fact]
+    [Trait("Category", "SkipWhenLiveUnitTesting")]
```
IlyaBiryukov: Are these tests skipped because thy apply GC pressure which we don't want to do inproc in LUT in VS?

IlyaBiryukov: Maybe add a comment explaining that, so next GC pressure tests would follow the suit.

AArnott: Sure. I'll add a comment. The reason is the tests are flaky. :(

PR #53, opened 2017-08-23
Prevent RPC from invoking methods on System.Object

### LINKED ISSUE #51: Methods on System.Object should not be invokable by RPC client
When attaching/constructing JsonRpc with target, all methods are registered into requestMethodToClrMethodMap.  Providing more control over this process would be welcome and more safe at the same time.
Even when using separate class which provides only desired methods, also object base methods like 'ToString' are registered.

Example: registering test class with methods:
TestMethod0, TestMethod1, TestMethod2, TestMethod3

Full list of extra methods:
GetHashCode, GetType, Equals, ToString, ReferenceEquals, Finalize, MemberwiseClone, FieldSetter, FieldGetter, GetFieldInfo

This, depending on the actual implementation, is a vulnerability.

My suggestion is to provide new (or extend existing JsonRpcMethodAttribute), which is mandatory for each  method published over JsonRPC.

AArnott:
Fixes #51

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -303,6 +303,14 @@ public class JsonRpcTests : TestBase
     }
 
     [Fact]
+    public async Task CannotCallMethodsOnSystemObject()
+    {
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(object.ToString)));
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(object.GetHashCode)));
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(object.GetType)));
+    }
+
+    [Fact]
     public async Task CannotCallPrivateMethod()
     {
         await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(Server.InternalMethod), 10));
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -660,7 +660,7 @@ namespace StreamJsonRpc
             var requestMethodToClrMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
             var candidateAliases = new Dictionary<string, string>(StringComparer.Ordinal);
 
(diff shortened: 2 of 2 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -660,7 +660,7 @@ protected virtual Task<ReturnType> InvokeCoreAsync<ReturnType>(int? id, string t
             var requestMethodToClrMethodMap = new Dictionary<string, string>(StringComparer.Ordinal);
             var candidateAliases = new Dictionary<string, string>(StringComparer.Ordinal);
 
-            for (TypeInfo t = target.GetType().GetTypeInfo(); t != null; t = t.BaseType?.GetTypeInfo())
+            for (TypeInfo t = target.GetType().GetTypeInfo(); t != null && t != typeof(object).GetTypeInfo(); t = t.BaseType?.GetTypeInfo())
```
IlyaBiryukov: What if target object overrides one of these methods? Should JsonRpc be allowed to call it?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -303,6 +303,14 @@ public void Attach_Null_Throws()
     }
 
     [Fact]
+    public async Task CannotCallMethodsOnSystemObject()
+    {
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(object.ToString)));
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(object.GetHashCode)));
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(object.GetType)));
```
IlyaBiryukov: How about other methods, like Finalize or Equals?

AArnott: I felt this was an adequate enough sampling of methods to demonstrate that they weren't available. The product would have to do something rather specific to allow some System.Object methods and not others.

src/StreamJsonRpc/JsonRpc.cs
AArnott: That's a very interesting question. Perhaps not.

AArnott: I prepared a change that did this. It relies on `MethodInfo.GetBaseDefinition()` but that doesn't exist on .netstandard1.1 (it was introduced in netstandard1.5). So rather than have divergent behavior across platforms or re-implementing the method myself (which is non-trivial and arguably impossible to do with 100% accuracy), I'm siding on not implementing it at this point.

Issue #55, opened 2017-08-30
HeaderDelimitedMessageHandler may write only header but not the body when the request is cancelled


IlyaBiryukov:
*Repro:*
Invoke a method on a server and then cancel the cancellation token passed to the method. Depending on your luck, the cancellation may happen so that the header is already written to the underlying stream [here](https://github.com/Microsoft/vs-streamjsonrpc/blob/master/src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs#L238), but the content is not, and it got cancelled [here](https://github.com/Microsoft/vs-streamjsonrpc/blob/master/src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs#L241). Sending another message on that JsonRPC makes it send the header and the content. The receiving end will get the header from the first (cancelled) message, and then the header and the content from the next message. It'll try to use the first header to read the whole message, and will read malformed json (it will contain the header and part of the body of the second message). Then it'll close the stream.

*Expected:*
Sending a JSON RPC message by the message handler should be atomic. We should write header and content together, or don't write anything at all.

AArnott:
@ilyabiryukov VS 15.3 is equivalent to the existing v1.1 milestone. So I've moved to that one.

PR #57, opened 2017-08-31
Make HeaderDelimitedMessageHandler write atomic

IlyaBiryukov:
HeaderDelimitedMessageHandler accepts cancellation token from the call
to JsonRpc.InvokeAsync. If the caller cancels the token at the right
moment, it may happen when the header is already written to the stream,
but the message content is not. If this happens, on the next call the
reciever will get a malformed message - a header from the first message,
a header and content from the next message. This will cause JsonRpc to
error out, close the stream, and cancel the next method.

This fix makes sure that writing in HeaderDelimiteMessageHandler occurs
atomically and cannot be interrupted by cancelled cancellation token.

Fix for #55.

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft;
 using Microsoft.VisualStudio.Threading;
+using Nerdbank;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using StreamJsonRpc;
@@ -22,10 +23,10 @@ public class JsonRpcTests : TestBase
     private const string HubName = "TestHub";
 
     private readonly Server server;
-    private readonly Stream serverStream;
+    private readonly FullDuplexStream serverStream;
     private readonly JsonRpc serverRpc;
 
-    private readonly Stream clientStream;
+    private readonly FullDuplexStream clientStream;
     private readonly JsonRpc clientRpc;
 
     public JsonRpcTests(ITestOutputHelper logger)
@@ -35,7 +36,7 @@ public class JsonRpcTests : TestBase
 
         this.server = new Server();
 
-        var streams = Nerdbank.FullDuplexStream.CreateStreams();
+        var streams = FullDuplexStream.CreateStreams();
         this.serverStream = streams.Item1;
         this.clientStream = streams.Item2;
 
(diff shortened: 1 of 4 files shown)


IlyaBiryukov:
@heejaechang Here is the fix in StreamJsonRpc

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
```diff
@@ -233,12 +233,18 @@ private enum HeaderParseState
             sendingBufferStream.Write(CrlfBytes, 0, CrlfBytes.Length);
 #pragma warning restore VSTHRD103 // Call async methods when in an async method
 
+            // Either write both the header and the message, or don't write anything.
+            // Fix for https://github.com/Microsoft/vs-streamjsonrpc/issues/55.
```
AArnott: I'd rather include details like the issue being fixed in the git commit comments. The code comments should be sufficiently explanatory on their own to help the user understand why the code is written the way it is without relying on context in a bug that is hopefully still accessible. Sometimes when the issue is particularly complex, or the bug is still active and the comment is on a disabled test, I'll go ahead and put the link to the bug in the code, but otherwise I like to avoid it.

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
```diff
@@ -233,12 +233,18 @@ private enum HeaderParseState
...
+
             // Transmit the headers.
+            // Ignore the cancellation token.
             sendingBufferStream.Position = 0;
-            await sendingBufferStream.CopyToAsync(this.SendingStream, MaxHeaderElementSize, cancellationToken).ConfigureAwait(false);
+            await sendingBufferStream.CopyToAsync(this.SendingStream, MaxHeaderElementSize).ConfigureAwait(false);
 
             // Transmit the content itself.
-            await this.SendingStream.WriteAsync(contentBytes, 0, contentBytes.Length, cancellationToken).ConfigureAwait(false);
+            // Ignore the cancellation token.
```
AArnott:
```csharp
// Ignore the cancellation token so we don't write the header without the message body.
```

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
AArnott: I'd rather include details like the issue being fixed in the git commit comments. The code comments should be sufficiently explanatory on their own to help the user understand why the code is written the way it is without relying on context in a bug that is hopefully still accessible. Sometimes when the issue is particularly complex, or the bug is still active and the comment is on a disabled test, I'll go ahead and put the link to the bug in the code, but otherwise I like to avoid it.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -376,6 +377,47 @@ public void SetEncodingToNullThrows()
     }
 
     [Fact]
+    [Trait("Bug", @"https://github.com/Microsoft/vs-streamjsonrpc/issues/55")]
+    [Trait("Bug-Workaround", @"https://github.com/Microsoft/vs-streamjsonrpc/issues/56")]
```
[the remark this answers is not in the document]
IlyaBiryukov: Sure

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
IlyaBiryukov: OK. Though it's easier to find the bug if it's mentioned in the comment than by looking at the commit that introduced the change, especially some time after, when more commits start affect these lines.
I can add more explanation about the reason why we do this and also leave the reference to the bug.

AArnott:
We can merge with Anson's approval.

PR #58, opened 2017-09-01
Fix race condition that can drop a cancellation request

### LINKED ISSUE #56: Cancellation request may be ignored if it comes just right after the request it cancels
*Repro*:
1. Invoke a request with cancellation
2. Immediately cancel it

*Actual:*
Intermittently, the cancellation token is ignored and the cancellation is never passed to the server

*Expected:*
The server method is cancelled

*Root Cause:*
[ReadAndHandleRequestsAsync()](https://github.com/Microsoft/vs-streamjsonrpc/blob/2351dd787be3fae550a36c3b3fdcfc9864f4bda9/src/StreamJsonRpc/JsonRpc.cs#L935) kicks off each request on its own thread, and the cancellation request, though arrived second, may be handled faster than the method invocation, and may not find its id in ```inboundCancellationSources```, and won't be able to cancel it.

IlyaBiryukov:
Handle incoming request sequentially

To ensure that cancellation request is processed after the method it is cancelling, we handle all incoming requests sequentilly, and only yield when we execute user code.

Fixes #56

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -755,10 +755,9 @@ namespace StreamJsonRpc
             return JsonRpcMessage.CreateError(id, JsonRpcErrorCode.InvocationError, exception.Message, JObject.FromObject(data, DefaultJsonSerializer));
         }
 
-        private async Task<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcMessage request, JsonSerializer jsonSerializer)
+        private async Task<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcMessage request)
         {
             Requires.NotNull(request, nameof(request));
-            Requires.NotNull(jsonSerializer, nameof(jsonSerializer));
 
             if (this.targetRequestMethodToClrMethodMap.Count == 0)
             {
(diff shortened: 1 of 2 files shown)


AArnott:
I suspect it may be worth rebasing these commits before merging so that they could also be included in a v1.1 branch in case we need to service VS 15.3 or 15.5.

IlyaBiryukov:
Rebased off master, @AArnot, if you can approve, I'll proceed with the insertion

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -796,6 +798,9 @@ private static JsonRpcMessage CreateError(JToken id, Exception exception)
                     }
                 }
 
+                // Yield now so method invocation is async and we can proceed to handle other requests meanwhile
+                await Task.Yield();
```
AArnott: It looks like `await Task.Yield();` unconditionally captures the context of the caller. It appears you are introducing the first uses of it to this library, so you *could* be introducing deadlocks if any caller were to use `Task.Wait()` on the Task we return while on the UI thread. 
I can't find any immediate replacement for `Task.Yield()` that doesn't capture the `SynchronizationContext` (surprisingly, it appears there isn't one). We can add a private one and use it here, or we could [add one to the vs-threading library](https://github.com/Microsoft/vs-threading/issues/181).

IlyaBiryukov: I know that Task.Yield() captures the context of the caller. In this case though it should be fine because the caller runs in a task on a background thread, started with ```Task.Run(this.ReadAndHandleRequestsAsync, this.disposeCts.Token);``` in [StartListening()](https://github.com/Microsoft/vs-streamjsonrpc/blob/master/src/StreamJsonRpc/JsonRpc.cs#L256)

PR #59, opened 2017-09-06
Add support for individual RPC method registration

### LINKED ISSUE #49: Suggestion: prefix-based server routing
Sometimes multiple json rpc servers have to share the same stream. If we want to support multiplexing of the servers, we should be able to distinguish them. One way to do that is to have a prefix in the method name. On json rpc level, server A with method Foo will have "A.Foo" method, and server B - "B.Foo". 

StreamJsonRpc servers then can add multiple targets and supply prefix for each one for the routing.

AArnott:
This lets an endpoint register individual methods or delegates to RPC method names.
This complements the existing feature of being able to register target objects where all public methods are automatically registered.

While at it, I had to revise the way we record RPC methods. Now we have a dictionary that translates the name of an RPC method to a list of all MethodInfos (and target objects) that may respond to that method. It's more than one since there can be overloads.

Closes #49

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -22,11 +22,11 @@ public class JsonRpcTests : TestBase
     private const string HubName = "TestHub";
 
     private readonly Server server;
-    private readonly Stream serverStream;
-    private readonly JsonRpc serverRpc;
+    private Stream serverStream;
+    private JsonRpc serverRpc;
 
-    private readonly Stream clientStream;
-    private readonly JsonRpc clientRpc;
+    private Stream clientStream;
+    private JsonRpc clientRpc;
 
     public JsonRpcTests(ITestOutputHelper logger)
         : base(logger)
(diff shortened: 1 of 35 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1124,6 +1216,38 @@ private Task TransmitAsync(JsonRpcMessage message, CancellationToken cancellatio
             return this.MessageHandler.WriteAsync(json, cancellationToken);
         }
 
+        internal struct MethodTarget : IEquatable<MethodTarget>
```
AArnott: `MethodTarget` makes sense given its content, but we already have a `TargetMethod` type, so that's potentially very confusing. Any suggestions for renaming either type?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -682,6 +829,22 @@ private static void SendObject(Stream receivingStream, object jsonObject)
         receivingStream.Write(buffer, 0, buffer.Length);
     }
 
+    private void ReinitializeRpcWithoutListening()
```
IlyaBiryukov: Regular initialization can now call into ReinitializeRpcWithoutListening() followed by StartListening()

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -690,9 +769,18 @@ protected virtual Task<TResult> InvokeCoreAsync<TResult>(int? id, string targetN
                     }
                     else
                     {
-                        requestMethodToClrMethodMap.Add(requestName, method.Name);
+                        requestMethodToClrMethodNameMap.Add(requestName, method.Name);
+                    }
+
+                    // Skip this method if its signature matches one from a derived type we have already scanned.
```
IlyaBiryukov: Do we need a unit test to cover that scenario?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -234,13 +236,83 @@ public static JsonRpc Attach(Stream sendingStream, Stream receivingStream, objec
...
+        /// It need not match the name of the CLR method/delegate given here.
+        /// </param>
+        /// <param name="handler">
+        /// The method or delegate to invoke when a matching RPC message arrives.
+        /// This method may accept parameters from the incoming JSON-RPC message.
+        /// </param>
+        /// <param name="target">An instance of the type that defines <paramref name="handler"/> which should handle the invocation.</param>
+        public void AddLocalRpcMethod(string rpcMethodName, MethodInfo handler, object target)
+        {
+            Verify.Operation(!this.startedListening, Resources.AttachTargetAfterStartListeningError);
```
IlyaBiryukov: This will defeat dynamic light up of new methods which may need to occur after we've started listening.

src/StreamJsonRpc/Reflection/MethodSignatureAndTarget.cs
```diff
@@ -0,0 +1,41 @@
...
+        public override bool Equals(object obj)
+        {
+            return obj is MethodSignatureAndTarget other
+                && this.Equals(other);
+        }
+
+        public bool Equals(MethodSignatureAndTarget other)
+        {
+            return this.Signature.Equals(other.Signature)
+                && object.Equals(this.Target, other.Target);
```
IlyaBiryukov: Should we use ReferenceEquals here? We probably don't want to call Equals on target objects.

src/StreamJsonRpc/Reflection/MethodSignatureAndTarget.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public override int GetHashCode()
+        {
+            return this.Signature.GetHashCode() + (this.Target?.GetHashCode() ?? 0);
```
IlyaBiryukov: Same comment as for Equals method - we should have hash code that would match ReferenceEquals, not target's Equals

src/StreamJsonRpc.Tests/JsonRpcTests.cs
AArnott: True. But then we wouldn't be testing the `Attach` method.

src/StreamJsonRpc/JsonRpc.cs
AArnott: We do. Code coverage shows that a great many tests cover this code.

src/StreamJsonRpc/Reflection/MethodSignatureAndTarget.cs
AArnott: Although we have test coverage here, there is no way the product (at present) will actually find a different Target object in this method (since we never invoke it with a different one). I've made the correction you suggest since I agree it's accurate. It's just unfortunate we can't test it. The alternative would be to throw in the currently untestable case, but since we are pretty confident about what the correct behavior is, and it's cheap, I'm going with your recommendation.

src/StreamJsonRpc/Reflection/MethodSignatureAndTarget.cs
AArnott: Fixed.

src/StreamJsonRpc/JsonRpc.cs
AArnott: True, but that's consistent with behavior before my change since the `AddLocalRpcTarget` method already did this.

IIRC we talked about how we want to enable this scenario, and that's exactly why I added a lock around access to the dictionary we mutate in this method. But actually enabling the scenario deserves a PR of its own, IMO. IIRC it will require a new property or ctor parameter to opt into the feature, since we want to avoid folks unknowingly using `Attach` and then adding target objects/methods and not realizing that they're introducing a race condition for incoming traffic.

AArnott: Opened #62 to track.

PR #61, opened 2017-09-17
Fix invoking RPC methods whose only parameter is a CancellationToken

### LINKED ISSUE #60: Cannot invoke a method that contains just the CancellationToken parameter
I have a method with a following signature: `public void Method(CancellationToken token) => DoSomething();`.

I tried to invoke this method with cancellation and pass an empty parameter array. However, this results in an exception:

```
StreamJsonRpc.RemoteInvocationException: Parameter count mismatch.
   at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
   at StreamJsonRpc.JsonRpc.<InvokeCoreAsync>d__63`1.MoveNext()
   at 
```
After debugging this I noticed that `JsonRpcMessage.GetParameters` method returns an empty object array for parameters (instead of an array with one item left for the cancellation token) and as a result invoking the method throws the exception.

I also tried to pass null or the cancellation token as a parameter, but this failed the check that validates the parameter count in `TargetMethod.TryGetParameters` (`if (paramCount < method.RequiredParamCount || paramCount > method.TotalParamCountExcludingCancellationToken)`).

AArnott:
Fixes #60

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -418,6 +418,33 @@ public class JsonRpcTests : TestBase
     }
 
     [Fact]
+    public async Task InvokeAsync_CanCallCancellableMethodWithNoArgs()
+    {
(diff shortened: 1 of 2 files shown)


AArnott:
Thanks, @IlyaBiryukov

PR #66, opened 2017-10-13
Adding docs for StreamJsonRpc

tinaschrepfer:
Adding an initial page for docs and sample usage for trivial cases


javierdlg:
Should we potentially start working on the wiki page as well?

AArnott:
@javierdlg I'd rather keep docs in git rather than the wiki so it can version with the source code it describes.

PR #68, opened 2017-11-20
Avoid loading an assembly that defines parameter types unless necessary

AArnott:
This fixes a perf regression since v1.3.6 because a local RPC target included public methods that were never invoked, but had a parameter whose type was defined in an assembly that did not need to load.


AArnott:
@javierdlg: this will improve performance by *not* loading the DLL, since we are no longer calling `GetParameters()` when we don't need to.

PR #72, opened 2017-11-24
Improve performance of transmitting multiple messages

### LINKED ISSUE #37: High lock contention in Roslyn DelimitedMessageHandler
From a discussion with @AArnott ,

I'm seeing super high CPU utilization in VS when we use StreamRPC to talk to our OOP server.  ETW logs reveal this as the top exclusive problem:

I’m investigating some local traces that show high CPU usage in VS during OOP scenarios.  The traces end up showing a large amount of time around locks and contention:

--

 
image

That’s 18% exclusive time in VS time just around these locking primitives.  Looking into what calls this I can see the following:

image

And

image

### LINKED ISSUE #52: Read- and Write- buffering stream classes are not provided/inaccessible
While it is possible to implement own transport protocol by extending DelimitedMessageHandler, the class ReadBufferingStream is made internal to the library.
Moreover, write buffering class would be welcome in order to increase efficiency when using network streams (like SslStream).

Writing and flushing each message separately, like in HeaderDelimitedMessageHandler.cs lines 240-242 is suboptimal, especially for large quantity of small messages.
~~~ C#
            // Transmit the content itself.
            await this.SendingStream.WriteAsync(contentBytes, 0, contentBytes.Length, cancellationToken).ConfigureAwait(false);
            await this.SendingStream.FlushAsync().ConfigureAwait(false);
~~~
_Note:_ Simply using BufferedStream does not solve the whole problem, as flushing it per each call will introduce the same inefficiency and not-flushing it right away brings the question where and when to do it.

AArnott:
This change adjusts the timing of flushing the stream and how quickly we exit the sending semaphore, allowing multiple message transmissions to line up right next to each other rather than being necessarily spaced apart for the entire time required to flush the previous message.

Fixes #52 
Probably also fixes #37 because we will hold the semaphore for a much smaller period of time.

This doesn't seem to significantly impact the "chatty" perf test we had before (or it makes it slightly slower). I added a new perf test for a burst of "notify" messages with these demonstrably improved results (*doubles* throughput):

```
BEFORE

Test Name:	PerfTests.BurstNotifyMessages(serverStream: FullDuplexStream { BeforeWrite = null, CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, ... }, clientStream: FullDuplexStream { BeforeWrite = null, CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, ... })
Test Outcome:	Passed
Result StandardOutput:	
10000 iterations completed in 418 ms.
Rate: 23875.9159099792 invocations per second.
Overhead: 0.0418 ms per invocation.

Test Name:	PerfTests.BurstNotifyMessages(serverStream: NamedPipeServerStream { CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, InBufferSize = 0, ... }, clientStream: NamedPipeClientStream { CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, InBufferSize = 0, ... })
Test Outcome:	Passed
Result StandardOutput:	
10000 iterations completed in 985 ms.
Rate: 10147.2209552289 invocations per second.
Overhead: 0.0985 ms per invocation.



AFTER

Test Name:	PerfTests.BurstNotifyMessages(serverStream: FullDuplexStream { BeforeWrite = null, CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, ... }, clientStream: FullDuplexStream { BeforeWrite = null, CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, ... })
Test Outcome:	Passed
Result StandardOutput:	
10000 iterations completed in 233 ms.
Rate: 42809.88685775 invocations per second.
Overhead: 0.0233 ms per invocation.

Test Name:	PerfTests.BurstNotifyMessages(serverStream: NamedPipeServerStream { CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, InBufferSize = 0, ... }, clientStream: NamedPipeClientStream { CanRead = True, CanSeek = False, CanTimeout = False, CanWrite = True, InBufferSize = 0, ... })
Test Outcome:	Passed
Result StandardOutput:	
10000 iterations completed in 499 ms.
Rate: 20016.477564331 invocations per second.
Overhead: 0.0499 ms per invocation.
```

### DIFF
--- a/src/StreamJsonRpc.Tests/DelimitedMessageHandlerTests.cs
+++ b/src/StreamJsonRpc.Tests/DelimitedMessageHandlerTests.cs
@@ -112,27 +112,20 @@ public class DelimitedMessageHandlerTests : TestBase
     }
 
     /// <summary>
-    /// Tests that by default, WriteCoreAsync calls cannot be called again till
+    /// Tests that WriteCoreAsync calls cannot be called again till
     /// the Task returned from the prior call has completed.
     /// </summary>
-    /// <param name="optimized">A value indicating whether we simulate a derived type that expresses readiness for a perf optimization.</param>
-    [Theory]
-    [InlineData(false)]
-    [InlineData(true)]
-    public async Task WriteAsync_SemaphoreIncludesWriteCoreAsync_Task(bool optimized)
+    [Fact]
+    public async Task WriteAsync_SemaphoreIncludesWriteCoreAsync_Task()
     {
         var handler = new DelayedWriter(this.sendingStream, this.receivingStream, Encoding.UTF8);
-        handler.WriteCoreAsyncCallsWriteBeforeReturningPublic = optimized;
         var writeTask = handler.WriteAsync("content", CancellationToken.None);
         var write2Task = handler.WriteAsync("content", CancellationToken.None);
 
-        if (!optimized)
-        {
-            // Give the library extra time to do the wrong thing asynchronously.
-            await Task.Delay(ExpectedTimeout);
-        }
+        // Give the library extra time to do the wrong thing asynchronously.
+        await Task.Delay(ExpectedTimeout);
 
-        Assert.Equal(optimized ? 2 : 1, handler.WriteCoreCallCount);
+        Assert.Equal(1, handler.WriteCoreCallCount);
         handler.WriteBlock.Set();
         await Task.WhenAll(writeTask, write2Task).WithTimeout(UnexpectedTimeout);
         Assert.Equal(2, handler.WriteCoreCallCount);
@@ -185,10 +178,6 @@ public class DelimitedMessageHandlerTests : TestBase
             : base(sendingStream, receivingStream, encoding)
         {
         }
-
-        internal bool WriteCoreAsyncCallsWriteBeforeReturningPublic { get; set; }
-
-        protected override bool WriteCoreAsyncCallsWriteBeforeReturning => this.WriteCoreAsyncCallsWriteBeforeReturningPublic;
 
         protected override Task<string> ReadCoreAsync(CancellationToken cancellationToken)
         {
--- a/src/StreamJsonRpc.Tests/DelimitedMessageHandlerTests.cs
+++ b/src/StreamJsonRpc.Tests/DelimitedMessageHandlerTests.cs
@@ -111,6 +111,33 @@ public class DelimitedMessageHandlerTests : TestBase
         await Assert.ThrowsAnyAsync<OperationCanceledException>(() => writeTask);
     }
 
+    /// <summary>
+    /// Tests that by default, WriteCoreAsync calls cannot be called again till
+    /// the Task returned from the prior call has completed.
+    /// </summary>
+    /// <param name="optimized">A value indicating whether we simulate a derived type that expresses readiness for a perf optimization.</param>
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
+    public async Task WriteAsync_SemaphoreIncludesWriteCoreAsync_Task(bool optimized)
+    {
+        var handler = new DelayedWriter(this.sendingStream, this.receivingStream, Encoding.UTF8);
+        handler.WriteCoreAsyncCallsWriteBeforeReturningPublic = optimized;
+        var writeTask = handler.WriteAsync("content", CancellationToken.None);
+        var write2Task = handler.WriteAsync("content", CancellationToken.None);
+
+        if (!optimized)
+        {
+            // Give the library extra time to do the wrong thing asynchronously.
+            await Task.Delay(ExpectedTimeout);
+        }
+
+        Assert.Equal(optimized ? 2 : 1, handler.WriteCoreCallCount);
+        handler.WriteBlock.Set();
+        await Task.WhenAll(writeTask, write2Task).WithTimeout(UnexpectedTimeout);
+        Assert.Equal(2, handler.WriteCoreCallCount);
+    }
+
     [Fact]
     public void ReadAsync_ThrowsObjectDisposedException()
     {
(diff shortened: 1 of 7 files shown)


AArnott:
This will likely be of interest to @heejaechang, @CyrusNajmabadi, and @PeterTruchly. Possibly @sharwell too.

src/StreamJsonRpc.Tests/DelimitedMessageHandlerTests.cs
```diff
@@ -111,6 +111,33 @@ public void WriteAsync_PreferOperationCanceledException_AtEntry()
...
+    public async Task WriteAsync_SemaphoreIncludesWriteCoreAsync_Task(bool optimized)
+    {
+        var handler = new DelayedWriter(this.sendingStream, this.receivingStream, Encoding.UTF8);
+        handler.WriteCoreAsyncCallsWriteBeforeReturningPublic = optimized;
+        var writeTask = handler.WriteAsync("content", CancellationToken.None);
+        var write2Task = handler.WriteAsync("content", CancellationToken.None);
+
+        if (!optimized)
+        {
+            // Give the library extra time to do the wrong thing asynchronously.
```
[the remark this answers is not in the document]
AArnott: The thing I had in mind was that it might invoke my `WriteCoreAsync` method twice when it's only expected to happen once. I want to give it time to have called it twice before claiming success that it only called it once.

src/StreamJsonRpc.Tests/DelimitedMessageHandlerTests.cs
```diff
@@ -152,18 +179,25 @@ private class DelayedWriter : DelimitedMessageHandler
...
         internal readonly AsyncManualResetEvent WriteBlock = new AsyncManualResetEvent();
 
+        internal int WriteCoreCallCount;
+
         public DelayedWriter(Stream sendingStream, Stream receivingStream, Encoding encoding)
             : base(sendingStream, receivingStream, encoding)
         {
         }
 
+        internal bool WriteCoreAsyncCallsWriteBeforeReturningPublic { get; set; }
```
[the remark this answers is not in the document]
AArnott: Yes, it's because I expose what otherwise would only be `protected`. It's just a test project, so whether it's internal vs. public is immaterial. And adding "Internal" to the name doesn't IMO sound like I'm _expanding_ visibility. It's cuts either way, but I figured this was a bit more clearer to my intention.

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
```diff
@@ -200,7 +203,7 @@ private enum HeaderParseState
         }
 
         /// <inheritdoc />
-        protected override async Task WriteCoreAsync(string content, Encoding contentEncoding, CancellationToken cancellationToken)
+        protected override Task WriteCoreAsync(string content, Encoding contentEncoding, CancellationToken cancellationToken)
         {
             var sendingBufferStream = new MemoryStream(MaxHeaderElementSize);
```
heejaechang: consider using pool such as https://github.com/Microsoft/Microsoft.IO.RecyclableMemoryStream or Roslyn's

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
```diff
@@ -241,12 +244,13 @@ private enum HeaderParseState
             // Transmit the headers.
             // Ignore the cancellation token so we don't write the header without the content.
             sendingBufferStream.Position = 0;
-            await sendingBufferStream.CopyToAsync(this.SendingStream, MaxHeaderElementSize).ConfigureAwait(false);
+            Task headersTask = sendingBufferStream.CopyToAsync(this.SendingStream, MaxHeaderElementSize);
```
heejaechang: is this okay to write to the stream concurrently? shouldn't headertask and contenttask needed to be continueWith? what guarantee ordering of headerTask and contentTask. contentTask can run before headerTask.

heejaechang: so a caller of this method acquires a lock to guard the stream from being messed up from being called concurrently? 

so, this method guarantee that it only return once message is written to the stream, but it doesn't guarantee that the other side actually got the message? if so, the guarantee that the message actually written to the stream has no significant value to the writer, I think. if so, how about changing design that. this method just write message to pooled stream (such as memory stream you have above) and enqueue that message (stream) to a queue (FIFO queue)
and return. and one dedicated thread or task, empty that queue. that way you basically serialize all message and 1 thread/task completely own the stream (so no messing up)

AArnott: Well, I think you have highlighted a problem: if the header is larger than `MaxHeaderElementSize` then `CopyToAsync` may not be done calling `this.SendingStream.WriteAsync` for all blocks before our `WriteAsync(contentBytes)` call. So we do have a problem here.

AArnott: We already have a FIFO queue (by virtue of the `AsyncSemaphore`). The Task returned by this method indicates when the sending stream has been completely written to.

heejaechang: I thought you were trying to improve throughput. using async (either async lock or await or task) can cause big gap between continuation I believe. especially when there are a lot of pending tasks.

using explicit queue + synchronous lock (or concurrent queue) + explicit loop will give you fastest throughput I believe. at the cost of 100% CPU of 1 core if there are a lot of pending messages.

AArnott: I'm trying to improve throughput by removing the requirement to flush while holding the semaphore. We have very significant gains by doing just that.
A dedicated thread that sleeps a lot still waits for the kernel to wake it up. Being async doesn't require the kernel to get involved and a single threadpool thread can go through tasks faster than the kernel can suspend and resume various threads. At least in some contexts async can be faster. But doubling throughput satisfies me for this PR anyway. :) 
Besides, it turns out there are a *lot* of these ServiceHub streams in VS nowadays, and a dedicated thread for each one of them would get a bit much, IMO.

PeterTruchly: I am glad that You already see performance improvements. I am sorry for not being able to provide my sample/test code at the moment, but the one I have integrates with non-public libs.

>A dedicated thread that sleeps a lot still waits for the kernel to wake it up. Being async doesn't require the kernel to get involved and a single threadpool thread can go through tasks faster than the kernel can suspend and resume various threads.

I agree, additional thread could be also a problem when many connections are to be handled using separate (session related/isolated) instances...

One 'idea' would be to mimic behavior of single thread handling the queue using tasks and making sure there is always at least 1 task planned when the action is necessary.

I will definitely evaluate this improved version (currently I use 1.1.102) and I will get back with findings, but no guarantees when this will happen...

PR #78, opened 2017-12-27
Add DelimitedMessageHandler.FlushCoreAsync() method

AArnott:
This further abstracts the use of Streams in `DelimitedMessageHandler` so that `WebSocket` can be used instead of Stream as the transport.

### DIFF
--- a/src/StreamJsonRpc/DelimitedMessageHandler.cs
+++ b/src/StreamJsonRpc/DelimitedMessageHandler.cs
@@ -158,7 +158,7 @@ namespace StreamJsonRpc
                         await this.WriteCoreAsync(content, contentEncoding, cts.Token).ConfigureAwait(false);
                     }
 
-                    await this.SendingStream.FlushAsync().ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


tomlm:
>         protected DelimitedMessageHandler(Stream sendingStream, Stream receivingStream, Encoding encoding)

consider adding DelimitedMessageHandler(Encoding encoding) : this(Stream.Null, Stream.Null, encoding) {} 
??

---
Refers to: src/StreamJsonRpc/DelimitedMessageHandler.cs:51 in 84beb23. [](commit_id = 84beb23cf0cd933c4a96b9bd9e56862a5d7d1b14, deletion_comment = False)

tomlm:
>             this.ReceivingStream = receivingStream;

because this is still checking for stream existence.

---
Refers to: src/StreamJsonRpc/DelimitedMessageHandler.cs:58 in 84beb23. [](commit_id = 84beb23cf0cd933c4a96b9bd9e56862a5d7d1b14, deletion_comment = False)

AArnott:
Thanks, @tomlm. I trust #81 should resolve your concerns.

PR #81, opened 2017-12-30
Add support for WebSocket

AArnott:
Adds the `WebSocketMessageHandler` class to support the `WebSocket` class in .NET so this library can be used on a web server. `WebSocket` doesn't employ standard .NET `Stream` so it was necessary to add special support for it. This could have been done outside the library as well, but it seems like a reasonably common use case so this change makes it an included part of the library.

Note I take special care to handle messages that exceed the buffer size.

Tests included.

### DIFF
--- a/src/StreamJsonRpc/WebSocketMessageHandler.cs
+++ b/src/StreamJsonRpc/WebSocketMessageHandler.cs
@@ -1,0 +1,137 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if WEBSOCKETS
+
+namespace StreamJsonRpc
+{
+    using System;
+    using System.IO;
+    using System.Net.WebSockets;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft;
+
+    /// <summary>
+    /// A message handler for the <see cref="JsonRpc"/> class
+    /// that uses <see cref="System.Net.WebSockets.WebSocket"/> as the transport.
+    /// </summary>
+    public class WebSocketMessageHandler : DelimitedMessageHandler
+    {
+        private readonly ArraySegment<byte> readBuffer;
+        private readonly byte[] writeBuffer;
+        private Decoder readDecoder;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WebSocketMessageHandler"/> class.
+        /// </summary>
+        /// <param name="webSocket">The <see cref="System.Net.WebSockets.WebSocket"/> used to communicate.</param>
+        /// <param name="bufferSize">
+        /// The size of the buffer to use for reading JSON-RPC messages.
+        /// Messages which exceed this size will be handled properly but may require multiple I/O operations.
+        /// </param>
+        public WebSocketMessageHandler(WebSocket webSocket, int bufferSize = 4096)
+            : base(Stream.Null, Stream.Null, Encoding.UTF8)
+        {
+            Requires.NotNull(webSocket, nameof(webSocket));
+            Requires.Range(bufferSize > 0, nameof(bufferSize));
+
+            this.WebSocket = webSocket;
+            this.readBuffer = new ArraySegment<byte>(new byte[bufferSize]);
+            this.writeBuffer = new byte[bufferSize];
+        }
+
+        /// <summary>
+        /// Gets the <see cref="System.Net.WebSockets.WebSocket"/> used to communicate.
+        /// </summary>
+        public WebSocket WebSocket { get; }
+
+        /// <inheritdoc />
+        protected async override Task<string> ReadCoreAsync(CancellationToken cancellationToken)
+        {
+            WebSocketReceiveResult result = await this.WebSocket.ReceiveAsync(this.readBuffer, cancellationToken).ConfigureAwait(false);
+            if (result.EndOfMessage)
+            {
+                // fast path: the entire message fit within the buffer.
+                return this.Encoding.GetString(this.readBuffer.Array, 0, result.Count);
+            }
+            else
+            {
+                // The message exceeds the size of our buffer.
+                if (this.readDecoder == null)
+                {
+                    this.readDecoder = this.Encoding.GetDecoder();
+                }
+
+                var jsonBuilder = new StringBuilder();
+                char[] decodedChars = new char[this.readBuffer.Array.Length];
+                void DecodeInput()
+                {
+                    int decodedCharsCount = this.readDecoder.GetChars(this.readBuffer.Array, 0, result.Count, decodedChars, 0, result.EndOfMessage);
+                    jsonBuilder.Append(decodedChars, 0, decodedCharsCount);
+                }
+
+                DecodeInput();
+                while (!result.EndOfMessage)
+                {
+                    result = await this.WebSocket.ReceiveAsync(this.readBuffer, cancellationToken).ConfigureAwait(false);
+                    DecodeInput();
+                }
+
+                return jsonBuilder.ToString();
+            }
+        }
+
+        /// <inheritdoc />
+        protected async override Task WriteCoreAsync(string content, Encoding contentEncoding, CancellationToken cancellationToken)
+        {
+            Requires.NotNull(content, nameof(content));
+            Requires.NotNull(contentEncoding, nameof(contentEncoding));
+
+            if (contentEncoding.GetByteCount(content) <= this.writeBuffer.Length)
+            {
+                // Fast path: send the whole message as a single chunk.
+                int bytesWritten = contentEncoding.GetBytes(content, 0, content.Length, this.writeBuffer, 0);
+                var bufferSegment = new ArraySegment<byte>(this.writeBuffer, 0, bytesWritten);
+                await this.WebSocket.SendAsync(bufferSegment, WebSocketMessageType.Text, true, cancellationToken).ConfigureAwait(false);
+            }
+            else
+            {
+                var encoder = contentEncoding.GetEncoder();
+                bool completed;
+                int bytesUsed;
+                int charsLeftToConvert = content.Length;
+                while (charsLeftToConvert > 0)
+                {
+                    unsafe
+                    {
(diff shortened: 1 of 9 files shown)


AArnott:
@tomlm Do you think disposing the `WebSocket` when the JsonRpc channel is shutdown is the right behavior? We could make it an option (later) if we need to. But if this is the right default behavior, I'd like to start with it being the only behavior until we get a request for *not* disposing the socket.

Also: I see that `WebSocket` includes special async close methods on it. So perhaps it's just a bad idea to dispose the WebSocket at all, rather to leave it to the web server to do it?

src/StreamJsonRpc/WebSocketMessageHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+                            char* pStart = pContent + content.Length - charsLeftToConvert;
+                            encoder.Convert(pStart, charsLeftToConvert, writeBuffer, this.writeBuffer.Length, false, out int charsUsed, out bytesUsed, out completed);
+                            charsLeftToConvert -= charsUsed;
+                        }
+                    }
+
+                    var bufferSegment = new ArraySegment<byte>(this.writeBuffer, 0, bytesUsed);
+                    await this.WebSocket.SendAsync(bufferSegment, WebSocketMessageType.Text, completed, cancellationToken).ConfigureAwait(false);
+                }
+            }
```
tomlm: huh, I didn't have to do this.  WebSocket seems to send whatever size message I give it...I did up to 50k sized messages without this logic..?
        {
            var buffer = Encoding.UTF8.GetBytes(content);
            await this.WebSocket.SendAsync(new ArraySegment<byte>(buffer), WebSocketMessageType.Text, true, cancellationToken)
                .ConfigureAwait(false);
        }

src/StreamJsonRpc/WebSocketMessageHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+                    this.readDecoder = this.Encoding.GetDecoder();
+                }
+
+                var jsonBuilder = new StringBuilder();
+                char[] decodedChars = new char[this.readBuffer.Array.Length];
+                void DecodeInput()
+                {
+                    int decodedCharsCount = this.readDecoder.GetChars(this.readBuffer.Array, 0, result.Count, decodedChars, 0, result.EndOfMessage);
+                    jsonBuilder.Append(decodedChars, 0, decodedCharsCount);
+                }
```
tomlm: I assume you have to do this extra jiggery-pokey to handle arbitrary encodings being across package boundaries?

src/StreamJsonRpc/WebSocketMessageHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+
+                DecodeInput();
+                while (!result.EndOfMessage)
+                {
+                    result = await this.WebSocket.ReceiveAsync(this.readBuffer, cancellationToken).ConfigureAwait(false);
+                    DecodeInput();
+                }
```
tomlm: why not just read all bytes[] and then do single decoding?

src/StreamJsonRpc/WebSocketMessageHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+                void DecodeInput()
+                {
+                    int decodedCharsCount = this.readDecoder.GetChars(this.readBuffer.Array, 0, result.Count, decodedChars, 0, result.EndOfMessage);
+                    jsonBuilder.Append(decodedChars, 0, decodedCharsCount);
+                }
+
+                DecodeInput();
+                while (!result.EndOfMessage)
+                {
+                    result = await this.WebSocket.ReceiveAsync(this.readBuffer, cancellationToken).ConfigureAwait(false);
```
tomlm: >(this.readBuffer [](start = 62, length = 16)

are you guarenteed that 2 people won't call ReceiveAsync() on different threads?  Can you share the buffer like this?

src/StreamJsonRpc/WebSocketMessageHandler.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// <inheritdoc />
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                this.WebSocket.Dispose();
+            }
+
+            base.Dispose(disposing);
+        }
```
tomlm: Hmmm...I think this is a good idea...

src/StreamJsonRpc/WebSocketMessageHandler.cs
AArnott: Actually your simpler approach is buggy, since it will fail if the incoming message exceeds `this.bufferSize`, since you never grow your buffer.

AArnott: Once you fix that to grow your buffer, you find that you're repeatedly allocating increasingly large buffers and copying the last one into it until you finally find the end size. To avoid that (since that entails GC pressure and a lot of lost processor time), I prefer to reuse the original buffer, and incrementally decode as we go. It should still be super-fast in the normal case, and in the large message case it should keep the throughput high better than just growing the buffer and copying each time.

src/StreamJsonRpc/WebSocketMessageHandler.cs
AArnott: Yup, yours is simpler. But it requires allocating a new buffer for each and every transmitted message. My approach reuses a buffer for each message. That reduces GC pressure and thus should promote high throughput when you experience memory pressure.

src/StreamJsonRpc/WebSocketMessageHandler.cs
AArnott: Do you mean you support disposing the `WebSocket` by default? What do you think of the issue with needing an async handshake to disconnect though, per my first comment on the PR to you?

src/StreamJsonRpc/WebSocketMessageHandler.cs
tomlm: yeah, I realized that after talking to you.  This uses 1 buffer, but increases size of it if need be:
 public class WebSocketMessageHandler : DelimitedMessageHandler, IDisposable
    {
        private int bufferSize;
        private byte[] readBuffer;

        public WebSocketMessageHandler(WebSocket webSocket, int bufferSize = 32768) : base(Stream.Null, Stream.Null, Encoding.UTF8)
        {
            this.bufferSize = bufferSize;
            this.WebSocket = webSocket;
            readBuffer = new byte[bufferSize];
        }

        public WebSocket WebSocket { get; set; }

        protected async override Task<string> ReadCoreAsync(CancellationToken cancellationToken)
        {
            WebSocketReceiveResult result = null;
            int nBytes = 0;
            do
            {
                result = await WebSocket.ReceiveAsync(new ArraySegment<byte>(this.readBuffer, nBytes, this.readBuffer.Length - nBytes), cancellationToken).ConfigureAwait(false);
                nBytes += result.Count;

                // if next read will run out of space (typically reads are 4k reads, but this will be reactive to whatever size we are getting
                if (nBytes > (readBuffer.Length - result.Count))
                {
                    // on rare occasion of big message permentaly auto-grow our buffer size to handle big message size
                    Array.Resize<byte>(ref this.readBuffer, this.readBuffer.Length + this.bufferSize);
                }
            } while (!result.EndOfMessage);
            return this.Encoding.GetString(this.readBuffer, 0, nBytes);
        }

        protected async override Task WriteCoreAsync(string content, Encoding contentEncoding, CancellationToken cancellationToken)
        {
            var buffer = contentEncoding.GetBytes(content);
            await this.WebSocket.SendAsync(new ArraySegment<byte>(buffer), WebSocketMessageType.Text, true, cancellationToken)
                .ConfigureAwait(false);
        }

        protected override void Dispose(bool disposing)
        {
            if (disposing)
            {
                this.WebSocket.Dispose();
            }

            base.Dispose(disposing);
        }
    }

---
In reply to: [160270306](https://github.com/Microsoft/vs-streamjsonrpc/pull/81#discussion_r160270306) [](ancestors = 160270306)

src/StreamJsonRpc/WebSocketMessageHandler.cs
tomlm: I really am not expert on WebSocket semantics.  Do you have someone you can consult with more experience?

---
In reply to: [160271489](https://github.com/Microsoft/vs-streamjsonrpc/pull/81#discussion_r160271489) [](ancestors = 160271489)

tomlm:
Yeah, looking at this I think the hosting application needs to be responsible for the Close() semantics, because CloseAsync() takes status information which informs the other side as the reason of the closure.  That would be dependent on which application the JsonRPC would be executing, and so I think it is NOT part of the dispose pattern.  Make sense?

---
In reply to: [354568470](https://github.com/Microsoft/vs-streamjsonrpc/pull/81#issuecomment-354568470) [](ancestors = 354568470)

src/StreamJsonRpc/WebSocketMessageHandler.cs
AArnott: Yup: As I said, increasing the buffer size incrementally as you discover the message gets larger is an option. And yes, as you say we could "save" that larger buffer as a field. And yes, that risks one large message keeping a lot of memory allocated ever-after. That's all risk and higher runtime cost than is required if you are willing to perform the work a block at a time. I've written a lot of buffer block style code and I'm quite comfortable with writing and testing such code, so I prefer to do so in the most efficient way possible so I'll never get a bug claiming the library is the cause of their perf problems. :)

src/StreamJsonRpc/WebSocketMessageHandler.cs
AArnott: > unsafe buffers make me nervous
Well, that's why it's useful to have it written in a well-tested library so you can depend on it rather than own it yourself. :) Everything you do in .NET ultimately becomes "unsafe" at some level. In this case, I have to use unsafe because the encoder doesn't offer an overload that takes `byte[]` and an offset, so the only way to get buffer offset behavior is to use pointers. 

> why not increase the write buffer to size and then do simple write

Reliable perf and high throughput.

PR #85, opened 2018-01-12
increase build timeout to 240 min

jepetty:


jepetty:
Updated yml and confirmed can build https://devdiv.visualstudio.com/DevDiv/Connected%20Experience/_build/index?buildId=1285454&_a=summary

PR #90, opened 2018-02-04
Allow JsonRpc modification while listening

### LINKED ISSUE #62: Allow adding target objects or methods after listening has started
I believe we should require a new property or ctor parameter to opt into the feature, since we want to avoid folks unknowingly using Attach and then adding target objects/methods and not realizing that they're introducing a race condition for incoming traffic.

AArnott:
Note that enabling this requires thread-safety in configuration modifications while receiving messages concurrently. This is already implemented before this change.

Closes #62

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -985,6 +985,13 @@ public class JsonRpcTests : TestBase
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


AArnott:
@jepetty Thanks for the review. Were you thinking the `<remarks>` section itself was not worded strongly enough, or rather than the same warning ought to be found elsewhere like the markdown docs?
I have another PR that builds on this, so I'm going to complete this. But I'm happy to follow up with more doc changes that you recommend.

Issue #95, opened 2018-04-16
Proxy generator for interfaces


AArnott:
Consider offering a code generator to produce proxy stubs automatically given an interface. We might support _events_ on that interface as a way to express callbacks (via notify messages) to the client. 

Cascade does this and likes it.

PR #100, opened 2018-05-09
Add support for RPC method name transforms when adding a local target objects

AArnott:
This allows for patterns already in use by the language server protocol and others where multiple target objects are added to the server side, with logical boundaries between these objects observable from the client by virtue of a 'namespace' being added to each method. Currently this is achievable by adding `JsonRpcMethodAttribute` to each method to change its name. This enhancement eliminates the need for manually transforming each individual method name.

It also gives the ability to do other interesting RPC method name transforms such as converting the typical PascalCase C# methods to be exposed with camelCase.

The method name transform is exposed as a `Func<string, string>` for maximum flexibility. But I add a `CommonMethodNameTransforms` static class with these two use cases (namespaced methods and camelCase) predefined as reusable functions.

### DIFF
--- a/src/StreamJsonRpc/CommonMethodNameTransforms.cs
+++ b/src/StreamJsonRpc/CommonMethodNameTransforms.cs
@@ -12,7 +12,7 @@ namespace StreamJsonRpc
     public static class CommonMethodNameTransforms
     {
         /// <summary>
-        /// Gets a function that converts a given string's first character to lowercase.
+        /// Gets a function that converts a given string from PascalCase to camelCase.
         /// </summary>
         public static Func<string, string> CamelCase
         {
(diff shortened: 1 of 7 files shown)


AArnott:
@jasongin would this change partially close the gap between StreamJsonRpc and what you're doing?

jasongin:
A little bit, yes. If I understand correctly, this approach is intending to avoid dictating any specific structure for the method namespacing. In our codebase we have built a lot around the concept of named "RPC services" which each have defined contract interfaces. I guess that could be layered on top the mechanism here.

src/StreamJsonRpc/CommonMethodNameTransforms.cs
```diff
@@ -0,0 +1,59 @@
...
+                    {
+                        throw new ArgumentNullException();
+                    }
+
+                    if (name.Length == 0)
+                    {
+                        return string.Empty;
+                    }
+
+                    return name.Substring(0, 1).ToLowerInvariant() + name.Substring(1);
```
ikeras: Because you're doing some allocations here, you may want to check if the first character is already lowercase. That avoids 3 allocations if it is.

ikeras: It's also worth noting that your algorithm for camelCase is different than what NewtonSoft uses with the camelCase converter. The main difference that I see is that the NewtonSoft version would lowercase all uppercase characters until the last one before a lowercase. So, for example, "HELLOWorld" would become "helloWorld" (instead of "hELLLOWorld"). Also, any fully uppercase string would be fully lowercase (i.e. "HELLO" would become "hello" instead of "hELLO").

AArnott:
@jasongin The point here was to allow, independent of the server class itself, for the person wiring it up to RPC to add a namespace to the methods (so that multiple server objects can be added yet distinct). This could be done before using `[JsonRpcMethodAttribute]` on each individual method to give it a namespace but that would be tedious and have to be compiled into each server type rather than be a runtime decision.

Can you tell me how interfaces overlap in concerns with this change? Your proxy generation is tracked by #95, which would be interface based.

src/StreamJsonRpc/CommonMethodNameTransforms.cs
AArnott: Regarding the allocations, you're right. However I considered that making the logic more complex with branching would add test cost and in the end the allocations would be insignificant. This function is only executed once per RPC method definition, not once per call, so it's a fixed cost, which is actually quite small relative to even one call at runtime. 

Interesting point about the camel casing. Certainly any changes to how this function works would result in a wire protocol breaking change, so whatever we check in will be locked in forever. I'll see if I can reuse Newtonsoft's implementation since they have more experience in this area than I do. Thanks for sharing.

jasongin:
> Can you tell me how interfaces overlap in concerns with this change?

There's no conflict with the approach in this PR; I was just pointing out that we use a consistent pattern for method name transforms, that corresponds to the way we represent named RPC services and call them via interfaces.

PR #103, opened 2018-05-11
Respect JsonRpcMethodAttribute appearing on interface methods

AArnott:
Previous to this change, a target object passed to StreamJsonRpc might look like this:

```cs
class Server
{
    public void Foo() { }

    [JsonRpcMethod("bar/soap")]
    public void Whatever() { }
}
```

The `JsonRpcMethodAttribute` that appears changes the name of the method over RPC from `Whatever` to `bar/soap` (which isn't a legal CLR name). 

But as we look at proxy generation based on interface, we find that these special name-changing attributes need to be able to appear on interfaces as well so that the client proxy can generate code that invokes the method by the proper name (`bar/soap`, in this case). So with this change, you can now do this:

```cs
interface IServer
{
    [JsonRpcMethod("bar/soap")]
    void Whatever();
}

class Server : IServer
{
    public void Foo() { }

    public void Whatever() { }
}
```

The `Server` object is still passed to the `JsonRpc` class as the target object and both methods are invokable, but the name change is inherited from the interface that `Whatever` happens to implement. But soon, when we can generate client proxies based on `IServer`, the generated client will know to invoke `bar/soap` instead of `Whatever`.

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -48,7 +48,7 @@ public class JsonRpcTests : TestBase
     private interface IServer
     {
         [JsonRpcMethod("AnotherName")]
-        string AManBy(string name);
+        string ARoseBy(string name);
 
         [JsonRpcMethod("IFaceNameForMethod")]
         int AddWithNameSubstitution(int a, int b);
@@ -1187,7 +1187,7 @@ public class JsonRpcTests : TestBase
     public async Task ServerRespondsWithMethodRenamedByInterfaceAttribute()
     {
         Assert.Equal("ANDREW", await this.clientRpc.InvokeAsync<string>("AnotherName", "andrew"));
-        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(IServer.AManBy), "andrew"));
+        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(IServer.ARoseBy), "andrew"));
     }
 #endif
 
@@ -1490,7 +1490,7 @@ public class JsonRpcTests : TestBase
             i = i + 1;
         }
 
-        public string AManBy(string name) => name.ToUpperInvariant();
+        public string ARoseBy(string name) => name.ToUpperInvariant();
 
         [JsonRpcMethod("ClassNameForMethod")]
         public int AddWithNameSubstitution(int a, int b) => a + b;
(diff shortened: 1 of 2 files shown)


src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -45,6 +45,15 @@ public JsonRpcTests(ITestOutputHelper logger)
         this.clientRpc = JsonRpc.Attach(this.clientStream);
     }
 
+    private interface IServer
+    {
+        [JsonRpcMethod("AnotherName")]
+        string AManBy(string name);
```
ikeras: nit: I would make it a "ARoseBy" :)

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -846,12 +846,14 @@ protected virtual Task<TResult> InvokeCoreAsync<TResult>(int? id, string targetN
...
 
+            var mapping = new MethodNameMap(target.GetType().GetTypeInfo());
+
             for (TypeInfo t = target.GetType().GetTypeInfo(); t != null && t != typeof(object).GetTypeInfo(); t = t.BaseType?.GetTypeInfo())
             {
                 foreach (MethodInfo method in t.DeclaredMethods)
                 {
-                    var attribute = (JsonRpcMethodAttribute)method.GetCustomAttribute(typeof(JsonRpcMethodAttribute));
-                    var requestName = attribute?.Name ?? method.Name;
+                    var attribute = mapping.FindAttribute(method);
```
ikeras: Should you keep this call, or should the place where you check attribute == null at the end of this method (for candidates), instead check to see if requestName == null?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
ikeras: Two quick thoughts: 1) don't see a test for invoking a member by it's non-attributed name, and the expected result (though could have missed it), and 2) didn't see a test for invoking an async method w/o async suffix for validating candidatemap

AArnott: Thanks. I guess I exposed a weakness in my would-be cultured education. I've fixed it to `ARoseBy`.

AArnott: There is indeed a negative test for calling a method by its CLR name. Here it is, with the correction you called for:

```cs
        await Assert.ThrowsAsync<RemoteMethodNotFoundException>(() => this.clientRpc.InvokeAsync(nameof(IServer.ARoseBy), "andrew"));
```

PR #104, opened 2018-05-14
Fix tests that fail on linux

AArnott:


AArnott:
CC: @dgriffen (github won't let me add more code reviewers right now due to its bug).

.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+git:
+  depth: false
+
+install:
+- nuget install xunit.runner.console -Version 2.2.0 -OutputDirectory testrunner
+- msbuild src/StreamJsonRpc.sln /nologo /m /v:quiet /t:restore
+
+script:
+- msbuild src/StreamJsonRpc.sln /nologo /m /v:minimal /t:build
+- dotnet test -f netcoreapp2.0 --no-build --no-restore src/StreamJsonRpc.Tests
```
dgriffen: Shouldn't we also run tests with mono here? Otherwise we're only testing .NET Core

AArnott: Supporting mono is beyond the scope of my initial goal with this PR. I already ran tests on mono locally on linux and they failed because mono doesn't support async pipes. Fixing that (if we want to) IMO merits a different PR.

dgriffen: Makes sense. I didn't realize mono was missing async pipe support.

PR #105, opened 2018-05-16
Add support for target object events

AArnott:
If the target object passed to `JsonRpc` defines events, event handlers will be added to these events and when raised, will automatically propagate to the remote party (i.e. the client) as JSON-RPC "notify" messages.

**Policy:** This is on by default. But users can opt out. Is that the right direction? Might there be server objects with events used for something else already, such that forwarding these events to the client (with args) may present a bug or security risk? I suspect the risk here is low, and I'd like for the feature to be on by default if possible.

As part of this I changed the way `AddLocalRpcTarget` accepts options so that instead of one option per parameter, it takes a single options object so that in the future, options like the opt-out and method name transform can be added to the object without having to add method overloads each time.

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -1498,7 +1498,18 @@ namespace StreamJsonRpc
                 this.jsonRpc = jsonRpc;
                 this.server = server;
                 this.eventInfo = eventInfo;
-                this.registeredHandler = OnEventRaisedMethodInfo.CreateDelegate(eventInfo.EventHandlerType, this);
+                try
+                {
+                    // This might throw if our EventHandler-modeled method doesn't "fit" the event delegate signature.
+                    // It will work for EventHandler and EventHandler<T>, at least.
+                    // If we want to support more, we'll likely have to use lightweight code-gen to generate a method
+                    // with the right signature.
+                    this.registeredHandler = OnEventRaisedMethodInfo.CreateDelegate(eventInfo.EventHandlerType, this);
+                }
+                catch (ArgumentException ex)
+                {
+                    throw new NotSupportedException("Unsupported event handler type for: " + eventInfo.Name, ex);
+                }
 
                 eventInfo.AddEventHandler(server, this.registeredHandler);
             }
@@ -1510,6 +1521,7 @@ namespace StreamJsonRpc
 
             private void OnEventRaised(object sender, EventArgs args)
             {
+                // We use null for the sender because we don't want to try to serialize the target object to the remote party.
                 this.jsonRpc.NotifyAsync(this.eventInfo.Name, new object[] { null, args });
             }
         }
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -321,13 +321,11 @@ namespace StreamJsonRpc
         /// should not inherit from each other and are invoked in the order which they are added.
         /// </summary>
         /// <param name="target">Target to invoke when incoming messages are received.</param>
-        /// <param name="methodNameTransform">
-        /// A function that takes the CLR method name and returns the RPC method name.
-        /// This method is useful for adding prefixes to all methods, or making them camelCased.
-        /// </param>
-        public void AddLocalRpcTarget(object target, Func<string, string> methodNameTransform)
+        /// <param name="options">A set of customizations for how the target object is registered. If <c>null</c>, default options will be used.</param>
+        public void AddLocalRpcTarget(object target, JsonRpcTargetOptions options)
         {
             Requires.NotNull(target, nameof(target));
+            options = options ?? JsonRpcTargetOptions.Default;
             this.ThrowIfConfigurationLocked();
 
             var mapping = GetRequestMethodToClrMethodMap(target);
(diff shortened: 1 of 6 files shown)


AArnott:
CC: @dgriffen  @ikeras

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -692,6 +708,16 @@ protected virtual void Dispose(bool disposing)
...
                 if (disposing)
                 {
+                    if (this.eventReceivers != null)
+                    {
+                        foreach (var receiver in this.eventReceivers)
+                        {
+                            receiver.Dispose();
+                        }
+
+                        this.eventReceivers.Clear();
```
ikeras: A little surprised you didn't set this to null instead of just clearing it

src/StreamJsonRpc/JsonRpcTargetOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// should be relayed to the client via a JSON-RPC notify message.
+        /// </summary>
+        /// <value>The default is <c>true</c>.</value>
+        public bool NotifyClientOfEvents { get; set; } = true;
+
+        /// <summary>
+        /// Gets an instance with default properties.
+        /// </summary>
+        /// <remarks>
+        /// Callers should *not* mutate properties on this instance since it is shared.
```
ikeras: You could introduce an interface that only had the getters for MethodNameTransform and NotifyClientOfEvents that this property returns to guarantee that it isn't mutated.

src/StreamJsonRpc/JsonRpc.cs
AArnott: Good point. Fixed.

src/StreamJsonRpc/JsonRpcTargetOptions.cs
AArnott: I would do that, or introduce a "frozen" field that property setters would throw if true for. But since this is an internal member, and currently only used in one place, solving the problem seemed overkill.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1455,5 +1481,50 @@ internal OutstandingCallData(object taskCompletionSource, Action<JsonRpcMessage>
...
+
+            public void Dispose()
+            {
+                this.eventInfo.RemoveEventHandler(this.server, this.registeredHandler);
+            }
+
+            private void OnEventRaised(object sender, EventArgs args)
+            {
+                // We use null for the sender because we don't want to try to serialize the target object to the remote party.
+                this.jsonRpc.NotifyAsync(this.eventInfo.Name, new object[] { null, args });
```
jasongin: The JSON-RPC protocol allows object (non-array) message parameters. Why not pass the `EventArgs` directly that way, instead of in an array? Then the proxy on the other side would raise the event in response to the notification, and the proxy itself is filled in as the source at that point. (That's how VSLS RPC events work.) Is the concern that the other side might not know how to dispatch events like that?

AArnott: Passing only the `EventArgs` object itself as the one parameter object (without an array), per my reading of the json-rpc spec, implies that the *members* of the `EventArgs` object are the parameters on the method to be invoked on the remote side. But the parameters on the method to be invoked are not the members of the `EventArgs` -- the parameters are `(object, EventArgs)`. This requires an array, since I have no object that contains both of these values, nor can I predict what the parameter names would be so that I could give that object the names to match those parameters.

AArnott: > There is no method being directly invoked on the remote side. It's an event being raised instead. 

That's making an assumption about the remote side, which is something we try to avoid in StreamJsonRpc because we interop with other JSON-RPC implementations. JSON-RPC is all about Remote Procedure Calls, so in fact per the protocol, the only thing we can do is send messages that presumably invoke a method on the remote side. That's what notify does. Now, if the remote side chooses to convey that to its user code by raising an event, that's their implementation detail. It's also beyond the scope of *this* PR, which is just about sending the notification message. 

In a subsequent PR I will be adding proxy generation for server interfaces, which will include translating this notify message into raising an event on the generated client proxy. At that point, as you say, there may or may not be event handlers wired up to the event, and no one cares what their parameter names are. 

I'd be very interested in carrying on this discussion and comparing this to your existing implementation -- perhaps we should meet in person.

>  Event-handler methods would expect the sender parameter to be the RPC proxy object, not something supplied by the other side.

I agree.

AArnott: > But isn't `[null, args]` also making an assumption about the other side

The client can't be allowed to dictate the signature of callbacks the server will send. The server has to set the contract both for its own methods and of the callbacks that it will send back to the client. So no, we're not making an assumption about the client's method signature -- we assert it by spec on the server, and a client that wants to interop needs to fall in line with that.

Now, as to whether it makes sense at a protocol level to define that the client must have a method that takes an object as a first parameter whose value will always be null, that's a good question. And ya, we have an opportunity here to define whether that's the spec or something else is. This is a good time to change it if what I have in this PR isn't the ideal.

The reason I went with it is it matches the signature that is naturally on the server, and will naturally be on the client. So artificially removing the 1st parameter from the message, and then expecting the remote end to artificially add it back felt like unnecessary work. But you are persuading me to rethink that. It's *clearly* an artifact of .NET event handler signatures that is getting 'leaked' into the protocol by going this route. And the remote side may not be .NET. 

Another consideration is the remote side may implement the callback as a standard method (rather than making it look like an EventHandler), in which case, taking a null first argument makes no sense. Taking an `EventArgs` parameter in that method would make sense, I think. Taking the *members* of an `EventArgs` into that method as individual parameters would *not* make sense, IMO. So that's leading me to think I should remove the null first parameter, but keep the array around the `EventArgs` object. 

On the remote side then, when I do proxy generation, I can see an incoming notify message with a method name that coincides with an event defined on the proxy's interface. I can then take the one parameter in the array, deserialize it to match the EventArgs type actually used on the event, then raise the event passing the `JsonRpc` instance in as the first parameter. I *could* make this all work if the json-rpc message carried a parameter object instead of an array, but IMO that would ruin the regular client case for which I have no proxy generation of an event.

Thoughts?

PR #107, opened 2018-05-16
Remove first (null) argument from notify events

AArnott:
This is a follow-up to #105 based on my conversations with @jasongin on that PR. It removes the static "null" argument from the notify messages that propagate a raised event from the local target object to the remote party.


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1523,7 +1523,7 @@ public void Dispose()
             private void OnEventRaised(object sender, EventArgs args)
             {
                 // We use null for the sender because we don't want to try to serialize the target object to the remote party.
-                this.jsonRpc.NotifyAsync(this.eventInfo.Name, new object[] { null, args });
```
jasongin: Update the comment.

AArnott: doh!
Thanks. Fixed.

PR #108, opened 2018-05-17
Add JsonRpc.Attach<T> client proxy generating method

AArnott:
When a JSON-RPC server offers a .NET interface that describes their API, StreamJsonRpc's `JsonRpc` class can now automatically generate a client proxy that implements that interface to give the client a strongly-typed way to send these RPC messages. This in effect replaces such code as:

```cs
var rpc = JsonRpc.Attach(stream);
Result result = await rpc.InvokeAsync<Result>("MyMethodAsync", 1, 2, "three");
```

with: 

```cs
IServer rpc = JsonRpc.Attach<IServer>(stream);
Result result = await server.MyMethodAsync(1,2, "three");
```

These client proxies also implement `IDisposable` if returned from the *static* `Attach<T>` method, since that is the only value that callers have that might be disposed to close the connection. There is also an instance method version of `Attach<T>`  so that more control and potentially multiple client proxies can be attached to the same stream.

These client proxies also support events defined on the `IServer` interface, such that if the server raises callbacks to the client, it will automatically raise any event handlers for that event on the client side. This is to support #105 in the client proxy.

The only supported members on the server interface are:
1. methods that return `Task` or `Task<T>`
1. events that are typed as `EventHandler` or `EventHandler<T>`

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
@@ -41,6 +41,13 @@ public class JsonRpcProxyGenerationTests : TestBase
         Task<int> AddAsync(int a, int b);
 
         Task IncrementAsync();
+
+        Task Dispose();
+    }
+
+    public interface IServerWithNonTaskReturnTypes
+    {
+        int Add(int a, int b);
     }
 
     [Fact]
@@ -84,14 +91,28 @@ public class JsonRpcProxyGenerationTests : TestBase
         Assert.True(this.clientStream.IsDisposed);
     }
 
+    [Fact]
+    public async Task DisposeCollision()
+    {
+        // We're calling IServer.Dispose -- NOT IDisposable.Dispose here.
+        // Verify that it invokes the server method rather than disposing of the client proxy.
+        await this.clientRpc.Dispose();
+        Assert.Equal(-1, this.server.Counter);
+        Assert.False(this.clientStream.IsDisposed);
+    }
+
+    [Fact]
+    public void NonTaskReturningMethod()
+    {
+        var streams = FullDuplexStream.CreateStreams();
+        Assert.Throws<ArgumentException>(() => JsonRpc.Attach<IServerWithNonTaskReturnTypes>(streams.Item1));
+    }
+
     // TODO:
-    // * interfaces that contain non-Task returning methods
-    // * Task returning methods (not Task<T>)
     // * RPC method names that vary from the CLR method names
-    // * no arguments
-    // * 1 argument
+    // * CancellationToken
     // * events
-    // * Dispose method on server interface
+    // * Internal interface
 
     internal class Server : IServer
     {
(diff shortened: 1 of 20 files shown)


src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -0,0 +1,353 @@
...
+                        il.Emit(OpCodes.Ldstr, evt.Name);
+                        il.Emit(OpCodes.Ldarg_0);
+                        il.Emit(OpCodes.Ldftn, raiseEventMethod);
+                        il.Emit(OpCodes.Newobj, delegateCtor);
+                        il.Emit(OpCodes.Callvirt, addLocalRpcMethod);
+                    }));
+                }
+
+                // .ctor(JsonRpc)
+                {
```
jepetty: Is there a reason beyond organization to introduce a new scope for this?

src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -0,0 +1,353 @@
...
+                        action(il);
+                    }
+
+                    il.Emit(OpCodes.Ret);
+                }
+
+                // IDisposable.Dispose()
+                if (disposable)
+                {
+                    var disposeMethod = proxyTypeBuilder.DefineMethod(nameof(IDisposable.Dispose), MethodAttributes.Private | MethodAttributes.HideBySig | MethodAttributes.NewSlot | MethodAttributes.Virtual);
```
jasongin: The `Dispose()` implementation could be written in C# instead of using `il.Emit()`, by putting it on an abstract base class that is extended by the generated type. The base class could also provide helper methods that could simplify some of the other generated code.

src/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
```diff
@@ -0,0 +1,343 @@
...
+        Assert.False(this.clientStream.IsDisposed);
+    }
+
+    [Fact]
+    [Trait("NegativeTest", "")]
+    public void NonTaskReturningMethod()
+    {
+        var streams = FullDuplexStream.CreateStreams();
+        var exception = Assert.Throws<NotSupportedException>(() => JsonRpc.Attach<IServerWithNonTaskReturnTypes>(streams.Item1));
+        this.Logger.WriteLine(exception.Message);
```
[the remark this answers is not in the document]
AArnott: Because the exception messages themselves are usually formatted, and just testing the exception *type* doesn't really test that the message was formatted properly. Writing automated tests to validate the message is tough to get to pass/fail consistently when running on non-ENU machines where translated messages can often make a message test fail.
So, I took the easy way out and the test writes the message out so I can just eyeball it to make sure it's doing it right.

src/StreamJsonRpc/ProxyGeneration.cs
AArnott: Besides readability, it allows me to redefine locals for each generated method. Sharing locals between generated methods tends to make for very brittle code.

src/StreamJsonRpc/ProxyGeneration.cs
AArnott: True. I noticed that the Cascade implementation derived from a base class. If I had to do anything particularly complicated that I could move to a base class I would consider that. So far, anything that could move to a base class was simple enough that I didn't mind emitting IL for it. That gives me some flexibility to potentially in the future support folks bringing their own base classes to the table if they want, and we'll derive from that. We wouldn't be able to do that if we had to derive from our own base class.

jasongin: OK

PR #109, opened 2018-05-17
Do not scan accessor methods on target objects

AArnott:
Now that we allow JsonRpc users to transform every method name that is an RPC target, I noticed that accessor methods to events and properties (e.g. `add_SomeEvent` and `get_SomeProperty`) were also being selected as potential RPC target methods. This of course is not what the server target object is expecting, so this change filters those out.


AArnott:
CC: @dgriffen @jepetty

PR #117, opened 2018-06-04
Fix handling of Task<T>-derived types returned from async methods

### LINKED ISSUE #116: Failure returning result of async method on .NET Core 2.1
A `Task<T>` returning async method compiled with C# 7.3 and running on .NET Core 2.1 actually returns an instance of `System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox<T>`. 

`JsonRpc.HandleInvocationTaskResult` uses reflection to acquire the value of `Task<T>.Result` in such a way that if a `Task<T>`-derived type is returned, it fails to find the `Result` property and returns `null` back to the caller.

https://github.com/Microsoft/vs-streamjsonrpc/blob/b770d6bf7cf05a1a6b9bbb83910bab05e0032b9c/src/StreamJsonRpc/JsonRpc.cs#L1194-L1195

AArnott:
.NET Core 2.1 changed how async `Task<T>` returning methods run, such that they now return a derived-type of `Task<T>`. StreamJsonRpc used reflection to read `Task<T>.Result` and it wasn't prepared for a `Task<T>`-derived type. As a result, StreamJsonRpc is broken on .NET Core 2.1, in that any async method invoked by StreamJsonRpc will return `null` to the client as the result of the method instead of the actual value.

With this fix, we properly walk up the type hierarchy till we find the `Task<T>` so we can reliably find the `Result` property and call it. 
In the future, if this were to fail, we now fail hard (throwing an `InternalErrorException` instead of quietly returning null) to make this more discoverable.

In the process, I added .NET Core 2.1 as a target so tests always run on it, but it led to [a failure on Travis CI](https://github.com/travis-ci/travis-ci/issues/9703) so I've held out that change for now. 
I did enhance an existing test to catch the problem regardless of the target framework though.

Fixes #116

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -848,6 +848,33 @@ namespace StreamJsonRpc
             return JsonRpcMessage.CreateError(id, JsonRpcErrorCode.InvocationError, exception.Message, JObject.FromObject(data, DefaultJsonSerializer));
         }
 
+        /// <summary>
+        /// Extracts the literal <see cref="Task{T}"/> type from the type hierarchy of a given type.
+        /// </summary>
+        /// <param name="taskTypeInfo">The original type of the value returned from an RPC-invoked method.</param>
+        /// <param name="taskOfTTypeInfo">Receives the <see cref="Task{T}"/> type that is a base type of <paramref name="taskTypeInfo"/>, if found.</param>
+        /// <returns><c>true</c> if <see cref="Task{T}"/> could be found in the type hierarchy; otherwise <c>false</c>.</returns>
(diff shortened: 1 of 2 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -848,6 +848,33 @@ private static JsonRpcMessage CreateError(JToken id, Exception exception)
             return JsonRpcMessage.CreateError(id, JsonRpcErrorCode.InvocationError, exception.Message, JObject.FromObject(data, DefaultJsonSerializer));
         }
 
+        /// <summary>
+        /// Extracts the literal <see cref="Task{T}"/> type from the type hierarchy of a given type.
+        /// </summary>
+        /// <param name="taskTypeInfo">The original type of an value returned from an RPC-invoked method.</param>
```
ikeras: 'a' value

PR #118, opened 2018-06-06
Add .NET Core 2.1 testing

AArnott:
Also take care to actually test our .netcoreapp1.0 library on .NET Core 1.x, instead of quietly running on 2.x because that's the only version of the runtime that is installed.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,6 +10,7 @@ skip_commits:
   - doc/*
   - README.md
   - .vsts-ci.yml
+  - .travis.yml
 nuget:
   disable_publish_on_pr: true
 image: Visual Studio 2017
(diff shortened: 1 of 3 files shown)


AArnott:
We've been blocked in doing this by https://github.com/travis-ci/travis-ci/issues/9703, but I'm trying to workaround the issues.

AArnott:
Alright. All issues resolved.

.appveyor.yml
```diff
@@ -19,7 +20,8 @@ environment:
   TreatWarningsAsErrors: true
   codecov_token: 9a7c2ba3-0a4b-4479-96e8-3bfd01a982f6
 before_build:
-  msbuild src\StreamJsonRpc.sln /nologo /m /v:quiet /t:restore
+- dotnet --info
```
ikeras: I assume this is here for logging so we can see the environment?

AArnott: Yes. It gives us a record of what runtimes are installed on appveyor for a given build. Right now they have quite a few installed -- everything we need.

PR #136, opened 2018-07-25
Ensures name transforms are used for events

### LINKED ISSUE #135: Name transforms are not used for events on target objects or proxies
When creating a proxy or passing a local target with name transforms, the transforms aren't applied to event names causing compatibility issues with other JsonRPC clients.

BertanAygun:
Adds usage of name transform functions to both proxy generation and also while raising notify messages. Adds unit tests for relevant cases. Fixes #135.

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
@@ -331,6 +331,42 @@ public class JsonRpcProxyGenerationTests : TestBase
         Assert.False(tcs.Task.IsCompleted);
     }
 
+    [Fact]
+    public async Task NamingTransformsAreAppliedToEvents()
+    {
+        var streams = FullDuplexStream.CreateStreams();
+        this.serverStream = streams.Item1;
+        this.clientStream = streams.Item2;
+
+        var camelCaseOptions = new JsonRpcProxyOptions { MethodNameTransform = CommonMethodNameTransforms.CamelCase };
+        var prefixOptions = new JsonRpcProxyOptions { MethodNameTransform = CommonMethodNameTransforms.Prepend("ns.") };
+
+        // Construct two client proxies with conflicting method transforms to prove that each instance returned retains its unique options.
+        var clientRpc = new JsonRpc(this.clientStream, this.clientStream);
+        var clientRpcWithCamelCase = clientRpc.Attach<IServer>(camelCaseOptions);
+        var clientRpcWithPrefix = clientRpc.Attach<IServer>(prefixOptions);
+        clientRpc.StartListening();
+
+        // Construct the server to only respond to one set of method names for now to confirm that the client is sending the right one.
+        this.serverRpc = new JsonRpc(this.serverStream, this.serverStream);
+        this.serverRpc.AddLocalRpcTarget(this.server, new JsonRpcTargetOptions { MethodNameTransform = camelCaseOptions.MethodNameTransform });
+        this.serverRpc.StartListening();
+
+        var tcs = new TaskCompletionSource<EventArgs>();
+        EventHandler handler = (sender, args) => tcs.SetResult(args);
+        clientRpcWithCamelCase.ItHappened += handler;
+        this.server.OnItHappened(EventArgs.Empty);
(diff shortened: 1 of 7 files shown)


src/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
```diff
@@ -331,6 +331,42 @@ public void InternalInterface()
...
+
+        var tcs = new TaskCompletionSource<EventArgs>();
+        EventHandler handler = (sender, args) => tcs.SetResult(args);
+        clientRpcWithCamelCase.ItHappened += handler;
+        this.server.OnItHappened(EventArgs.Empty);
+        var actualArgs = await tcs.Task.WithCancellation(this.TimeoutToken);
+        Assert.NotNull(actualArgs);
+
+        clientRpcWithCamelCase.ItHappened -= handler;
+        clientRpcWithPrefix.ItHappened += handler;
```
ikeras: I know it won't matter here, but I always prefer unsubscribing handlers holistically. That way if the test gets extended in the future someone doesn't miss this (so suggesting that after your final assert you add clientRpcWithPrefix.ItHappened -= handler;

BertanAygun: Resolved.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1586,16 +1586,22 @@ private class EventReceiver : IDisposable
...
                 Requires.NotNull(server, nameof(server));
                 Requires.NotNull(eventInfo, nameof(eventInfo));
 
+                options = options ?? JsonRpcTargetOptions.Default;
+
                 this.jsonRpc = jsonRpc;
                 this.server = server;
                 this.eventInfo = eventInfo;
+
+                this.rpcEventName = options.MethodNameTransform != null ? options.MethodNameTransform(eventInfo.Name) : eventInfo.Name;
```
AArnott: Are we confident that having a single transform for method names and event names is appropriate? The alternative I'm thinking about is offering a second transform function property so folks can have one policy for events and one for methods. Since they are implemented by different parties, and the tense of the method name may be different, I wonder if it might be useful at some point.

src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -79,6 +79,9 @@ internal static TypeInfo Get(TypeInfo serviceInterface, bool disposable)
 
                 VerifySupported(!serviceInterface.DeclaredProperties.Any(), Resources.UnsupportedPropertiesOnClientProxyInterface, serviceInterface);
 
+                var methodNameTransformPropertyGetter = typeof(JsonRpcProxyOptions).GetRuntimeProperty(nameof(JsonRpcProxyOptions.MethodNameTransform)).GetMethod;
+                var funcOfStringStringInvoke = typeof(Func<string, string>).GetRuntimeMethod(nameof(JsonRpcProxyOptions.MethodNameTransform.Invoke), new Type[] { typeof(string) });
```
AArnott: Should we promote these to `static readonly` fields as we have done already for other members?

src/StreamJsonRpc.Tests/TargetObjectEventsTests.cs
```diff
@@ -120,6 +120,26 @@ public void EventsAreNotOfferedAsTargetMethods()
...
+                eventNameTransformSeen = true;
+            }
+
+            return name;
+        };
+
+        var serverRpc = new JsonRpc(this.serverStream, this.serverStream);
+        serverRpc.AddLocalRpcTarget(this.server, new JsonRpcTargetOptions { MethodNameTransform = methodNameTransform });
+        this.server.TriggerEvent(EventArgs.Empty);
+        Assert.True(eventNameTransformSeen);
```
AArnott: This doesn't actually assert that the notify message used the result of the method transform. And if it did, your transform doesn't actually change the string. Can you revise this test so that it actually verifies that?

src/StreamJsonRpc/JsonRpc.cs
BertanAygun: Resolved

PR #137, opened 2018-07-30
Reduce GC pressure

AArnott:
This is a bunch of micro-optimizations to remove allocations based on a study of PerfView memory allocation traces from work on https://github.com/AArnott/Nerdbank.Streams/issues/15.

This drops per-iteration allocations by about 50% (from 323 to 152 in my testing).

### DIFF
--- a/src/StreamJsonRpc/Reflection/TargetMethod.cs
+++ b/src/StreamJsonRpc/Reflection/TargetMethod.cs
@@ -24,7 +24,7 @@ namespace StreamJsonRpc
         internal TargetMethod(
             JsonRpcMessage request,
             JsonSerializer jsonSerializer,
-            IEnumerable<MethodSignatureAndTarget> candidateMethodTargets)
+            List<MethodSignatureAndTarget> candidateMethodTargets)
         {
             Requires.NotNull(request, nameof(request));
             Requires.NotNull(jsonSerializer, nameof(jsonSerializer));
--- a/src/StreamJsonRpc/Reflection/TargetMethod.cs
+++ b/src/StreamJsonRpc/Reflection/TargetMethod.cs
@@ -112,21 +112,22 @@ namespace StreamJsonRpc
                     return null;
                 }
 
-                var args = new List<object>(2);
-                args.Add(request.Parameters);
-
-                if (method.Parameters.Length > 1 && method.Parameters[1].ParameterType == typeof(CancellationToken))
-                {
-                    args.Add(CancellationToken.None);
-                }
-
                 if (method.Parameters.Length > 2)
                 {
                     // We don't support methods with more than two parameters.
                     return null;
                 }
 
-                return args.ToArray();
+                bool includeCancellationToken = method.Parameters.Length > 1 && method.Parameters[1].ParameterType == typeof(CancellationToken);
+
+                var args = new object[includeCancellationToken ? 2 : 1];
+                args[0] = request.Parameters;
+                if (includeCancellationToken)
+                {
+                    args[1] = CancellationToken.None;
+                }
+
+                return args;
             }
 
             // The number of parameters must fall within required and total parameters.
--- a/src/StreamJsonRpc/DataContracts/JsonRpcMessage.cs
+++ b/src/StreamJsonRpc/DataContracts/JsonRpcMessage.cs
@@ -143,18 +143,11 @@ namespace StreamJsonRpc
             Requires.NotNull(jsonSerializer, nameof(jsonSerializer));
 
             int index = 0;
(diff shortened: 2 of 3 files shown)


src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
```diff
@@ -203,50 +206,58 @@ private enum HeaderParseState
         /// <inheritdoc />
         protected override async Task WriteCoreAsync(string content, Encoding contentEncoding, CancellationToken cancellationToken)
         {
-            var sendingBufferStream = new MemoryStream(MaxHeaderElementSize);
+            this.sendingBufferStream.SetLength(0);
```
BertanAygun: This does seem to release the memory allocated for the earlier buffer according to some online docs. If that's the case and if the reset doesn't honor the initial size that was passed in the constructor earlier, I can see how this can end up causing more memory allocation if it needs to grow its buffer during write operations. Would it be worth just doing a seek to 0 since we already explicitly copy with size.

Maybe the docs are not current though.

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
```diff
@@ -203,50 +206,58 @@ private enum HeaderParseState
...
 #pragma warning restore VSTHRD103 // Call async methods when in an async method
 
-            // Either write both the header and the content, or don't write anything.
-            // If we write only the header when the cancellation comes, that would confuse the recieving side
-            // and corrupt the data data it reads.
-            cancellationToken.ThrowIfCancellationRequested();
+                // Either write both the header and the content, or don't write anything.
+                // If we write only the header when the cancellation comes, that would confuse the recieving side
+                // and corrupt the data data it reads.
+                cancellationToken.ThrowIfCancellationRequested();
```
BertanAygun: How likely it is that we would get a cancellation here between starting to send a message and preparing the headers? Would the savings by removing the intermediate buffere be worth not supporting cancellation here?

src/StreamJsonRpc/Reflection/TargetMethod.cs
```diff
@@ -15,36 +15,38 @@ namespace StreamJsonRpc
...
-
-            KeyValuePair<MethodSignatureAndTarget, object[]> methodWithParameters = targetMethods.FirstOrDefault();
-            if (methodWithParameters.Key.Signature != null)
-            {
-                this.target = methodWithParameters.Key.Target;
-                this.method = methodWithParameters.Key.Signature.MethodInfo;
-                this.parameters = methodWithParameters.Value;
-                this.AcceptsCancellationToken = methodWithParameters.Key.Signature.HasCancellationTokenParameter;
+                object[] args = this.TryGetParameters(request, method.Signature, jsonSerializer, request.Method);
+                if (this.method == null && args != null)
```
BertanAygun: Can you rename the local method variable to candidateMethod so that it is not confused by "this.method" :)

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
AArnott: The odds of cancellation are slim. But I think even if we removed this line, we would keep the temporary buffer. The reason being that writing many tiny bytes to a Stream of an unknown transport can actually be inefficient. It can be better to be chunky when writing to a Stream. When we switch to Pipes this won't be an issue, but that's a 2.0 feature.

src/StreamJsonRpc/HeaderDelimitedMessageHandler.cs
AArnott: The docs can be a little confusing. But they don't claim that capacity is ever reduced -- only potentially expanded. In fact in the code I've confirmed that capacity is *not* reduced. So this is an effective way to reuse the existing backing array.

src/StreamJsonRpc/Reflection/TargetMethod.cs
AArnott: Done

AArnott:
> *Do not merge yet*: Very ironically, after making any one of the first few changes included here, GC pressure goes significantly *higher* (by 10X in my testing using the test behind the above issue). I have no idea why, as it makes no sense whatsoever. So we need to investigate this more deeply.
> That said, IMO this PR has a bunch of good changes and PerfView and ildasm confirms that each micro-optimization results in that allocation disappearing.

I have removed the commits that cause the 10X regression. There were 5 of them. I don't know why on earth they would regress the scenario I'm testing, but I'll figure it out and reserve those for a subsequent PR once I figure them out.

Issue #111, opened 2018-05-18
Allow for custom message routing/handling


AArnott:
## Requirements
1. A specially initialized `JsonRpc` object may send messages that are routed in a non-standard way. For example, sent messages may be routed to different recipients depending on the method name.
1. A specialized router that receives and forwards JSON-RPC messages to various destinations, possibly changing their contents along the way. One destination may be a "local" `JsonRpc` instance.

### Rough notes from earlier:
1. ✔️  Revise `DelimitedMessageHandler` contract to convert between `JObject` and the transport (consider #80) instead of a single string containing the entire message.
1. ✔️ Allow for a plugin serializer to convert between `JToken` and `byte[]` (on the stream). The idea being to support a binary serialization later on as an option. ❓ But web sockets probably would prefer to remain encoded as text rather than as binary, so how would that work? Perhaps we don't know whether it's a string or `byte[]` on the transport side. We just give and receive `JToken` objects with this plugin and it serializes it to the transport however it wants? But that couples the serializer to the transport. 👎 
1. ❌ Allow 'middleware' to intercept `JObject` being written or read and choose to transform it, or redirect it.
1. ✔️ Alternative to intercepting/passing around JObject would be to let our own strongly typed JSON-RPC classes (`JsonRpcMessage`, `JsonRpcError`, etc.) be the medium of exchange.

DustinCampbell:
I'm interested in this too. Currently, it's pretty difficult to extract the `id` of a `JsonRpcMessage`, which is needed for tracking requests and handling the [`$/cancelRequest` notification](https://microsoft.github.io/language-server-protocol/specification#cancelRequest) from LSP.

AArnott:
@DustinCampbell StreamJsonRpc has built-in support for cancellation the way you mention and automatically associates responses with their requests. Why do you need the `id` of the `JsonRpcMessage`?

Or is it that you're code is an intermediary (and thus you're commenting on a routing issue)?

DustinCampbell:
I know that there's support for dispatching to methods with a `CancellationToken`. However, it was unclear to me how that would work if the client sends a notification to the server to request cancellation of a previous request. That's a scenario supported by LSP.

AArnott:
Yes, when StreamJsonRpc dispatches to a method that accepts `CancellationToken`, it wires it up such that when an LSP-style cancellation notification comes in, StreamJsonRpc will automatically cancel the token.
[...]

PR #141, opened 2018-08-12
Avoid allocating string to represent whole JSON messages

### LINKED ISSUE #80: Refactor DelimitedMessageHandler so it is Stream-agnostic
We have at least one user that wants to use this library over a WebSocket, which is not exposed as a Stream. StreamJsonRpc can work with this by deriving from `DelimitedMessageHandler`, but it requires faking out the `DelimitedMessageHandler` with `Stream.Null` instances in order to indirectly control the return values of the `CanRead` and `CanWrite` properties.

When we have a chance for some minor binary breaking changes, we may want to refactor this class so that implementations that aren't dependent on Stream will work.

AArnott:
**Important:** This PR contains API binary breaking changes, and source breaking changes that will mostly impact any externally defined `DelimitedMessageHandler`-derived typed (which we don't know of any that exist).

We avoid allocating strings by changing the contract with our message handlers from reading and writing strings to reading and writing `JToken`. In this implementation, we still allocate contiguous memory for the entire message as an encoded byte buffer (which we were doing previously as well), but we no longer allocate a second "copy" of that for a string representation.

In this commit we define `IMessageHandler`. This new `IMessageHandler` interface defines a minimal, and more flexible abstraction than `DelimitedMessageHandler` did.

This change also makes use of `ValueTask` to reduce overhead from Task allocations.

Closes #80 since `DelimitedMessageHandler` is now not the only way to customize message serialization and the new abstraction does not require `Stream`.
Makes progress toward #111 since the new `IMessageHandler` abstraction allows for redirecting/modifying messages in any way.


AArnott:
Good to know. Thanks, @tinaschrepfer

Issue #65, opened 2017-10-12
Allow server to specify data object in error result messages


AArnott:
We should define an Exception-derived type (that includes an object property) that the RPC server method can throw in order to specify the `JToken` (`JValue` or `JObject`) for the `data` field in the JSON-RPC error result message.

Today, we always [create a data object with the stack and code fields in it](https://github.com/Microsoft/vs-streamjsonrpc/blob/ce4d9894e7d5cefb180cf2d74c9c7905d4cbec4c/src/StreamJsonRpc/JsonRpc.cs#L843). 

By request from @tinaschrepfer

tinaschrepfer:
It would be great if the caller can specify custom values for all 3 fields: error.code, error.message, error.data.  I can see the language server protocol specifying custom values for each field for certain scenarios.

AArnott:
The `message` and `code` fields are already within the control of the server method that throws the exception (by setting `Exception.Message` and `Exception.HResult`). However, the `code` field can only be an integer. So we'd want to expand that to be a string as well.

Issue #64, opened 2017-10-02
allow StreamJsonRpc to raise cancellation when caller side close the connection


heejaechang:
I see that, if I raise cancellation and close Rpc connection right away, sometimes, cancellation message doesn't reach server side before Rpc connection is disconnected. making server side not know the operation is actually cancelled.

for us, this is bad since we are waiting for other connection to happen inside of service hub, and hang there since it doesn't know the operation is actually cancelled.

I think StreamJsonRpc should either guarantee of ordering (cancellation and disconnect) or give a way to configure streamJsonRpc to always raise cancellation if connection is disconnected.

heejaechang:
tagging @IlyaBiryukov and @AArnott

IlyaBiryukov:
The workaround is to subscribe to server's JsonRpc.Disconnected event and cancel the server methods when the disconnection happens

heejaechang:
that's true, but that makes one to always create linked cancellation token. one from the given one, and one from the disconnect method.

AArnott:
> I think StreamJsonRpc should either guarantee of ordering (cancellation and disconnect) 

There is no way StreamJsonRpc can "guarantee ordering" since a disconnection can happen at any time.

> or give a way to configure streamJsonRpc to always raise cancellation if connection is disconnected.

This seems reasonable. Our existing `Disconnected` event predated our `CancellationToken` support. And not all server methods may even accept a `CancellationToken`. So if we do offer a way to automatically cancel in-progress operations when the connection drops, it will of course only impact those methods that take `CancellationToken` while other methods will still have to listen for the `Disconnected` event or add `CancellationToken` to their method signature, which would be a non-breaking change for the JSON-RPC clients.

AArnott:
@heejaechang I take it this is the root cause of your servicehub hosting process spinning the CPU endlessly? If so, would you use this new feature of StreamJsonRpc if we fixed it fast enough? Which release would you check the fix into, so we can make the feature addition to the right branch of this project as well?

heejaechang:
I am thinking bring it to 15.5 preview1. right now, I implemented what @IlyaBiryukov suggested.

AArnott:
Is that workaround you refer to hooking the `JsonRpc.Disconnected` event? If so, we'll keep this issue open to consider fixing in a future version, but without worrying about helping with your fix in the short term. That sounds like a good call to me.

heejaechang:
[...]

PR #142, opened 2018-08-13
Strip -rc from v1.4

AArnott:


dgriffen:
Are we deciding to leave #65 and #64 out of the initial 1.4 release then?

AArnott:
@dgriffen: ya, I think so. There's always 1.5 that can include those features if we need them sooner than 2.0 will be ready.

PR #148, opened 2018-09-09
Add option to automatically cancel server methods when the connection…

### LINKED ISSUE #64: allow StreamJsonRpc to raise cancellation when caller side close the connection

AArnott:
Closes #64

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -95,6 +95,11 @@ namespace StreamJsonRpc
         private bool startedListening;
 
         /// <summary>
+        /// Backing field for the <see cref="CancelLocallyInvokedMethodsWhenConnectionIsClosed"/> property.
+        /// </summary>
+        private bool cancelLocallyInvokedMethodsWhenConnectionIsClosed;
+
+        /// <summary>
         /// Backing field for the <see cref="SynchronizationContext"/> property.
         /// </summary>
         private SynchronizationContext synchronizationContext;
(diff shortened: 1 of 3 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -260,6 +265,23 @@ public Task Completion
...
+        /// <summary>
+        /// Gets or sets a value indicating whether to cancel all methods dispatched locally
+        /// that accept a <see cref="CancellationToken"/> when the connection with the remote party is closed.
+        /// </summary>
+        public bool CancelLocallyInvokedMethodsWhenConnectionIsClosed
+        {
+            get => this.cancelLocallyInvokedMethodsWhenConnectionIsClosed;
+            set
+            {
+                // We don't typically allow changing this setting after listening has started because
```
ikeras: What cases do we allow it? Seems like we always throw if we've started listening

AArnott: The method we're calling here does not throw if `AllowModificationWhileListening` has been set to true.

https://github.com/Microsoft/vs-streamjsonrpc/blob/cffa2677a9f5fce140e086c43f2446efdd61f7ff/src/StreamJsonRpc/JsonRpc.cs#L1494-L1497

PR #151, opened 2018-09-11
Remove unused JsonRpc.JsonSerializerFormatting property

AArnott:
It was used in the 1.x era, but no more.


ikeras:
The idea here is that because we're updating the version, we're okay with public API breaking changes? Does this count as a major version increase, and how does that update?

Issue #124, opened 2018-06-26
Add support for compact binary serialization


AArnott:
BSON is binary serialization version of JSON. We can use the BsonWriter from [this package](https://www.nuget.org/packages/newtonsoft.json.bson) to possibly get performance wins.

Also consider MessagePack, if indeed it can be a drop-in replacement (via the newtonsoft.json JsonWriter/JsonReader plugin mechanism) and is truly better as [claimed](https://stackoverflow.com/questions/6355497/performant-entity-serialization-bson-vs-messagepack-vs-json/6357042#6357042) by its author. [More info about MessagePack](https://github.com/neuecc/MessagePack-CSharp).

Also consider [gRPC](https://grpc.io/).

Let's measure it and add it if it's significant.

peterwald:
Consider [Bond](https://microsoft.github.io/bond/) over [GRPC](https://microsoft.github.io/bond/manual/bond_over_grpc.html).

AArnott:
I tried out [MessagePack for C#](https://github.com/neuecc/MessagePack-CSharp). It took ~1 second to transmit 200K records, and 63MB of transmitted data.

Using newtonsoft.json to serialize took 14-23 seconds for the same data, and undoubtedly _much_ more bytes of transmitted data, although I did not measure it.

AArnott:
The above measurement was when using integer-keyed properties. When using it in "newtonsoft compability mode", it took on average 1.4 seconds and 88 MB of transmission bytes.

AArnott:
I found some reports that BSON serialization is marginally faster (and sometimes slower) than JSON.

PR #160, opened 2018-09-17
Allow for JSON-alternative formatters for JSON-RPC messages

### LINKED ISSUE #48: Add full support of jsonrpc v 2.0 named parameters
Right now we only support parameter objects by locating a server method that accepts one argument of JToken type in [TargetMethod.cs](https://github.com/Microsoft/vs-streamjsonrpc/blob/master/src/StreamJsonRpc/Reflection/TargetMethod.cs)

We should support locating server methods by matching the parameter names to the argument names, e.g. like [here](https://devdiv.visualstudio.com/DevDiv/_git/CascadePrototype/pullrequest/80621?_a=files&path=%2FJsonRpc%2FArguments.cs)

AArnott:
This is a significant refactoring to remove `JToken` as the standard for serialization in favor of public JSON-RPC message classes, which may be serialized by any means via an extensibility point.

As mentioned in the message for commit 03b92a4, this abstracts *all* JSON serialization away from the `JsonRpc` class, including these bonuses:

> * This opens the way for JSON-alternative serialization formats such as compact binary (#124).
> * This also (finally) exposes the JSON-RPC message types publicly, which will likely help with message routing (#111).
> * Adds full support for named parameters (fixes #48).
> * Use `long` for message IDs on the client instead of `int`
> 

A key point of this change is in d4188ba, which separates message *handlers* from *formatters*. As described in that commit:

> The handlers are concerned with reading and writing messages such that they have separation between each other. The formatters are concerned with actual serialization and deserialization of an individual message.
>
> Formatters and handlers can be (largely) mixed and matched. The one exception to this right now is that the HeaderDelimitedMessageHandler only accepts text-based formatters because it needs to be able to get and set the encoding used as part of the headers.

After this change, `MessagePack` or other forms of serialization become *very* easy additions. I have a `MessagePack` change that builds on this that is literally 2 lines of code (excluding the class and 2 method declarations).


src/StreamJsonRpc/IJsonRpcMessageTextFormatter.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// </summary>
+        /// <value>Never null.</value>
+        /// <exception cref="ArgumentNullException">Thrown at an attempt to set the value to null.</exception>
+        Encoding Encoding { get; set; }
+
+        /// <summary>
+        /// Deserializes a sequence of bytes to a <see cref="JsonRpcMessage"/>.
+        /// </summary>
+        /// <param name="contentBuffer">The bytes to deserialize.</param>
+        /// <param name="encoding">The encoding to read the bytes in <paramref name="contentBuffer"/> with.</param>
```
ikeras: might want to mention that this cannot be null

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// <summary>
+    /// Uses Newtonsoft.Json serialization to serialize <see cref="JsonRpcMessage"/> as JSON (text).
+    /// </summary>
+    public class JsonMessageFormatter : IJsonRpcMessageTextFormatter
+    {
+        /// <summary>
+        /// The key into an <see cref="Exception.Data"/> dictionary whose value may be a <see cref="JToken"/> that failed deserialization.
+        /// </summary>
+        internal const string ExceptionDataKey = "JToken";
```
ikeras: not that I think it matters much, but I suppose you could use nameof(JToken) here

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -0,0 +1,357 @@
...
+            return
+                idToken?.Type == JTokenType.Integer ? idToken.Value<long>() :
+                idToken?.Type == JTokenType.String ? idToken.Value<string>() :
+                idToken == null ? (object)null :
+                throw new JsonSerializationException("Unexpected type for id property: " + idToken.Type);
+        }
+
+        private void WriteJToken(IBufferWriter<byte> contentBuffer, JToken json)
+        {
+            var jsonWriter = new JsonTextWriter(new StreamWriter(contentBuffer.AsStream(), this.Encoding, 4096));
```
ikeras: Both this and JsonTextReader are disposable

src/StreamJsonRpc/Reflection/TargetMethod.cs
```diff
@@ -100,80 +102,47 @@ private void AddErrorMessage(string message)
...
+                        CultureInfo.CurrentCulture,
+                        Resources.MethodParameterCountDoesNotMatch,
+                        method,
+                        methodParameterCount,
+                        request.ArgumentCount));
+                    return false;
+                case JsonRpcRequest.ArgumentMatchResult.ParameterArgumentTypeMismatch:
+                    this.AddErrorMessage(string.Format(CultureInfo.CurrentCulture, Resources.MethodParametersNotCompatible, method));
+                    return false;
+                case JsonRpcRequest.ArgumentMatchResult.MissingArgument:
```
ikeras: FWIW, I would put return false here even though you don't need to, as otherwise it's too easy to add another case below here that does something else

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: I agree it doesn't really matter since the term will never be renamed. But I don't think it's appropriate here because the string isn't actually locked to that identifier. It's part of a public *protocol* contract for looking up the error. If the identifier were named something differently later, we would still keep this string to avoid breaking the wire protocol. It's just that the wire protocol was inspired by the name of a thing we're using.

---
In reply to: [218242767](https://github.com/Microsoft/vs-streamjsonrpc/pull/160#discussion_r218242767) [](ancestors = 218242767)

PR #162, opened 2018-09-18
Add MessagePack sample and tests

### LINKED ISSUE #124: Add support for compact binary serialization

AArnott:
To test this adequately I've split the JsonRpcTests class into two derived types to "replay" most tests against both MessagePack + Length handler as well as the original Json + Header handler stack.

At least for now, I'm leaving `MessagePackFormatter` in the test project rather than the library project. There are several optional paths folks can go down when using `MessagePack`, and any of them take about 1 line of code anyway. So rather than restrict folks to just one configuration, or writing code to support them all, the plan is to document the test project's version as a sample for folks to copy and/or inspire their own version.

Closes #124

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -6,6 +6,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Serialization;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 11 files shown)


src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1635,11 +1505,13 @@ public int PlusTwo(int arg)
         }
     }
 
+    [DataContract]
     public class Foo
     {
-        [JsonProperty(Required = Required.Always)]
+        [DataMember(Order = 0, IsRequired = true)]
```
jepetty: Is ordering required for MessagePack?

AArnott: Yes, in its optimal mode. It *can* be configured to use named properties but it makes the binary format much bigger, so it isn't the recommended approach.

heejaechang:
@AArnott so, this gives a capability for JsonRpc to use a messagepackformatter for communication but actual custom type still uses json converter?

AArnott:
No. `JsonConverter` only applies if you're using the default `JsonMessageFormatter` (the newtonsoft.json adatper).

If you're using `MessagePack`, you need to either make sure that all your exchange types are `MessagePack` compliant (this can be done by adding `[DataContract]` and `[DataMember]` attributes to your contracts) _or_ using the `MessagePack.Resolvers.ContractlessStandardResolver.Instance` resolver, which makes it act more like newtonsoft.json where it just serializes everything. The latter option does _not_ (AFAIK) make it act _just_ like newtonsoft.json, but it allows it to work without special dependencies or annotations in your exchange types. The most compact and performant serialization comes from using attributes that specify the order that members should be serialized in (or more precisely: the position in an array the value for a member should be dropped into).

PR #169, opened 2018-09-20
Add JSON-RPC message and event tracing

### LINKED ISSUE #115: Log interface needed in vs-streamjsonrpc
Ideally I think JsonRpc should provide some way to let caller to add logging/tracing (error, info).

In my example, when JsonRpc Server has some error executing some RPC call, it will only transmit the exception back to the client while Server does not know what error happens at all. It would be great to provide a way to let Server to log these errors or info.

Today, it is more popular to add telemetry and tracing to identify issues and bugs. It would be great to have such functionalities.

AArnott:
Closes #115

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -68,7 +68,7 @@ namespace StreamJsonRpc
         private readonly Action<object> cancelPendingOutboundRequestAction;
 
         /// <summary>
-        /// A delegate for the <see cref="HandleInvocationTaskResult(object, Task)"/> method.
+        /// A delegate for the <see cref="HandleInvocationTaskResult(JsonRpcRequest, Task)"/> method.
         /// </summary>
         private readonly Func<Task, object, JsonRpcMessage> handleInvocationTaskResultDelegate;
 
@@ -159,7 +159,7 @@ namespace StreamJsonRpc
             Requires.NotNull(messageHandler, nameof(messageHandler));
 
             this.cancelPendingOutboundRequestAction = this.CancelPendingOutboundRequest;
-            this.handleInvocationTaskResultDelegate = (t, id) => this.HandleInvocationTaskResult(id, t);
+            this.handleInvocationTaskResultDelegate = (t, request) => this.HandleInvocationTaskResult((JsonRpcRequest)request, t);
 
             this.MessageHandler = messageHandler;
         }
@@ -264,6 +264,16 @@ namespace StreamJsonRpc
             /// Occurs when a locally invoked method from a <see cref="JsonRpcRequest"/> throws an exception (or returns a faulted <see cref="Task"/>).
             /// </summary>
             LocalInvocationError,
+
+            /// <summary>
+            /// Occurs when a successful result message for a prior invocation is received.
+            /// </summary>
+            ReceivedResult,
+
+            /// <summary>
+            /// Occurs when an error message for a prior invocation is received.
+            /// </summary>
+            ReceivedError,
 
             /// <summary>
             /// Occurs when the connection is closed.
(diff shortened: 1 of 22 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1553,10 +1716,63 @@ private void CancelPendingOutboundRequest(object state)
...
+        private void TraceMessageSent(JsonRpcMessage message)
+        {
+            if (this.TraceSource.Switch.ShouldTrace(TraceEventType.Information))
+            {
+                this.TraceSource.TraceData(TraceEventType.Information, (int)TraceEvents.MessageSent, message);
+            }
+
+            if (this.TraceSource.Switch.ShouldTrace(TraceEventType.Verbose))
+            {
+                this.TraceSource.TraceData(TraceEventType.Verbose, (int)TraceEvents.MessageReceived, "Sent: {0}", this.GetMessageJson(message));
```
jasongin: `TraceEvents.MessageSent`

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1553,10 +1716,63 @@ private void CancelPendingOutboundRequest(object state)
             }).Forget();
         }
 
+        private void TraceLocalMethodAdded(string rpcMethodName, MethodSignatureAndTarget targetMethod)
+        {
+            Requires.NotNullOrEmpty(rpcMethodName, nameof(rpcMethodName));
+
+            if (this.TraceSource.Switch.ShouldTrace(TraceEventType.Information))
```
jasongin: Are these `ShouldTrace()` checks necessary? Or an optimization? I thought `TraceEvent()` would already check the switch condition.

AArnott: So did I. But when I added tracing to the `MultiplexingStream` inside Nerdbank.Streams, I found that tracing is _very_ slow even with it turned off unless I added explicit checks like this.

AArnott: slow and/or GC pressure heavy.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1553,10 +1716,63 @@ private void CancelPendingOutboundRequest(object state)
...
+            try
+            {
+                return JToken.FromObject(message);
+            }
+            catch (Exception ex)
+            {
+                // In the event that a formatter is in use that is not JsonMessageFormatter,
+                // there is no guarantee that the message is serializable with Newtonsoft.Json.
+                // This was a best effort, and it didn't work out.
+                return $"<JSON representation not available: {ex.Message}>";
```
jasongin: Should there be a method on `IJsonRpcMessageFormatter` to get a string representation of a message for tracing?

AArnott: I was thinking about that too... Maybe. But then I'd also have to add a `IJsonRpcMessageHandler.Formatter` property so that I could retrieve the formatter. 
Given you had the same idea, maybe I'll go ahead and do that.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1125,6 +1216,33 @@ private static bool TryGetTaskOfTType(TypeInfo taskTypeInfo, out TypeInfo taskOf
...
+        private JsonRpcError CreateError(object id, Exception exception)
+        {
+            if (exception == null)
+            {
+                throw new ArgumentNullException(nameof(exception));
+            }
+
+            if (this.TraceSource.Switch.ShouldTrace(TraceEventType.Error))
+            {
+                this.TraceSource.TraceEvent(TraceEventType.Error, (int)TraceEvents.LocalInvocationError, "Exception thrown from request \"{0}\": {1}", id, exception);
```
jasongin: I think this error message might be more informative if it included the method name. Or, maybe that would be obvious from the stack trace?

AArnott: I agree. In an earlier iteration I had a method name in it, but since there are multiple places that errors are created, emitting the event from here ensured they were all captured -- but at the loss of the method name. 
The method name will typically show up in the stack trace, but if the original method itself is not async but returns a Task from an inner async method, it won't show up. So I'll circle back and look for a way to add it in all cases.

AArnott: Done.

src/StreamJsonRpc/JsonRpc.cs
AArnott: Done.

PR #170, opened 2018-09-25
Enable JSON-RPC servers to completely control their error response

### LINKED ISSUE #65: Allow server to specify data object in error result messages

AArnott:
Servers could already control the error.message property via the Message property on the exception they throw.
With this change, they can also control the error.code and error.data properties, by throwing a LocalRpcException which defines properties for these two other fields.

This also replaces the RemoteStackTrace and RemoteErrorCode properties which assumed a particular schema for the error.data object in favor of a new type that servers and clients can *optionally* use to convey this and more information.
The default behavior continues to include the information on the server we had before, plus the exception *type*.

Fixes #65

### DIFF
--- a/src/StreamJsonRpc/Exceptions/LocalRpcException.cs
+++ b/src/StreamJsonRpc/Exceptions/LocalRpcException.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace StreamJsonRpc
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+    using System.Text;
+    using Microsoft;
+    using StreamJsonRpc.Protocol;
+
+    /// <summary>
+    /// An exception that may be thrown within a locally invoked server method, and carries with it data that influences the JSON-RPC error message's error object.
+    /// </summary>
+    public class LocalRpcException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LocalRpcException"/> class.
+        /// </summary>
+        public LocalRpcException()
+        {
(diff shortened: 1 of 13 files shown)


src/StreamJsonRpc/Exceptions/LocalRpcException.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// Gets or sets the value for the error.data property.
+        /// </summary>
+        public object ErrorData { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value for the error.code property.
+        /// </summary>
+        /// <value>Any value, excluding the [-32768, -32000] range, which is reserved.</value>
```
ikeras: Is the reason you don't validate that these aren't set because other code uses this and does set that value range appropriately? Perhaps having this version validate and then override error code setter in a derived class to allow only the reserved range to be set.

AArnott: I should update this doc, since it suggests that the value would never be in that reserved range. It certainly can be, but apps shouldn't be *setting* within that range.
As for enforcing by code that no one ever use that range, I'm leaning toward thinking that's overkill. Particularly since whatever might set that odd error code is an exception path for the app and probably poorly tested, I wouldn't want to hide whatever the original error was with a new error.

PR #172, opened 2018-09-28
Fault instead of cancel Tasks tracking outbound requests

### LINKED ISSUE #84: Consider throwing something other than TaskCanceledException on outstanding requests when remote end closes stream
Currently, if the server closes the connection, any outstanding requests on the client complete with tasks faulted with `TaskCanceledException`. This makes sense if the client had disposed its own stream, but not if that is in response to the server closing its end first.

AArnott:
Specifically, fault the tasks with a ConnectionLostException for this scenario.

This is technically a breaking change, but it's one that Roslyn has requested, and we're making lots of breaking changes in 2.0, so it's a reasonable time to do it. The new exception derives from an existing base class for an exception so it should be easy for folks to catch properly if they're looking for the base class.

Fixes #84


AArnott:
When #171 and this PR meet, we need to tack on an update (605bf76b0b) to the documentation to reflect the changed behavior.

Issue #183, opened 2018-10-13
Proxy for methods that don't return tasks


seertenedos:
Currently the proxy only supports interfaces where the methods return tasks. My issue is that in my case i need am trying to add your library between 2 processes over named pipes and the interfaces can't be changed because there is a large amount of code written to use them. Basically i am splitting an application into a few child processes. Since this is an older application everything including interfaces were not done in the async way so nothing returns tasks. Is there a reason why is has to be a return type of task?

AArnott:
As you may already be aware, splitting a process into multiple child processes means preparing each process for new kinds of reliability and performance problems. When your app was a monolithic process, it all went down at once if it crashed, or there was only one thread to become unresponsive due to kernel scheduling or other concerns. But when you're dealing with IPC between processes, you have to be prepared for any subset of processes to crash or become unresponsive. 

For those reasons, synchronous IPC is generally a bad idea. Now I hear what you're saying about legacy requirements. I work on Visual Studio and believe me, we have our share of legacy APIs to honor. But given the recommendation is strongly on the async side, and that JSON-RPC is fundamentally an async protocol, we want StreamJsonRpc to drop you into the "happy path" of doing things right.

In fact, supporting non-Task returning methods would be potentially very problematic because that would mean StreamJsonRpc is responsible to artificially block the calling thread until the result came back. But that can quickly lead to hangs if, for example, the calling thread is your app's UI thread. It can deadlock if the server needs to call back to your client before returning a response, and your client can't respond because its UI thread is hung.

We have ways to address your requirements, and avoid the deadlocks too. We do it at the level of the general requirement of "how do I block the calling thread while executing an async operation?", which is a problem each application must decide how to solve. Our solution is to use the [JoinableTaskFactory](https://aka.ms/vsthreading) as found in the [Microsoft.VisualStudio.Threading](https://www.nuget.org/packages/Microsoft.VisualStudio.threading) library. That can block the UI thread (or any other thread) while executing async operations. In the case of StreamJsonRpc, that means you might have something like this:

```cs
interface IJsonRpcService
{
    Task SomeOperationAsync();
}

// then on the client, you have:
var client = JsonRpc.Attach<IJsonRpcService>(pipe);

// Call the operation, blocking the thread till the operation is done:
joinableTaskFactory.Run(async delegate
{
    await client.SomeOperationAsync();
});
```

[...]

PR #184, opened 2018-10-13
Add doc around sync-blocking RPC requests

AArnott:
Inspired by the requirement expressed in #183

### DIFF
--- a/doc/dynamicproxy.md
+++ b/doc/dynamicproxy.md
@@ -18,3 +18,24 @@ A proxy can only be dynamically generated for an interface that meets these requ
 1. All methods return `Task` or `Task<T>`
 1. All events are typed with `EventHandler` or `EventHandler<T>`
 1. Methods *may* accept a `CancellationToken` as the last parameter.
+
+## Async methods and generated proxies
+
+An interface used to generate a dynamic client proxy must return `Task` or `Task<T>` from all methods.
+This allows the client proxy to be generated with asynchronous methods as appropriate for JSON-RPC (and IPC in general)
+which is fundamentally asynchronous.
+
+### Server-side concerns
+
+On the server side, these same methods may be simple and naturally synchronous. Returning values from the server wrapped
+in a `Task` may seem unnatural.
+The server need not itself explicitly implement the interface -- it could implement the same method signatures as are
+found on the interface except return `void` (or whatever your `T` is in your `Task<T>` method signature on the interface)
+and it would be just fine. Of course implementing the interface may make it easier to maintain a consistent contract
+between client and server.
+
+### Client-side concerns
+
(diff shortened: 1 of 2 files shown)


doc/dynamicproxy.md
```diff
@@ -18,3 +18,24 @@ A proxy can only be dynamically generated for an interface that meets these requ
...
+
+An interface used to generate a dynamic client proxy must return `Task` or `Task<T>` from all methods.
+This allows the client proxy to be generated with asynchronous methods as appropriate for JSON-RPC (and IPC in general)
+which is fundamentally asynchronous.
+
+### Server-side concerns
+
+On the server side, these same methods may be simple and naturally synchronous. Returning values from the server wrapped
+in a `Task` may seem unnatural.
+The server need not itself literally implement the interface -- it could implement the same method signatures as are
```
ikeras: I would say "explicitly"

doc/sendrequest.md
```diff
@@ -54,6 +54,33 @@ will reflect cancellation only after the request's transmission is avoided or th
...
+When you have an unavoidable requirement to synchronously block while waiting for *any* async code to complete (whether JSON-RPC related or otherwise),
+we recommend use of the [JoinableTaskFactory](https://aka.ms/vsthreading) as found in the [Microsoft.VisualStudio.Threading](https://www.nuget.org/packages/Microsoft.VisualStudio.threading) library. That can block the UI thread (or any other thread) while executing async operations. In the case of StreamJsonRpc, that means the client might have something like this:
+
+```cs
+interface IJsonRpcService
+{
+    Task SomeOperationAsync();
+}
+
+var client = JsonRpc.Attach<IJsonRpcService>(pipe);
```
BertanAygun: Is there any requirement for JsonRPC to be JTF aware, for example in the scenario you provided above? I am guessing no but wanted to check.

When server calls back the client via JsonRPC, would their callback always execute at thread pool so it would still be client's callback handler responsibility to use JTF to switch to main thread if needed?

doc/dynamicproxy.md
AArnott: Done.

doc/sendrequest.md
AArnott: StreamJsonRpc already references the threading library, but it isn't JTF aware. It's just typical async code that a JTF aware client could call into and block. But if there were callbacks that needed to access a UI thread blocked by JTF.Run, they'd need to do all the necessary joining to avoid the deadlock.

AArnott: And yes, StreamJsonRpc uses the threadpool to listen for incoming messages, and invocations based on those messages are made via a `SynchronizationContext` that the owner provides (or we use the default one, which schedules to the thread pool).

PR #186, opened 2018-10-17
Remove virtual modifier from InvokeCoreAsync methods

AArnott:
There's no supported use case for overriding these.


AArnott:
Yes @ikeras, this is for 2.0's initial release.

PR #187, opened 2018-10-17
Offer virtual method for controlling error messages

AArnott:
This is in response to https://github.com/Microsoft/python-language-server/pull/276#issuecomment-430803359 where we discovered that an easier way to customize error messages than a message routing extension would be helpful.

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -264,6 +264,8 @@ namespace StreamJsonRpc
 
             /// <summary>
             /// Occurs when a locally invoked method from a <see cref="JsonRpcRequest"/> throws an exception (or returns a faulted <see cref="Task"/>).
+            /// <see cref="TraceListener.TraceData(TraceEventCache, string, TraceEventType, int, object[])"/> is invoked with the thrown <see cref="Exception"/>, request method name, request ID, and the argument object/array.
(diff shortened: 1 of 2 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -942,6 +947,26 @@ protected virtual void Dispose(bool disposing)
...
+        /// This method may be overridden in a derived class to change the way error details are expressed.
+        /// </remarks>
+        protected virtual JsonRpcError.ErrorDetail CreateErrorDetails(JsonRpcRequest request, Exception exception)
+        {
+            var localRpcEx = exception as LocalRpcException;
+            return new JsonRpcError.ErrorDetail
+            {
+                Code = (JsonRpcErrorCode?)localRpcEx?.ErrorCode ?? JsonRpcErrorCode.InvocationError,
+                Message = exception.Message,
+                Data = localRpcEx != null ? localRpcEx.ErrorData : new CommonErrorData(exception),
```
ikeras: Can ErrorData be null? If so, does it matter if that's what returned for Data?

AArnott: Yes, the error.data property [may be null](https://www.jsonrpc.org/specification#error_object). And yes, the intent here is that if the special exception is thrown that gives control over the data object, that we honor it, even if it is null.

PR #189, opened 2018-10-18
Remove UTF-8 preamble from JSON-RPC messages

AArnott:
This fixes a regression from 1.x

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
@@ -1,8 +1,9 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
 using System.Text;
-using System.Threading.Tasks;
+using Nerdbank.Streams;
 using StreamJsonRpc;
 using StreamJsonRpc.Protocol;
 using Xunit;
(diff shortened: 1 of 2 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -40,7 +45,7 @@ public class JsonMessageFormatter : IJsonRpcMessageTextFormatter
         /// that uses <see cref="Encoding.UTF8"/> for its text encoding.
```
jakebailey: If the default is different, then this comment should probably be updated. I don't `DefaultEncoding` value will be visible to people other than those who read the code.

src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
```diff
@@ -14,4 +14,11 @@ public JsonMessageFormatterTests(ITestOutputHelper logger)
         : base(logger)
     {
     }
+
+    [Fact]
+    public void DefaultEncodingLacksPreamble()
```
jakebailey: Is there any way to run a test that uses the formatter and checks for the lack of a preamble rather than checking the private field directly?

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: Updated. :)

src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
AArnott: It isn't checking a private field. It's checking the public property. But testing that the Encoding property is actually the one used to format the message is a good idea. I'll add that.

jakebailey: Yeah, I goofed and misread the `private Encoding encoding` line thinking `Encoding` was the name...

PR #232, opened 2019-02-02
Beef up docs default and non-default protocol settings

### LINKED ISSUE #230: Documentation: Default Protocol Case Poorly Documented
The current documentation does not provide any simple indication of what protocol is used by default (the `Content-Length: X \r\n\r\n[JSON BLOB]` message), nor the default text encoding. 

While this info is available if you go into extensibility page, the examples don't explain, for example, a server must prefix messages with a content-length header (instead of raw json), or they will be ignored. 

Adding a simple 'if you follow the examples in this walkthrough, we send this style of message, and expect this back, with this encoding' to the examples (probably to the connecting section) would make things considerably easier to read.

### LINKED ISSUE #231: Documentation: Changing between included protocols or headers not documented
While the `extensibility` page states there are alternative protocol handlers included in the library, none of the documentation shows how to consume them (not implement them). 

For example, if I wanted to switch to `LengthHeaderMessageHandler` or `HttpClientMessageHandler`, there's nothing in the documentation that says how to go from the simple documented `JsonRpc rpc = JsonRpc.Attach(stream)` connection method to an alternative protocol.

AArnott:
Fixes #230
Fixes #231

### DIFF
--- a/doc/connecting.md
+++ b/doc/connecting.md
@@ -14,7 +14,7 @@ Certain decisions about the protocol details must be made up front while constru
 This library includes several built-in protocol variants and options, and you can add your own. This is all documented in our [extensibility](extensibility.md) document.
 
 In the remaining samples in this section we use the convenient static `Attach` method which
-instantiates the class with the default settings and begins listening for messages immediately. Samples for changing aspects of the protocol are further below in the next section.
+instantiates the class with the default settings and begins listening for messages immediately. Samples for changing aspects of the protocol are in [a section below](#Configuring).
 
 ### Client
 
@@ -49,9 +49,9 @@ and before disposing the stream, you can hold the connection open as long as the
 await rpc.Completion;
 ```
 
-## Configuring/customizing the protocol
+## Configuring/customizing the protocol <a name="Configuring"></a>
 
-To alter the protocol in any way from the defaults, use the constructor directly instead.
+To alter the protocol in any way from the defaults, use the `JsonRpc` constructor directly, instead of using the static `Attach` method.
 This gives you a chance to provide your own `IJsonRpcMessageHandler`, set text encoding, etc. before sending or receiving any messages.
 Remember after configuring your instance to start listening by calling the `JsonRpc.StartListening()` method. This step is not necessary when using the static `Attach` method because it calls `StartListening` for you.
 
--- a/doc/connecting.md
+++ b/doc/connecting.md
@@ -7,12 +7,16 @@ or a pair of half-duplex `Stream` objects (e.g. STDIN and STDOUT streams). All A
 this document will assume a single bidirectional `Stream` for brevity.
 
 A JSON-RPC connection is created and managed via the `JsonRpc` class.
-Certain decisions about the protocol details must be made up front while constructing the `JsonRpc` class.
-These decisions are documented in our [extensibility](extensibility.md) document.
-For this document, we will use the defaults for these up-front decisions.
-Other configurations can be changed after instantiating the `JsonRpc` class. We will review these in this document.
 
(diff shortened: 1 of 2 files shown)


AArnott:
@iUnknwn I'd love to get your take on whether you feel these docs are adequate.

doc/connecting.md
```diff
@@ -7,12 +7,16 @@ or a pair of half-duplex `Stream` objects (e.g. STDIN and STDOUT streams). All A
...
+
 Certain decisions about the protocol details must be made up front while constructing the `JsonRpc` class.
-These decisions are documented in our [extensibility](extensibility.md) document.
-For this document, we will use the defaults for these up-front decisions.
-Other configurations can be changed after instantiating the `JsonRpc` class. We will review these in this document.
+This library includes several built-in protocol variants and options, and you can add your own. This is all documented in our [extensibility](extensibility.md) document.
 
-## Client
+In the remaining samples in this section we use the convenient static `Attach` method which
+instantiates the class with the default settings and begins listening for messages immediately. Samples for changing aspects of the protocol are further below in the next section.
```
ikeras: You may want to mention, or link to, the actual section. That way, if you ever insert another section in between, you don't have to remember to update this text.

doc/connecting.md
```diff
@@ -45,6 +49,41 @@ and before disposing the stream, you can hold the connection open as long as the
 await rpc.Completion;
 ```
 
+## Configuring/customizing the protocol
+
+To alter the protocol in any way from the defaults, use the constructor directly instead.
```
ikeras: I would say "use the JsonRpc constructor directly, instead of using the Attach method"

doc/connecting.md
AArnott: I wish I knew a good way to link _within_ a markdown document. I've seen ways documented, but haven't found one that works on github hosted docs.

AArnott: OK, I found at least one way. I've applied it.

doc/connecting.md
AArnott: Done.

iUnknwn:
This is a terrific improvement - thank you.

The only comment I'd have is it would be great to just include a small note that the procedure shown in `Configuring` is the client use case - to do the same for the server, you'd call `new JsonRpc(handler, target_obj)`.

Otherwise, this looks great.

AArnott:
Thanks for the feedback, @iUnknwn. Does #233 address your concern?

iUnknwn:
My recommendation would be to be explicit about it - something more like:

A client can switch to the alternate header format by doing:
```cs
Stream send, recv;
var formatter = new JsonMessageFormatter(Encoding.UTF8);
var handler = new LengthHeaderMessageHandler(send, recv, formatter);
// for servers, modify the constructor to include a target object (details below)
var jsonRpc = new JsonRpc(handler);
jsonRpc.StartListening();
```

For a server, the code is identical, except that a target object should be added (either by adding a target in the `JsonRpc` constructor (`new JsonRpc(handler, target_object)`, or via `[alternative methods]`). 

This way, there's a clear set of instructions for each use case (and less ambiguity on how the procedure would work for the other)

AArnott:
Thanks. I hesitate to go further to distinguish the server vs. client code because json-rpc is a peer to peer protocol so either or both sides can offer target methods and objects. The API for both sides are the same. So rather than split out their use cases as if client and server were very different, I prefer to emphasize their similarities. But it's all pointless if folks can't understand the docs.
If you believe you could get what you need from the docs when factored this way, then I'm satisfied, even if it isn't ideal for you. We can make further changes if we get more feedback from others.

PR #239, opened 2019-02-12
Fix handling of nullable value types as RPC method parameters and return types

### LINKED ISSUE #237: InvalidOperationException on null result for InvokeAsync<int?>
Modifying the Sample Client and Server to use the latest version, and to return int? rather than int, with the server returning null, results in an InvalidOperationException on the client side.

seems like the problem is:

https://github.com/Microsoft/vs-streamjsonrpc/blob/0bf2419fb161b5b55bbc62e43a57419c0a93708d/src/StreamJsonRpc/JsonMessageFormatter.cs#L465

but int? is a value type, yet null is assignable to this type.

perhaps it should be:
`Verify.Operation(!typeof(T).GetTypeInfo().IsValueType && Nullable.GetUnderlyingType(typeof(T)) != null, "null result is not assignable to a value type.");`

AArnott:
Both return types and value types were failing when we check that a value is assignable to the required CLR type because we weren't expressly checking that a value type is a nullable one, and we assumed all value types cannot be `null`.

Fixes #237

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -393,6 +393,26 @@ public abstract class JsonRpcTests : TestBase
     }
 
     [Fact]
+    public async Task NullableParameters()
+    {
+        int? result = await this.clientRpc.InvokeAsync<int?>(nameof(Server.MethodAcceptsNullableArgs), null, 3);
+        Assert.Equal(1, result);
+        result = await this.clientRpc.InvokeAsync<int?>(nameof(Server.MethodAcceptsNullableArgs), 3, null);
+        Assert.Equal(1, result);
+        result = await this.clientRpc.InvokeAsync<int?>(nameof(Server.MethodAcceptsNullableArgs), 3, 5);
+        Assert.Equal(2, result);
+    }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1339,6 +1359,10 @@ public static int MethodWithDefaultParameter(int x, int y = 10)
             return x + y;
         }
 
+        public int? MethodReturnsNullableInt(int a) => a > 0 ? (int?)a : null;
+
+        public int MethodAcceptsNullableArgs(int? a, int? b) => (a.HasValue ? 1 : 0) + (b.HasValue ? 1 : 0);
```
ikeras: nit: For this test method, I think it would be better to have:

```
public int MethodAcceptsNullableArgs(int? a, int? b) => (a.HasValue ? a.Value : 0) + (b.HasValue ? b.Value : 0);
```

That way you can further validate that you don't accidently have an issue where you're swapping the values of a & b.

AArnott: I considered that, but the problem is that then I can't discern between 0 and null being received by the server method. I very much wanted to assert that the server method can tell the difference between a `default(int)` and `default(int?)`.

PR #248, opened 2019-03-08
Package metadata and TFM update

AArnott:
1. Remove unnecessary `#if NETFRAMEWORK` exclusion around exception serialization. This brings our netstandard2.0 build to exactly the same functionality as our net46 build. We keep our net46 build though since netstandard2.0 doesn't work on net46.
1. Remove bad release notes link from package metadata
1. Add net472 build
1. Test net472 build and the netcoreapp build on netcoreapp2.2.


tmat:
@AArnott Do you have an ETA when the new packages is gonna be available on nuget.org and the net472 version inserted to VS? (no rush, just curious)

AArnott:
I'm hoping to get [these issues](https://github.com/Microsoft/vs-streamjsonrpc/issues?q=is%3Aopen+is%3Aissue+milestone%3Av2.0) fixed in the next couple days before the next release. If I can't do that, we can release what we have now.

tmat:
Sounds good.

PR #250, opened 2019-03-10
Use safe length prefixing technique with IBufferWriter<T>

### LINKED ISSUE #247: Don't use disallowed length prefixing technique
Per https://github.com/dotnet/corefx/issues/34259, we need to scrub out our use of this pattern for length prefixing:

https://github.com/Microsoft/vs-streamjsonrpc/blob/1abe58684493002e67b82cb93ea5584f3e4a4c79/src/StreamJsonRpc/LengthHeaderMessageHandler.cs#L103-L105

AArnott:
Fixes #247

### DIFF
--- a/src/StreamJsonRpc/LengthHeaderMessageHandler.cs
+++ b/src/StreamJsonRpc/LengthHeaderMessageHandler.cs
@@ -32,7 +32,7 @@ namespace StreamJsonRpc
         /// <summary>
         /// A wrapper to use for the <see cref="PipeMessageHandler.Writer"/> when we need to count bytes written.
         /// </summary>
-        private CountingWriterWrapper<byte> writerWrapper;
+        private PrefixingBufferWriter<byte> prefixingWriter;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="LengthHeaderMessageHandler"/> class.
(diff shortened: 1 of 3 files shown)


src/StreamJsonRpc/LengthHeaderMessageHandler.cs
```diff
@@ -32,7 +32,7 @@ public class LengthHeaderMessageHandler : PipeMessageHandler
         /// <summary>
         /// A wrapper to use for the <see cref="PipeMessageHandler.Writer"/> when we need to count bytes written.
         /// </summary>
-        private CountingWriterWrapper<byte> writerWrapper;
+        private PrefixingBufferWriter<byte> prefixingWriter;
```
ikeras: Neat! So this reserves some amount of space at the start that you can write later?

AArnott: Yes. And it's optimized so that no extra buffer copying is required, provided the payload that follows the header doesn't exceed a predicted length (otherwise 1 copy of the payload that exceeds that prediction is required).

AArnott: And David Fowler has expressed an interest in me contributing this class to corefx.

PR #261, opened 2019-04-17
LOC CHECKIN | Microsoft/vs-streamjsonrpc master | 20190417

v-zbsail:


AArnott:
Thanks. I've rebased your commit onto the v1.5 branch (the oldest branch we support where these translation updates would apply), updated resx files, and [pushed](4247ce0).

PR #264, opened 2019-05-02
Throw OCE from canceled requests even when remote disconnects before sending cancellation acknowledgment

richardstanton:
After canceling a request, frequently the client gets disposed. In 1.5 this would result in an expected cancelation state. In 2.0 this started throwing a ConnectionLostException if the dispose happened subsequently. Since the lost connection was a result of canceled request and not the cause of it, this adjusts the state of the task to match the expected state.

Tests were added to verify dispose does the expected behavior, as well as a lost connection still results in the ConnectionLostException.

* [x] Ask mode approval obtained from @Michael-Eng

Fixes: #265

@AArnott @tinaschrepfer

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
@@ -150,7 +150,7 @@ public class JsonRpcWithFatalExceptionsTests : TestBase
     }
 
     [Fact]
-    public async Task CancelMayStillReturnErrorFromServer()
+    public async Task CancelExceptionPreferredOverConnectionLost()
     {
         using (var cts = new CancellationTokenSource())
         {
@@ -159,7 +159,10 @@ public class JsonRpcWithFatalExceptionsTests : TestBase
             cts.Cancel();
             this.server.AllowServerMethodToReturn.Set();
 
-            await Assert.ThrowsAnyAsync<OperationCanceledException>(() => invokeTask);
+            // When the remote hangs up while the local side has an outstanding request,
+            // we expect ConnectionLostException to be thrown locally unless the request was already canceled anyway.
+            var ex = await Assert.ThrowsAnyAsync<OperationCanceledException>(() => invokeTask);
+            Assert.Equal(cts.Token, ex.CancellationToken);
             Assert.Equal(Server.ThrowAfterCancellationMessage, this.serverRpc.FaultException.Message);
             Assert.Equal(1, this.serverRpc.IsFatalExceptionCount);
         }
--- a/src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
@@ -159,9 +159,53 @@ public class JsonRpcWithFatalExceptionsTests : TestBase
             cts.Cancel();
             this.server.AllowServerMethodToReturn.Set();
 
-            await Assert.ThrowsAnyAsync<ConnectionLostException>(() => invokeTask);
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(() => invokeTask);
             Assert.Equal(Server.ThrowAfterCancellationMessage, this.serverRpc.FaultException.Message);
             Assert.Equal(1, this.serverRpc.IsFatalExceptionCount);
+        }
+
+        Assert.True(((IDisposableObservable)this.clientMessageHandler).IsDisposed);
+        Assert.True(((IDisposableObservable)this.serverMessageHandler).IsDisposed);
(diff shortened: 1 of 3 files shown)


richardstanton:
Associated bug:
#265

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
```diff
@@ -170,6 +170,50 @@ public async Task CancelMayStillReturnErrorFromServer()
...
+        }
+
+        Assert.True(((IDisposableObservable)this.clientMessageHandler).IsDisposed);
+        Assert.True(((IDisposableObservable)this.serverMessageHandler).IsDisposed);
+        Assert.True(this.clientRpc.IsDisposed);
+        Assert.False(this.serverRpc.IsDisposed);
+    }
+
+    [Fact]
+    public async Task UnexpectedDisconnectResultsInConnectionLostException()
```
AArnott: I'm glad you included this test, but it isn't really about fatal server exceptions--just about the server hanging up. Can you move this to `JsonRpcTests.cs` and simplify (e.g. remove the cts, dispose of `serverRpc` itself, etc.)?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1076,7 +1076,15 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
                                     this.TraceSource.TraceEvent(TraceEventType.Warning, (int)TraceEvents.RequestAbandonedByRemote, "Aborting pending request \"{0}\" because the connection was lost.", id);
                                 }
 
-                                tcs.TrySetException(new ConnectionLostException());
+                                if (cancellationToken.IsCancellationRequested || this.IsDisposed)
```
AArnott: The ` || this.IsDisposed` is beyond the scope of #265 as I understand it. And if we're careful about which exception we throw in that case, perhaps we should be throwing `ObjectDisposedException` instead since the object was disposed while methods were still running inside the local instance.

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
```diff
@@ -170,6 +173,50 @@ public async Task CancelMayStillReturnErrorFromServer()
         Assert.False(this.serverRpc.IsDisposed);
     }
 
+    [Fact]
+    public async Task DisposedClientResultsInCancelled()
```
AArnott: This isn't related to server fatal exceptions. Can you move it to `JsonRpcTests.cs`?

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
```diff
@@ -170,6 +173,50 @@ public async Task CancelMayStillReturnErrorFromServer()
...
+            this.clientRpc.Dispose();
+            this.server.AllowServerMethodToReturn.Set();
+
+            // Connection was closed before error was sent from the server
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(() => invokeTask);
+            Assert.Equal(0, this.serverRpc.IsFatalExceptionCount);
+        }
+
+        Assert.True(((IDisposableObservable)this.clientMessageHandler).IsDisposed);
+        Assert.True(((IDisposableObservable)this.serverMessageHandler).IsDisposed);
```
AArnott: This line caused the test to fail non-deterministically. There's no guarantee at this point that the server has already been disposed of, since it responds to the client's disconnection asynchronously.

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
AArnott: In fact we have other tests that verify this (e.g. `CloseStreamsOnSynchronousMethodException`) so I think we can just remove it.

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
AArnott: I took care of it.

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
AArnott: I removed this line.

src/StreamJsonRpc/JsonRpc.cs
AArnott: I removed the disposal check.

src/StreamJsonRpc.Tests/JsonRpcWithFatalExceptionsTests.cs
AArnott: I removed this.

PR #270, opened 2019-05-12
Fix race condition in StartListening

### LINKED ISSUE #269: StartListening race condition may cause InvokeAsync to throw
`StartListening` contains this line:

https://github.com/microsoft/vs-streamjsonrpc/blob/b187fbb6ad7271dbb19c3ef0eb1f99079a180459/src/StreamJsonRpc/JsonRpc.cs#L720

This causes another thread to start receiving messages and invoking local RPC methods. If one of those methods call `JsonRpc.InvokeAsync` on the same `JsonRpc` instance, and if the line above hasn't yet completed the assignment of the `Task` to the `readLinesTask` field, then this line in `InvokeCoreAsync` will throw:

https://github.com/microsoft/vs-streamjsonrpc/blob/b187fbb6ad7271dbb19c3ef0eb1f99079a180459/src/StreamJsonRpc/JsonRpc.cs#L1105

We need to guarantee that `StartListening` doesn't race itself against the task it schedules causing non-deterministic failures.

AArnott:
`StartListening` sets a `readLinesTask` field to the result of a `Task.Run` call that kicks off the listener. But if the listener starts, receives its first request, invokes the local RPC method, which then calls `JsonRpc.InvokeAsync`, all before the original thread that called `StartListening` is able to execute the field assignment, `InvokeAsync` will throw because it thinks `StartListening` has not yet been called.

The fix I've chosen is to put a couple of extra locks in place to guarantee the field assignment is executed before the listening method actually receives its first message.

I've also removed the `startedListening` field which is redundant with a `readLinesTask!=null` check.

Fixes #269

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -92,7 +92,6 @@ namespace StreamJsonRpc
         private int nextId = 1;
         private bool disposed;
         private bool hasDisconnectedEventBeenRaised;
-        private bool startedListening;
 
         /// <summary>
         /// Backing field for the <see cref="CancelLocallyInvokedMethodsWhenConnectionIsClosed"/> property.
@@ -230,7 +229,7 @@ namespace StreamJsonRpc
         {
             get
             {
-                Verify.Operation(this.startedListening, Resources.MustBeListening);
+                Verify.Operation(this.HasListeningStarted, Resources.MustBeListening);
                 return this.completionSource.Task;
             }
         }
@@ -290,6 +289,11 @@ namespace StreamJsonRpc
         /// Gets the user-specified <see cref="SynchronizationContext"/> or a default instance that will execute work on the threadpool.
         /// </summary>
         private SynchronizationContext SynchronizationContextOrDefault => this.SynchronizationContext ?? DefaultSynchronizationContext;
+
+        /// <summary>
+        /// Gets a value indicating whether listening has started.
+        /// </summary>
+        private bool HasListeningStarted => this.readLinesTask != null;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="JsonRpc"/> class and immediately starts listening.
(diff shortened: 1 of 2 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1333,6 +1343,15 @@ private void OnJsonRpcDisconnected(JsonRpcDisconnectedEventArgs eventArgs)
 
         private async Task ReadAndHandleRequestsAsync()
         {
+            lock (this.syncObject)
```
heejaechang: so, this relies on the behavior of Task.Run that it will run func<Task> in another thread? since lock statement allows re-entrance on the same thread. 

if that is guaranteed, seems fine. otherwise, it would still have a chance to have the issue.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -520,12 +524,18 @@ public void AddLocalRpcMethod(string rpcMethodName, MethodInfo handler, object t
...
             Verify.NotDisposed(this);
-            this.readLinesTask = Task.Run(this.ReadAndHandleRequestsAsync, this.disposeCts.Token);
+
+            // We take a lock around this Task.Run and field assignment,
+            // and also immediately within the invoked Task itself,
+            // to guarantee that the assignment will complete BEFORE we actually read the first message.
+            // See the StartListening_ShouldNotAllowIncomingMessageToRaceWithInvokeAsync test.
+            lock (this.syncObject)
+            {
+                this.readLinesTask = Task.Run(this.ReadAndHandleRequestsAsync, this.disposeCts.Token);
```
heejaechang: just to make sure Task.Run run things in other thread, put TaskScheduler.Default explicitly?

heejaechang: ah. ignore what I wrote. looks like Task.Run always use TaskScheduler.Default and you can't change it.

src/StreamJsonRpc/JsonRpc.cs
AArnott: Yes, `Task.Run`'s documented behavior is to schedule work on `TaskScheduler.Default` (the thread pool). It never inlines the delegate passed to it.

PR #279, opened 2019-05-17
JsonMessageFormatter GC pressure fixes

### LINKED ISSUE #276: JsonMessageFormatter.WriteJToken bleeds memory pressure
We need to stop creating a new `StreamWriter` for every message we write. It's bleeding *far* too much memory each time by allocating a `byte[]` and `char[]` buffer.

Name | Inc % | Inc
-- | -- | --
streamjsonrpc.ni!JsonMessageFormatter.WriteJToken | 36.2 | 673,039,680
|+ mscorlib.ni!System.IO.StreamWriter..ctor(System.IO.Stream, System.Text.Encoding, Int32) | 32.6 | 607,224,576
|+ clr!JIT_New | 3.1 | 57,478,312
|+ mscorlib.ni!System.IO.StreamWriter.Flush(Boolean, Boolean) | 0.3 | 4,690,312
|+ clr!JIT_NewCrossContext_Portable | 0.1 | 1,926,104
|+ newtonsoft.json.ni!JObject.WriteTo | 0.1 | 1,720,224

This is possibly a significant perf regression since our v1.5 version.

AArnott:
This dramatically reduces GC pressure from `JsonMessageFormatter`.

1. Use shared array pool and recycle `Sequence<byte>` instance
1. Avoid allocating `Task<JsonRpcMessage>` for non-yielding server methods
1. Replace StreamWriter with BufferTextWriter
For each outgoing message, `JsonMessageFormatter` was creating a new `StreamWriter`, which allocated two large buffers. Since the `IBufferWriter<byte>` instance may vary across calls, we can't simply reuse a `StreamWriter` to avoid the allocations.
Instead we use a `TextWriter`-derived class which IS reusable across different target `IBufferWriter<byte>` instances. We allocate only one (char[]) buffer and encode directly to the memory returned from `IBufferWriter<byte>.GetMemory` so it's faster as well as much more GC friendly than the `StreamWriter`.
1. Use `SequenceTextReader` to reduce GC pressure  
In `JsonMessageFormatter`, creating a new `StreamReader` for every `ReadOnlySequence<byte>` is a very expensive process, as each new `StreamReader` allocates large `byte[]` and `char[]` arrays to use as buffers. The new `SequenceTextReader` only allocates one `char[]` array and that array is reused for every `ReadOnlySequence<byte>` to be read.
1.  Utilize `char[]` pooling for `JsonTextReader`
1. Avoid JsonSerializer allocation for each transmission
1.  Avoid linking `CancellationTokens` in `MessageHandlerBase`
It might be nice to forcefully cancel folks when the `MessageHandlerBase` is disposed of, but linking tokens requires memory allocations. Practically speaking, the only caller of these methods is `JsonRpc`, which itself will cancel the token it explicitly passed to these methods at about the time it disposes of the `MessageHandlerBase` anyway. So there's no reason to allocate in order to dispose in *both* ways at once.    

Fixes #276

This reduces GC pressure for small messages by 6X on .NET Framework and 10X on .NET Core. 
To give a clear idea of the "regression" from v1.5 to v2.0, and how this PR fixes it (and then goes further to be even better than 1.5), see the tables below. Particularly the last column regarding allocated memory.

v1.5

|             Method | Runtime |     Toolchain |     Mean |    Error |   StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------- |-------- |-------------- |---------:|---------:|---------:|------:|--------:|------------:|------------:|------------:|--------------------:|
| InvokeAsync_NoArgs |     Clr |        net472 | 72.63 us | 1.365 us | 1.402 us |  1.00 |    0.00 |      4.6387 |           - |           - |            19.13 KB |
| InvokeAsync_NoArgs |    Core | netcoreapp2.2 | 98.32 us | 1.950 us | 4.784 us |  1.26 |    0.06 |      3.9063 |           - |           - |             3.88 KB |

v2.0:

|             Method | Runtime |     Toolchain |     Mean |    Error |    StdDev |   Median | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------- |-------- |-------------- |---------:|---------:|----------:|---------:|------:|--------:|------------:|------------:|------------:|--------------------:|
| InvokeAsync_NoArgs |     Clr |        net472 | 260.0 us | 41.86 us | 123.41 us | 308.0 us |  1.00 |    0.00 |     20.7520 |      0.1221 |           - |             85.4 KB |
| InvokeAsync_NoArgs |    Core | netcoreapp2.2 | 333.1 us | 11.43 us |  32.98 us | 328.0 us |  1.76 |    1.01 |     20.0195 |           - |           - |            32.49 KB |

fix276:

|             Method | Runtime |     Toolchain |      Mean |    Error |   StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------------- |-------- |-------------- |----------:|---------:|---------:|------:|--------:|------------:|------------:|------------:|--------------------:|
| InvokeAsync_NoArgs |     Clr |        net472 |  81.25 us | 1.551 us | 1.786 us |  1.00 |    0.00 |      3.1738 |           - |           - |            13.34 KB |
| InvokeAsync_NoArgs |    Core | netcoreapp2.2 | 120.44 us | 2.368 us | 5.149 us |  1.47 |    0.09 |      2.6855 |           - |           - |             3.39 KB |

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -11,6 +11,7 @@ namespace StreamJsonRpc
     using System.IO;
     using System.Linq;
     using System.Reflection;
+    using System.Runtime.InteropServices;
     using System.Runtime.Serialization;
     using System.Text;
     using System.Threading;
@@ -40,6 +41,11 @@ namespace StreamJsonRpc
         /// UTF-8 encoding without a preamble.
         /// </summary>
         private static readonly Encoding DefaultEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+
+        /// <summary>
+        /// The reusable <see cref="TextWriter"/> to use with newtonsoft.json's serializer.
+        /// </summary>
+        private readonly BufferTextWriter bufferTextWriter = new BufferTextWriter();
 
         /// <summary>
         /// The version of the JSON-RPC protocol being emulated by this instance.
@@ -262,13 +268,11 @@ namespace StreamJsonRpc
 
         private void WriteJToken(IBufferWriter<byte> contentBuffer, JToken json)
         {
-            using (var streamWriter = new StreamWriter(contentBuffer.AsStream(), this.Encoding, 4096))
+            this.bufferTextWriter.Initialize(contentBuffer, this.Encoding);
+            using (var jsonWriter = new JsonTextWriter(this.bufferTextWriter))
             {
-                using (var jsonWriter = new JsonTextWriter(streamWriter))
-                {
-                    json.WriteTo(jsonWriter);
-                    jsonWriter.Flush();
-                }
+                json.WriteTo(jsonWriter);
+                jsonWriter.Flush();
             }
         }
 
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -48,6 +48,16 @@ namespace StreamJsonRpc
         private static readonly IArrayPool<char> JsonCharArrayPool = new JsonArrayPool<char>(ArrayPool<char>.Shared);
 
         /// <summary>
+        /// An exactly default instance of the <see cref="JsonSerializer"/> to use where no special settings
+        /// are needed.
+        /// </summary>
+        /// <remarks>
+        /// This is useful when calling such APIs as <see cref="JToken.FromObject(object, JsonSerializer)"/>
+        /// because <see cref="JToken.FromObject(object)"/> allocates a new serializer with each invocation.
+        /// </remarks>
+        private static readonly JsonSerializer DefaultSerializer = JsonSerializer.CreateDefault();
+
+        /// <summary>
         /// The reusable <see cref="TextWriter"/> to use with newtonsoft.json's serializer.
         /// </summary>
         private readonly BufferTextWriter bufferTextWriter = new BufferTextWriter();
@@ -205,7 +215,7 @@ namespace StreamJsonRpc
             // Pre-tokenize the user data so we can use their custom converters for just their data and not for the base message.
             this.TokenizeUserData(message);
 
-            var json = JToken.FromObject(message);
+            var json = JToken.FromObject(message, DefaultSerializer);
 
             // Fix up dropped fields that are mandatory
             if (message is Protocol.JsonRpcResult && json["result"] == null)
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -48,6 +48,11 @@ namespace StreamJsonRpc
         private readonly BufferTextWriter bufferTextWriter = new BufferTextWriter();
 
         /// <summary>
+        /// The reusable <see cref="TextReader"/> to use with newtonsoft.json's deserializer.
+        /// </summary>
+        private readonly SequenceTextReader sequenceTextReader = new SequenceTextReader();
+
+        /// <summary>
         /// The version of the JSON-RPC protocol being emulated by this instance.
         /// </summary>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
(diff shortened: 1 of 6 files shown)


AArnott:
CC: @heejaechang @sharwell @RyanToth3

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -467,5 +471,218 @@ public override T GetResult<T>()
...
+                        {
+                            ArrayPool<byte>.Shared.Return(rentedByteBuffer);
+                        }
+                    }
+
+                    this.charBufferPosition = 0;
+
+                    if (this.memoryPosition == this.memory.Length)
+                    {
+                        this.Flush();
```
dgriffen: Why is this done recursively instead of iteratively?

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -467,5 +471,218 @@ public override T GetResult<T>()
                 return result.ToObject<T>(this.jsonSerializer);
             }
         }
+
+        private class BufferTextWriter : TextWriter
```
AArnott: A future iteration of this PR removes this class in favor of consuming a [more complete and tested version from Nerdbank.Streams](https://github.com/AArnott/Nerdbank.Streams/pull/71/files#diff-dee8a6ac9111d50058f2955c13f847da).

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: I don't see recursion or iteration here. The calling method is tasked with encoding `char` to `byte`. When the `byte[]` buffer is full, we need to flush it so there's room for more.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -467,5 +501,19 @@ public override T GetResult<T>()
                 return result.ToObject<T>(this.jsonSerializer);
             }
         }
+
+        private class JsonArrayPool<T> : IArrayPool<T>
```
dgriffen: What is the purpose of this class? It seems to directly wrap `ArrayPool`.

ikeras: It's not clear why you need this from just reading the code; but looks like IArrayPool is actually a NewtonSoft thing and you created this as a shim between System.Buffers.ArrayPool and NewtonSoft.Json.IArrayPool. If so, you may want to add a comment here to make that clear.

AArnott: indeed. That's all it does. But as it implements newtonsoft's `IArrayPool<T>` interface, it allows us to use the .NET `ArrayPool<T>` type for the pool we offer to newtonsoft. The .NET class of course does not implement newtonsoft's interface directly.

AArnott: Apparently so, since both you and @dgriffen had the same concern.

AArnott: I've added an xml doc comment on the class.

src/StreamJsonRpc/StreamJsonRpc.csproj
```diff
@@ -36,7 +36,7 @@
     <PackageReference Include="System.Net.WebSockets" Version="4.3.0" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.2" />
     <PackageReference Include="System.IO.Pipelines" Version="4.5.3" />
-    <PackageReference Include="Nerdbank.Streams" Version="2.1.37" />
+    <PackageReference Include="Nerdbank.Streams" Version="2.2.5-alpha" />
```
AArnott: We can't introduce a dependency on an unstable package in our stable 2.0 branch. I need to either retarget this PR to `master` or bump this to a stable version of Nerdbank.Streams when that's available.

PR #282, opened 2019-05-20
Pin the .NET Core SDK used to build

AArnott:
Also install the .NET Core SDK before using it


azure-pipelines/build.yml
```diff
@@ -1,10 +1,22 @@
...
+  displayName: Show dotnet SDK info
+
 - script: |
     dotnet tool install --tool-path .. nbgv
     ..\nbgv cloud
     echo ##vso[task.setvariable variable=PATH]%Path%;$(Build.SourcesDirectory)
   workingDirectory: src
   displayName: Set build number
+  env:
+    DOTNET_MULTILEVEL_LOOKUP: 0 # workaround .NET CLI bug that shows up in Azure Pipelines
```
AArnott: This is the workaround for https://github.com/dotnet/cli/issues/11380

PR #291, opened 2019-06-09
Be more liberal with the accepted type for a message id

robmen:
Some servers do not strictly follow the JSON-RPC spec and stringify the
id, even if the client sent it as a number.

Note: We've already filed a bug against the service that stringifies all ids but we're not expecting a fix from them soon.

PS: I know you are not accepting pull requests yet but wanted to get this on your radar.


AArnott:
De minimis PRs like this are fine. And thanks for filing the bug against the other server. Your justification for this change seems reasonable.

We just need to add a test to https://github.com/microsoft/vs-streamjsonrpc/blob/master/src/StreamJsonRpc.Tests/JsonRpcServerInteropTests.cs

PR #304, opened 2019-07-04
Add support for ValueTask and ValueTask<T> returning methods

### LINKED ISSUE #302: Support for `ValueTask`/`ValueTask<T>` returning RPC methods
It's not that in a remoted case a single `Task<object>` allocation is likely to hurt in the grand scheme of things. But when folks use interfaces to call the server, and the server *might* actually be present locally instead of remoted, then `ValueTask` return values allow alloc-free calls to these local server methods.

AArnott:
This is *not* more efficient than traditional Task. It's worse. But it allows proxy generation to support interfaces with ValueTask returning methods which is useful because those interfaces MAY not always be remoted. If the interface is there to support remoting but may also work directly, then ValueTask can provide a significant improvement to GC pressure since arbitrary values can be returned without allocating a Task<T> object.

Closes #302

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -1329,14 +1329,14 @@ namespace StreamJsonRpc
         /// <param name="taskTypeInfo">The original type of the value returned from an RPC-invoked method.</param>
         /// <param name="taskOfTTypeInfo">Receives the <see cref="Task{T}"/> type that is a base type of <paramref name="taskTypeInfo"/>, if found.</param>
         /// <returns><c>true</c> if <see cref="Task{T}"/> could be found in the type hierarchy; otherwise <c>false</c>.</returns>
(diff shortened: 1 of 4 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1580,8 +1614,8 @@ private JsonRpcMessage HandleInvocationTaskOfTResult(JsonRpcRequest request, Tas
 
             if (message is JsonRpcResult resultMessage)
             {
-                // This method should only be called for methods that declare to return Task<T> or a derived type.
-                Assumes.True(TryGetTaskOfTType(t.GetType().GetTypeInfo(), out TypeInfo taskOfTTypeInfo));
+                // This method should only be called for methods that declare to return Task<T> or a derived type or ValueTask<T>.
```
ikeras: nit: "// This method should only be called for methods that declare to return Task<T>, a derived type, or ValueTask<T>."

AArnott: Fixed (not quite that way, but I hope it fits what you were going for).

PR #311, opened 2019-07-31
Add async JSON deserialization support

### LINKED ISSUE #310: JsonMessageFormatter should support async streaming deserialization
When a Pipe is the native transport being read from and a very large message is written and flushed in several distinct parts, the handler/formatter style of letting the PipeReader buffer everything so we can deserialize synchronously leads us to run into the Pipe buffer size limit (which defaults to 32KB). Where possible we should allow async streaming deserialization so that the limit doesn't throw an exception.

AArnott:
This replaces buffering large messages with async streaming deserialization where the formatter supports it.
The JsonMessageFormatter supports it with this change.

Fixes #310

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -9,9 +9,9 @@ namespace StreamJsonRpc
     using System.Diagnostics;
     using System.Globalization;
     using System.IO;
+    using System.IO.Pipelines;
     using System.Linq;
     using System.Reflection;
-    using System.Runtime.InteropServices;
     using System.Runtime.Serialization;
     using System.Text;
     using System.Threading;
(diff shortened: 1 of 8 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -25,7 +25,7 @@ namespace StreamJsonRpc
     /// <summary>
     /// Uses Newtonsoft.Json serialization to serialize <see cref="JsonRpcMessage"/> as JSON (text).
```
BertanAygun: Do Dispose and Close methods do different things for this object?

AArnott: Which object?

AArnott: Close and Dispose should be equivalent on any I/O object that I'm familiar with.

BertanAygun: The comment was supposed to be on the line with var jsonReader = new JsonTextReader. I was wondering why you used finally instead of using

AArnott: I've switched to the using block syntax.

PR #314, opened 2019-08-02
Fixed a small bug in the decoding of JSON-RPC 1.0 result messages

haakoo:
The code did not decode JSON-RPC 1.0 result (response) messages. It did instead decode error messages as a result message.

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -196,7 +196,7 @@ namespace StreamJsonRpc
                         this.VerifyProtocolCompliance(json["id"] != null, json, "\"id\" property missing.");
                         return
                             json["method"] != null ? this.ReadRequest(json) :
-                            json["error"]?.Type == JTokenType.Null ? this.ReadResult(json) :
+                            json["result"]?.Type == JTokenType.Null ? this.ReadResult(json) :
                             json["error"]?.Type != JTokenType.Null ? (JsonRpcMessage)this.ReadError(json) :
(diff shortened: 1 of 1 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -196,7 +196,7 @@ public JsonRpcMessage Deserialize(JToken json)
                         this.VerifyProtocolCompliance(json["id"] != null, json, "\"id\" property missing.");
                         return
                             json["method"] != null ? this.ReadRequest(json) :
-                            json["error"]?.Type == JTokenType.Null ? this.ReadResult(json) :
+                            json["result"]?.Type == JTokenType.Null ? this.ReadResult(json) :
```
AArnott: Your change would cause us to try to read a result if and only if the result is null. Yet the JSON-RPC 1.0 spec dictates that the only way we can tell if a response message is an error vs. a result is if `error == null`.  A result in a successful response can be `null`, so we check `error` first to determine whether the message is an error or a response.

haakoo: Right, my mistake. But now we might end up with an `JsonRpcError` even though the error value is null. Which results in an "InvalidOperationException" when the code tries to extract information from a non existing error-object. But since 1.0 isn't a priority, I'll work around it.

AArnott: I don't understand why the problem you say is there would be there. We don't end up with a `JsonRpcError` when `error: null`. In fact right now (if you read the red line instead of the green) you can clearly see that if `error: null` we `ReadResult`. So what are you talking about?

PR #316, opened 2019-08-07
Added lock form bufferTextWritter

milopezc:
This a fix for Bug 959430: StreamJsonRpc - System.InvalidOperationException : This instance must be flushed before being reinitialized. (https://devdiv.visualstudio.com/DevDiv/_workitems/edit/959430)


AArnott:
Rejecting because thread-safety is guaranteed by the formatter's caller.

PR #318, opened 2019-08-12
Added try finally block to ensure writter is being flushed

milopezc:
Added try finally block to ensure writter is being flushed.
If an exception happens when writting to the json we need to make sure the writter gets flushed, otherwise initializing will keep failing.

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -1362,6 +1362,26 @@ public abstract class JsonRpcTests : TestBase
         Assert.Equal("Success!", result);
     }
 
+    [Fact]
+    public async Task FormatterFatalException()
+    {
+        var streams = Nerdbank.FullDuplexStream.CreateStreams();
+        this.serverStream = streams.Item1;
+        this.clientStream = streams.Item2;
+
+        ExceptionThrowingFormatter serverFormatter = new ExceptionThrowingFormatter();
+        this.serverRpc = new JsonRpc(new HeaderDelimitedMessageHandler(this.serverStream, serverFormatter));
+        this.serverRpc.AddLocalRpcTarget(this.server);
+        this.serverRpc.StartListening();
+
+        this.clientRpc = new JsonRpc(this.clientStream);
+        this.clientRpc.StartListening();
+
+        // Failure on server side should cut the connection
+        serverFormatter.ThrowException = true;
+        await Assert.ThrowsAsync<ConnectionLostException>(() => this.clientRpc.InvokeAsync<string>(nameof(Server.AsyncMethod), "Fail"));
+    }
+
     protected override void Dispose(bool disposing)
     {
         if (disposing)
@@ -1879,7 +1899,7 @@ public abstract class JsonRpcTests : TestBase
 
     private class ExceptionThrowingFormatter : JsonMessageFormatter, IJsonRpcMessageFormatter
     {
-        public bool ThrowException = false;
+        public bool ThrowException;
 
         public new void Serialize(IBufferWriter<byte> bufferWriter, JsonRpcMessage message)
         {
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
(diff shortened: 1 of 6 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1987,13 +1968,29 @@ private object GetMessageJson(JsonRpcMessage message)
...
         {
-            this.TraceMessageSent(message);
+            try
+            {
+                this.TraceMessageSent(message);
+                await this.MessageHandler.WriteAsync(message, cancellationToken).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException)
+            {
+            }
```
AArnott: please rethrow all exceptions you catch. The callers are expecting to be able to discern between success and failure based on the exceptions thrown from this method.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1987,13 +1977,28 @@ private object GetMessageJson(JsonRpcMessage message)
...
         {
-            this.TraceMessageSent(message);
+            try
+            {
+                this.TraceMessageSent(message);
+                await this.MessageHandler.WriteAsync(message, cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception exception)
+            {
+                if ((bool)(this.MessageHandler as IDisposableObservable)?.IsDisposed)
```
AArnott: This will throw if `MessageHandler` does not implement `IDisposableObservable` because you can't cast a `null` to `bool`. 
Instead, try this syntax:

```cs
if ((this.MessageHandler as IDisposableObservable)?.IsDisposed ?? false)
```

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1987,13 +1977,28 @@ private object GetMessageJson(JsonRpcMessage message)
...
+                {
+                    var e = new JsonRpcDisconnectedEventArgs(
+                        string.Format(CultureInfo.CurrentCulture, Resources.ErrorWritingJsonRpcResult, exception.GetType().Name, exception.Message),
```
AArnott: You're using `Resources.ErrorWritingJsonRpcResult` which is specific to result messages, but this could be an error in writing *any* message. Can you define a new string called `ErrorWritingJsonRpcMessage` that is slightly more generic to use here?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1801,16 +1801,6 @@ private async Task HandleRpcAsync(JsonRpcMessage rpc)
...
-                        catch (Exception exception)
-                        {
-                            var e = new JsonRpcDisconnectedEventArgs(
-                                string.Format(CultureInfo.CurrentCulture, Resources.ErrorWritingJsonRpcResult, exception.GetType().Name, exception.Message),
-                                DisconnectedReason.StreamError,
-                                exception);
-
-                            // Fatal error. Raise disconnected event.
-                            this.OnJsonRpcDisconnected(e);
-                        }
```
AArnott: I think we need to bring this back. The `TransmitAsync` method only breaks the connection if the handler deemed the error fatal to the integrity of the stream. But *this* method should deem *all* errors fatal because it means the client will hang while waiting for a response.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1851,4 +1875,20 @@ public override void Post(SendOrPostCallback d, object state)
             base.Post(d, state);
         }
     }
+
+
+    private class ExceptionThrowingFormatter : JsonMessageFormatter, IJsonRpcMessageFormatter
+    {
+        public bool ThrowException = false;
```
AArnott: fxcop nit: don't initialize fields to their default value.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1851,4 +1875,20 @@ public override void Post(SendOrPostCallback d, object state)
...
+
+        public new void Serialize(IBufferWriter<byte> bufferWriter, JsonRpcMessage message)
```
AArnott: I'm *really* surprised this works. the `new` modifier on this method should mean it's entirely independent of all base class methods, and thus I wouldn't expect it to ever be called by the library. How is this getting called?
In any case, I think overriding a method with `override` would be the more appropriate choice instead of `new`.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1338,6 +1339,29 @@ public async Task ExceptionControllingErrorCode()
         Assert.Equal(2, exception.ErrorCode);
     }
 
+    [Fact]
+    public async Task FormatterNonFatalException()
```
AArnott: We need another test like this that makes the exception happen on the *server* side, and we should assert that that *kills* the connection since a failure to send a response is fatal to JSON-RPC.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
milopezc: The trick here is that I made this class also implement IJsonRpcMessageFormatter, is kind of a hacky way to get this called from the message handlers. Without that interface, this never gets called as you said. I tried overriding at first since it seemed like the best for me too, but I get an error saying that I can't override the serialize method because it is not marked as virtual, abstract or override. Should I mark the method in JsonMessageFormatter as override or virtual and use override here?

---
In reply to: [314126468](https://github.com/microsoft/vs-streamjsonrpc/pull/318#discussion_r314126468) [](ancestors = 314126468)

AArnott: Very clever. Your technique is fine. I just hadn't seen it used hardly ever so I didn't think of it to look at it.

src/StreamJsonRpc/Resources.resx
```diff
@@ -281,4 +281,8 @@
   <data name="ParameterObjectsNotSupportedInJsonRpc10" xml:space="preserve">
     <value>JSON-RPC 1.0 does not support named arguments (parameters passed within an object). Use positional arguments (parameter arrays) or set protocol version to 2.0.</value>
   </data>
+  <data name="ErrorWritingJsonRpcMessage" xml:space="preserve">
+    <value>Error writing JSON RPC Result: {0}: {1}</value>
```
AArnott: "Result" here should probably be "message" since it may apply to any message type.

PR #319, opened 2019-08-13
Implement ability to add remote RPC targets

tinaschrepfer:
Implementation according to design review.


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -577,6 +579,13 @@ public T Attach<T>(JsonRpcProxyOptions options)
             return proxy;
         }
 
+        public void Relay(Stream relaySendingStream, Stream relayReceivingStream, object relayTarget = null)
+        {
+            this.relayRpc = JsonRpc.Attach(relaySendingStream, relayReceivingStream, relayTarget);
```
AArnott: Setting the field even if it was set previously sounds like it would be an error. Should we throw `InvalidOperationException` when it has already been set? (tip: add a test for this).

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -577,6 +579,13 @@ public T Attach<T>(JsonRpcProxyOptions options)
...
+
+            this.relayRpc.relayRpc = this;
```
AArnott: This appears to cause a circular loop between the two `JsonRpc` instances in which the original one is imperceptible from the one created here. Is that right (and good)?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1102,6 +1111,21 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
                 request.Arguments = arguments ?? EmptyObjectArray;
             }
 
+            return await this.InvokeCoreAsync<TResult>(request, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected async Task<TResult> InvokeCoreAsync<TResult>(JsonRpcRequest request, CancellationToken cancellationToken)
+        {
+            long id = default(long);
+            if (request.Id is long)
```
AArnott: don't dereference arguments without checking for null first.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1102,6 +1111,21 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
                 request.Arguments = arguments ?? EmptyObjectArray;
             }
 
+            return await this.InvokeCoreAsync<TResult>(request, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected async Task<TResult> InvokeCoreAsync<TResult>(JsonRpcRequest request, CancellationToken cancellationToken)
+        {
+            long id = default(long);
```
AArnott: In the event that `request.Id is null`, this will leave `id=0`, which is *not* the same thing as `null`. Being `null` implies a notification while `0` is a request that expects a response.
Also, in the general relaying configuration, you can't assume that the `request.Id` is even a number. It's allowed to be `string`, and some JSON-RPC servers do that. So if we're relaying messages from non-StreamJsonRpc sources, we'd need to properly handle all legal values for id.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1102,6 +1111,21 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
                 request.Arguments = arguments ?? EmptyObjectArray;
             }
 
+            return await this.InvokeCoreAsync<TResult>(request, cancellationToken).ConfigureAwait(false);
+        }
+
+        protected async Task<TResult> InvokeCoreAsync<TResult>(JsonRpcRequest request, CancellationToken cancellationToken)
```
AArnott: Does this overload need to be `protected` or can it be `private`?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -577,6 +579,13 @@ public T Attach<T>(JsonRpcProxyOptions options)
             return proxy;
         }
 
+        public void Relay(Stream relaySendingStream, Stream relayReceivingStream, object relayTarget = null)
```
AArnott: Please add detailed xml doc comments to this method, including a remarks section that explains the scenarios that it is meant to work with.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: Yup, I haven't gone back and polished the PR.  Just wanted to get a read from you initially to see if this is the right approach.

src/StreamJsonRpc/JsonRpc.cs
tinaschrepfer: It can probably be private.  I originally had a subclass of JsonRpc, but found that was not necessary.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1442,79 +1484,128 @@ private async ValueTask<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcRequ
...
+                        {
+                            long id = Interlocked.Increment(ref this.nextId);
+                            request.Id = id;
+                        }
+
+                        JsonRpcMessage remoteResponse = null;
+                        foreach (var remoteTarget in localRemoteTargets)
+                        {
+                            CancellationToken token = request.IsResponseExpected ? localMethodCancellationSource.Token : CancellationToken.None;
+                            var response = await remoteTarget.InvokeCoreAsync(request, token).ConfigureAwait(false);
```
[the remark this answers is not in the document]
tinaschrepfer: Changed ordering of DispatchIncomingMessageAsync according to our discussion.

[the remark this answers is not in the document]
tinaschrepfer: Addressed this following our Teams conversation.

PR #320, opened 2019-08-15
Add support for IProgress<T> parameters

milopezc:
Added support for IProgress<T>.


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -79,6 +89,11 @@ public class JsonMessageFormatter : IJsonRpcAsyncMessageTextFormatter
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
         private Encoding encoding;
 
+        /// <summary>
+        /// Backing field for the <see cref="Rpc"/> property.
+        /// </summary>
+        private JsonRpc rpc;
+
         /// <summary>
```
AArnott: At this point, it appears the only use we have for this field is to access its `TraceSource`, which is a far looser coupling than `JsonRpc` itself. 
Can we swap this out for `TraceSource` (updating the container interface  and its name as well) so JsonRpc can pass it the TraceSource instead of itself?

AArnott: Well.... never mind. I guess the most important part of this field *isn't* the ability to access it as much as the ability to 'lock in' that it's exactly *one* instance that the formatter is associated with, which is crucial both for thread-safety and for the request IDs that we track.
Can you add a remarks section to this field then, explaining that this assists with these things?

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -445,6 +524,22 @@ private JsonRpcError ReadError(JToken json)
             };
         }
 
+        private void ClearProgressObject(JToken requestId)
+        {
+            long reqId = requestId.Value<long>();
```
AArnott: please add `Requires.NotNull(requestId, nameof(requestId));` here, since you dereference it.

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: I see we're using it to send notifications, of course. :) Definitely need to keep it.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+    {
+        /// <summary>
+        /// Dictionary used to map progress id token to its corresponding <see cref="ProgressParamInformation" /> instance containing the progress object and the necessary fields to report the results.
+        /// </summary>
+        public readonly Dictionary<long, ProgressParamInformation> ProgressMap = new Dictionary<long, ProgressParamInformation>();
+
+        /// <summary>
+        /// Special method name for progress notification.
+        /// </summary>
+        internal const string ProgressRequestSpecialMethod = "$/progress";
```
AArnott: >internal [](start = 8, length = 8)

Why are all these members internal? Given you use this helper class from the test assembly, I'm guessing these needn't be accessed from outside this class. Can we make them all private? 
Or if some of them must be used from outside the class, should they not be made public so that formatters implemented external to this library can use them?

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// <summary>
+    /// Class containing useful methods to help message formatters implement support for <see cref="IProgress{T}"/>.
+    /// </summary>
+    public class MessageFormatterProgressTracker
+    {
+        /// <summary>
+        /// Dictionary used to map progress id token to its corresponding <see cref="ProgressParamInformation" /> instance containing the progress object and the necessary fields to report the results.
+        /// </summary>
+        public readonly Dictionary<long, ProgressParamInformation> ProgressMap = new Dictionary<long, ProgressParamInformation>();
```
AArnott: >public [](start = 8, length = 6)

Given I don't think we want callers to *clear* our dictionary, but really just add and remove (and we already have a method on this class that adds) can we make this private and add a method to this class to remove as well, if callers need to do that?

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -445,6 +524,22 @@ private JsonRpcError ReadError(JToken json)
...
+            {
+                long progressId;
+
+                if (this.formatterProgressTracker.RequestProgressMap.TryGetValue(reqId, out progressId))
+                {
+                    this.formatterProgressTracker.RequestProgressMap.Remove(reqId);
+                    this.formatterProgressTracker.ProgressMap.Remove(progressId);
+                }
+            }
+        }
```
AArnott: This is a lot of accessing of fields directly on a helper class. The fact that we're consuming the other class's lock field directly is a pretty good indication that that class isn't protecting itself enough. Can we move this into the helper class itself?
Also, I don't see similar code in the `MessagePackFormatter`. Was it left out by accident? Once we promote this code into the helper class, let's be sure to call it from the MessagePackFormatter class as well.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// </summary>
+        /// <param name="value">The object which should implement <see cref="IProgress{T}"/> to create the <see cref="ProgressParamInformation"/> instance.</param>
+        /// <returns>The assigned <see cref="long"/> progres ID.</returns>
+        public long AddProgressObjectToMap(object value)
+        {
+            Requires.NotNull(value, nameof(value));
+
+            if (this.RequestIdBeingSerialized == null)
+            {
+                throw new NotSupportedException("IProgress<T> objects should not be part of any response or notification.");
```
AArnott: >"IProgress<T> objects should not be part of any response or notification." [](start = 48, length = 74)

This should be moved to a localized string along with the others you've already done.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            Requires.NotNull(objectType, nameof(objectType));
+
+            if (objectType.IsConstructedGenericType && objectType.GetGenericTypeDefinition().Equals(typeof(IProgress<>)))
+            {
+                return objectType;
+            }
+
+            return objectType.GetTypeInfo().GetInterfaces().FirstOrDefault(i => i.IsConstructedGenericType && i.GetGenericTypeDefinition() == typeof(IProgress<>));
+        }
```
AArnott: this is a relatively expensive set of code to run for every single value being deserialized (at least for the JSON formatter, the way the JsonConverter is used). I think we should probably build up a static `Dictionary<Type, Type>` and record/recall every calculation we do here (with thread-safety in mind). That should help prevent perf regressions from this change, which we want to avoid since it will impact all messages, even if they don't use `Progress<T>`.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+            /// <summary>
+            /// Gets the <see cref="MethodInfo"/> of <see cref="IProgress{T}.Report(T)"/>.
+            /// </summary>
+            public MethodInfo ReportMethod { get; }
+
+            /// <summary>
+            /// Gets the instance of the object implementing <see cref="IProgress{T}"/>.
+            /// </summary>
+            public object ProgressObject { get; }
+        }
```
AArnott: Instead of exposing these two properties, can we make them private and offer the formatter just one method that will invoke the method on the caller's behalf?
That way, in the future if we optimize the code (e.g. reuse the one-element array that reflection requires) it will benefit all the formatters.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        /// <summary>
+        /// Class that implements <see cref="IProgress{T}"/> and sends <see cref="ProgressRequestSpecialMethod"/> notification when reporting.
+        /// </summary>
+        public class JsonProgress<T> : IProgress<T>
```
AArnott: >public [](start = 8, length = 6)

Can we make this class private, and add two helper methods to the helper class itself:
```cs
IProgress<T> CreateProgress<T>(object token)
object CreateProgress(Type valueType);
```

The MessagePack formatter can use the generic method while the JSON formatter would use the non-generic overload. The 1-2 lines in the callers that create this (at least in the JSON formatter) are rather complicated, so offering a method to do it for the formatters within the helper class is goodness.

Note the lack of a `JsonRpc` parameter. Since this class shares in the interest of that JsonRpc instance remaining constant, can we take that as a ctor parameter or property on the helper class?

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            private readonly JsonRpc rpc;
+            private readonly long? token;
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="JsonProgress{T}"/> class.
+            /// </summary>
+            /// <param name="rpc">The <see cref="JsonRpc"/> instance used to send the <see cref="ProgressRequestSpecialMethod"/> norification.</param>
+            /// <param name="token">The <see cref="long"/> progress token used to obtain the <see cref="ProgressParamInformation"/> intance from <see cref="ProgressMap"/>.</param>
+            public JsonProgress(JsonRpc rpc, long? token)
```
AArnott: >long? [](start = 45, length = 5)

We can't assume that the client is using numbers for their progress tokens. This can hopefully be simply `object` like in the JsonRpcRequest class. The formatter may provide a `JToken` or `long`, or whatever the formatter will be able to re-serialize when transmitting the message.
We need to add at least a couple `JsonRpcServerInteropTests` to the mix to exercise both number and non-number progress tokens to make sure we've shaken out all the `long` assumptions from the code.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+        public class JsonProgress<T> : IProgress<T>
+        {
+            private readonly JsonRpc rpc;
+            private readonly long? token;
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="JsonProgress{T}"/> class.
+            /// </summary>
+            /// <param name="rpc">The <see cref="JsonRpc"/> instance used to send the <see cref="ProgressRequestSpecialMethod"/> norification.</param>
+            /// <param name="token">The <see cref="long"/> progress token used to obtain the <see cref="ProgressParamInformation"/> intance from <see cref="ProgressMap"/>.</param>
```
AArnott: >intance [](start = 132, length = 7)

typo: intance -> instance

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+        /// </summary>
+        public class JsonProgress<T> : IProgress<T>
+        {
+            private readonly JsonRpc rpc;
+            private readonly long? token;
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="JsonProgress{T}"/> class.
+            /// </summary>
+            /// <param name="rpc">The <see cref="JsonRpc"/> instance used to send the <see cref="ProgressRequestSpecialMethod"/> norification.</param>
```
AArnott: >norification [](start = 129, length = 12)

typo: norification -> notification

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+            public JsonProgress(JsonRpc rpc, long? token)
+            {
+                this.rpc = rpc ?? throw new ArgumentNullException(nameof(rpc));
+                this.token = token ?? throw new ArgumentNullException(nameof(token));
+            }
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="JsonProgress{T}"/> class.
+            /// </summary>
+            /// <param name="rpc">The <see cref="JsonRpc"/> instance used to send the <see cref="ProgressRequestSpecialMethod"/> norification.</param>
```
AArnott: >norification [](start = 129, length = 12)

typo: norification -> notification

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+            /// <param name="token">The <see cref="JToken"/> progress token used to obtain the <see cref="ProgressParamInformation"/> intance from <see cref="ProgressMap"/>.</param>
```
AArnott: >intance [](start = 134, length = 7)

typo: intance -> instanceCan we

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,168 @@
...
+            public JsonProgress(JsonRpc rpc, JToken token)
```
AArnott: >JToken [](start = 45, length = 6)

We should avoid `JToken` or any other newtonsoft.json type in shared infrastructure. Can you remove it?

src/StreamJsonRpc.Tests/MessagePackFormatter.cs
```diff
@@ -119,5 +146,71 @@ public void Serialize(IBufferWriter<byte> contentBuffer, JsonRpcMessage message)
...
+            }
+
+            public TIProgressOfT Deserialize(ref MessagePackReader reader, MessagePackSerializerOptions options)
+            {
+                long token = reader.ReadInt64();
+
+                if (this.formatter.formatterProgressTracker.ProgressMap.TryGetValue(token, out MessageFormatterProgressTracker.ProgressParamInformation progressInfo))
+                {
+                    Type progressType = typeof(MessageFormatterProgressTracker.JsonProgress<>).MakeGenericType(typeof(T));
+                    IProgress<T> p = new MessageFormatterProgressTracker.JsonProgress<T>(this.formatter.Rpc, token);
```
AArnott: I don't see you using this local that you've defined. Can we remove this line?

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
milopezc: I thought the client would only provide the Progress object and then we assign a token to it and replace it in the message, storing the Progress object in the dictionary. I made the change for object anyway since I think it won't affect any functionality.

---
In reply to: [314177878](https://github.com/microsoft/vs-streamjsonrpc/pull/320#discussion_r314177878) [](ancestors = 314177878)

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
milopezc: Fixed, made them public or private according to how we use them.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
milopezc: Fixed in latest update.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
milopezc: Moved in latest update.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+        public static Type FindIProgressOfT(Type objectType)
+        {
+            Requires.NotNull(objectType, nameof(objectType));
+
+            if (objectType.IsConstructedGenericType && objectType.GetGenericTypeDefinition().Equals(typeof(IProgress<>)))
+            {
+                return objectType;
+            }
+
+            Type iProgressOfTType = null;
```
AArnott: > = null [](start = 33, length = 7)

nit: you can drop the local initializer. You initialize it unconditionally later so C# won't complain.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
AArnott: Yes, on the *client* side you're right. But this `JsonProgress<T>` class isn't used on the client at all. It's used on the server-side. And on the server side, we can't control what the token will be since the client determines the token.

---
In reply to: [314497199](https://github.com/microsoft/vs-streamjsonrpc/pull/320#discussion_r314497199) [](ancestors = 314497199,314177878)

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -135,11 +165,19 @@ public Version ProtocolVersion
         /// <summary>
         /// Gets the <see cref="Newtonsoft.Json.JsonSerializer"/> used when serializing and deserializing method arguments and return values.
         /// </summary>
-        public JsonSerializer JsonSerializer { get; } = new JsonSerializer()
+        public JsonSerializer JsonSerializer { get; }
+
+        /// <inheritdoc/>
+        public JsonRpc Rpc
         {
```
AArnott: Make this an explicit interface implementation so that folks are much less likely to try to set this property. You can do this by changing this line to: `JsonRpc IJsonRpcInstanceContainer.Rpc`

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -445,6 +526,13 @@ private JsonRpcError ReadError(JToken json)
...
         }
 
+        private void ClearProgressObject(JToken requestId)
+        {
+            Requires.NotNull(requestId, nameof(requestId));
+
+            this.formatterProgressTracker.ClearProgressMaps(requestId.Value<long>());
+        }
+
         [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
```
AArnott: At this point, the value of this method is rather low. I suggest you delete it and use the one line that is worth something to replace the one line used by its two callers to call this method.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+
+            return iProgressOfTType;
+        }
+
+        /// <summary>
+        /// Saves a <see cref="ProgressParamInformation"/> instance obtained from the given <see cref="object"/> into <see cref="progressMap"/>.
+        /// </summary>
+        /// <param name="value">The object which should implement <see cref="IProgress{T}"/> to create the <see cref="ProgressParamInformation"/> instance.</param>
+        /// <returns>The assigned <see cref="long"/> progres ID.</returns>
+        public long AddProgressObjectToMap(object value)
```
AArnott: >AddProgressObjectToMap [](start = 20, length = 22)

This method name describes more of the internals of what it does than what the outside caller should have to think about. Can you name this based on the use case for the caller? For example: `GetTokenForProgress`

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+
+                return progressId;
+            }
+        }
+
+        /// <summary>
+        /// Clears the objects in <see cref="requestProgressMap"/> and <see cref="progressMap"/> associated to the given request ID.
+        /// </summary>
+        /// <param name="requestId">The id of the request whose associated object need to be cleared.</param>
+        public void ClearProgressMaps(long requestId)
```
AArnott: >ClearProgressMaps [](start = 20, length = 17)

This method name describes more of the internals of what it does than what the outside caller should have to think about. Can you name this based on the use case for the caller? For example: `OnResponseReceived`
Then modify the xml doc comments similar to my advice in the above comment.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Gets the <see cref="ProgressParamInformation"/> object associated with the given progress id.
+        /// </summary>
+        /// <param name="progressId">The key to obtain the <see cref="ProgressParamInformation"/> object from <see cref="progressMap"/>.</param>
+        /// <param name="valueType">Output parameter to store the obtained <see cref="ProgressParamInformation"/> object.</param>
+        public bool GetProgressTypeToReport(object progressId, out ProgressParamInformation valueType)
```
AArnott: >GetProgressTypeToReport [](start = 20, length = 23)

This returns more than the progress type. The naming convention for methods that return bool and have an out parameter is `TryGet...`. So how about naming this `TryGetProgressObject`?

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+                    return true;
+                }
+
+                valueType = null;
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="JsonProgress{T}"/> class.
```
AArnott: >JsonProgress [](start = 53, length = 12)

Scrub mention of the `JsonProgress<T>` class from xml doc comments on your public members. This class is no longer visible to them, and it's not info they need to know. What they need to know is that it is an `IProgress<T>` instance for use on the receiving end of an RPC call.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+            /// <summary>
+            /// Gets the <see cref="MethodInfo"/> of <see cref="IProgress{T}.Report(T)"/>.
+            /// </summary>
+            private MethodInfo ReportMethod { get; }
+
+            /// <summary>
+            /// Gets the instance of the object implementing <see cref="IProgress{T}"/>.
+            /// </summary>
+            private object ProgressObject { get; }
+
```
AArnott: nit: convert private properties to private readonly fields

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -0,0 +1,239 @@
...
+        /// Gets or sets the the next id value to assign as token for the progress objects.
+        /// </summary>
+        private long NextProgressId { get; set; }
+
+        /// <summary>
+        /// Converts given <see cref="Type"/> to its <see cref="IProgress{T}"/> type.
+        /// </summary>
+        /// <param name="objectType">The type which may implement <see cref="IProgress{T}"/>.</param>
+        /// <returns>The <see cref="IProgress{T}"/> from given <see cref="Type"/> object, or <c>null</c>  if no such interface was found in the given <paramref name="objectType" />.</returns>
+        public static Type FindIProgressOfT(Type objectType)
```
AArnott: >FindIProgressOfT [](start = 27, length = 16)

The only caller of this in a formatter appears to use it to test whether the result is non-null, and they want to know whether `AddProgressObjectToMap` will accept the object later. So how about making this method private (to support your own uses), and adding a new public method called `IsSupportedProgressType` that returns `bool` so that callers know better when to call it and have less code to write?

src/StreamJsonRpc.Tests/MessagePackFormatter.cs
```diff
@@ -47,13 +62,28 @@ public MessagePackFormatter(bool compress)
...
+                Verify.Operation(this.rpc == null, "This formatter already belongs to another JsonRpc instance. Create a new instance of this formatter for each new JsonRpc instance.");
+
+                this.rpc = value;
+            }
         }
 
+        /// <summary>
+        /// Gets <see cref="MessagePackSerializerOptions.LZ4Default"/> if <see cref="compress"/> is true, otherwise <c>null</c>.
+        /// </summary>
+        private MessagePackSerializerOptions CompressOptions
```
AArnott: >CompressOptions [](start = 45, length = 15)

nit: rename to `SerializerOptions` since there's more to it than just whether it's compressed. Right *now* that's all the thought we put into it, but to the user of this property, they just use it as the serializer options.

src/StreamJsonRpc.Tests/MessagePackFormatter.cs
```diff
@@ -47,13 +62,28 @@ public MessagePackFormatter(bool compress)
...
+        {
+            get => this.compress ? MessagePackSerializerOptions.LZ4Default : null;
```
AArnott: >null [](start = 77, length = 4)

We must never return `null` here, since that means that the `MessagePack` library's (mutable) default will apply, but we must never trust a mutable default. We must be explicit. So this should be `MesasgePackSerializerOptions.Default` instead of `null`.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
AArnott: Leaving this active till we get server interop tests to validate we work properly with non-long tokens.

---
In reply to: [314849437](https://github.com/microsoft/vs-streamjsonrpc/pull/320#discussion_r314849437) [](ancestors = 314849437,314497199,314177878)

AArnott:
Also, can you please add a markdown file under the doc/ folder (and link to it from an existing page) that explains how `IProgress<T>` support works on the client and server side, and either links to the `IProgress<T>` issue that documents the protocol or includes a copy of it?

PR #322, opened 2019-08-15
Added ETW events

Shyam-Gupta:
A sample PerfView log:

image

Call stacks getting captured with the events:

image

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -1108,12 +1108,25 @@ namespace StreamJsonRpc
                 {
                     if (!request.IsResponseExpected)
                     {
+                        if (JsonRpcEventSource.Instance.IsEnabled())
+                        {
+                            JsonRpcEventSource.Instance.SendNotificationStart($"Request Method: \"{request.Method}\", Arguments: \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
+                        }
+
                         await this.TransmitAsync(request, cts.Token).ConfigureAwait(false);
+
+                        if (JsonRpcEventSource.Instance.IsEnabled())
+                        {
+                            JsonRpcEventSource.Instance.SendNotificationStop($"Request Method: \"{request.Method}\", Arguments: \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
+                        }
+
                         return default;
                     }
 
                     Verify.Operation(this.readLinesTask != null, Resources.InvalidBeforeListenHasStarted);
                     var tcs = new TaskCompletionSource<TResult>();
+                    string responseDetails = string.Empty;
+
                     Action<JsonRpcMessage> dispatcher = (response) =>
                     {
                         lock (this.dispatcherMapLock)
@@ -1123,6 +1136,8 @@ namespace StreamJsonRpc
 
                         try
                         {
+                            responseDetails = string.Empty;
+
                             if (response == null)
                             {
                                 if (this.TraceSource.Switch.ShouldTrace(TraceEventType.Warning))
@@ -1139,6 +1154,11 @@ namespace StreamJsonRpc
                                 {
                                     tcs.TrySetException(new ConnectionLostException());
                                 }
+
+                                if (JsonRpcEventSource.Instance.IsEnabled())
+                                {
+                                    responseDetails = $"Aborting pending request \"{id}\" because the connection was lost.";
+                                }
                             }
                             else if (response is JsonRpcError error)
                             {
@@ -1150,10 +1170,25 @@ namespace StreamJsonRpc
                                 {
                                     tcs.TrySetException(CreateExceptionFromRpcError(error, targetName));
                                 }
+
+                                if (JsonRpcEventSource.Instance.IsEnabled())
+                                {
+                                    responseDetails = JsonRpcEventSource.GetErrorDetails(error);
+                                }
                             }
                             else if (response is JsonRpcResult result)
                             {
                                 tcs.TrySetResult(result.GetResult<TResult>());
+
+                                if (JsonRpcEventSource.Instance.IsEnabled())
+                                {
+                                    responseDetails = $"Success!";
+                                }
+                            }
+
+                            if (JsonRpcEventSource.Instance.IsEnabled())
+                            {
+                                JsonRpcEventSource.Instance.SendRequestStop($"ResponseDetails: {responseDetails}; RequestDetails: Id: \"{request.Id}\", Method: \"{request.Method}\"");
                             }
                         }
                         catch (Exception ex)
@@ -1170,6 +1205,11 @@ namespace StreamJsonRpc
 
                     try
                     {
+                        if (JsonRpcEventSource.Instance.IsEnabled())
+                        {
+                            JsonRpcEventSource.Instance.SendRequestStart($"Request Id: \"{request.Id}\", Method: \"{request.Method}\", Arguments: \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
+                        }
+
                         await this.TransmitAsync(request, cts.Token).ConfigureAwait(false);
                     }
                     catch
@@ -1507,6 +1547,18 @@ namespace StreamJsonRpc
                     this.TraceSource.TraceEvent(TraceEventType.Information, (int)TraceEvents.LocalInvocation, "Invoking {0}", targetMethod);
                 }
 
+                if (JsonRpcEventSource.Instance.IsEnabled())
+                {
+                    if (request.IsResponseExpected)
+                    {
+                        JsonRpcEventSource.Instance.InvokeMethodStart($"TargetMethod = {targetMethod}; RequestDetails: Id = \"{request.Id}\", Method = \"{request.Method}\", Arguments = \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
+                    }
+                    else
+                    {
+                        JsonRpcEventSource.Instance.InvokeNotificationStart($"TargetMethod = {targetMethod}; RequestDetails: Method = \"{request.Method}\", Arguments = \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
+                    }
+                }
+
                 // Yield now so method invocation is async and we can proceed to handle other requests meanwhile.
                 // IMPORTANT: This should be the first await in this async method,
                 //            and no other await should be between this one and actually invoking the target method.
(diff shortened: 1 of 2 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -507,10 +519,23 @@ public static T Attach<T>(Stream stream)
         public static T Attach<T>(Stream sendingStream, Stream receivingStream)
             where T : class
         {
+            string typeName = typeof(T).FullName;
+
+            if (JsonRpcEventSource.Instance.IsEnabled())
+            {
+                JsonRpcEventSource.Instance.GenerateProxyUsingStreamsAndStartListeningStart(typeName);
```
AArnott: I'm not a fan of all these events that basically just serve as bounds for a method call. CPU stack events already give this idea. Why do you need extra start and stop events for these methods?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1130,6 +1190,11 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
                                     this.TraceSource.TraceEvent(TraceEventType.Warning, (int)TraceEvents.RequestAbandonedByRemote, "Aborting pending request \"{0}\" because the connection was lost.", id);
                                 }
 
+                                if (JsonRpcEventSource.Instance.IsEnabled())
+                                {
+                                    JsonRpcEventSource.Instance.LogWarning($"Aborting pending request \"{id}\" because the connection was lost.");
```
AArnott: It looks like you're duplicating several trace messages, but not all of them.
How did you pick which ones to trace to both places? 
Why is it valuable to use ETW to trace this when we already have a TraceSource?
If it's valuable, couldn't someone just add a trace listener and log ETW from there to avoid the duplication here?

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// The event ID for the <see cref="LocalMethodAdded(string)"/>.
+        /// </summary>
+        private const int LocalMethodAddedEvent = 27;
+
+        /// <summary>
+        /// Creates a string representation of arguments for logging.
+        /// </summary>
+        /// <param name="arguments">A single argument or an array of arguments.</param>
+        /// <returns>String representation of arguments.</returns>
+        public static string GetArgumentsString(object arguments)
```
AArnott: nit: avoid `public` for what are really internal members (since the containing type is internal).

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -0,0 +1,312 @@
...
+                    {
+                        stringBuilder.AppendLine($"arg{i}: {args[i].ToString()}; ");
+                    }
+                }
+
+                return stringBuilder.ToString();
+            }
+            else
+            {
+                return arguments.ToString();
```
AArnott: this will throw `NullReferenceException` if `arguments == null`.

src/StreamJsonRpc/JsonRpc.cs
Shyam-Gupta: Removed these events as per our discussion yesterday.

src/StreamJsonRpc/JsonRpc.cs
Shyam-Gupta: Removed them as per yesterday's discussion.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Done

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Fixed.

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -0,0 +1,232 @@
...
+            const int maxLength = 1024;
+
+            if (arguments is object[] args)
+            {
+                var stringBuilder = new StringBuilder();
+                if (args != null && args.Length > 0)
+                {
+                    for (int i = 0; i < args.Length; ++i)
+                    {
+                        stringBuilder.Append($"arg{i}: {args[i].ToString()}, ");
```
AArnott: You can drop the `.ToString()`. Formatting arguments *always* get that method called on them implicitly anyway.
Also, I wonder if rendering arguments like this is a good idea. Maybe objects don't override ToString(), and those that do may use quotes, colons, and commas anyway leading to a very confusing output here. 
As this is something of a tough problem and can lead to large output, do we *need* the include arguments? They wouldn't be needed as part of a perf investigation I think, and they're already available in the `TraceSource` log. Can we just remove ETW tracing of arguments to keep things simple for this first iteration?

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1150,10 +1178,25 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
...
+
+                                if (JsonRpcEventSource.Instance.IsEnabled())
+                                {
+                                    responseDetails = $"Success!, Result.Id = \"{result.Id}\"";
+                                }
+                            }
+
+                            if (JsonRpcEventSource.Instance.IsEnabled())
+                            {
+                                JsonRpcEventSource.Instance.SendRequestStop($"ResponseDetails: {responseDetails}; RequestDetails: Id: \"{request.Id}\", Method: \"{request.Method}\", Arguments: \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
```
Shyam-Gupta: For some reason, Duration_Msec is not getting captured for most of the SendRequestStop events. For such events, no ActivityId shows up in PerfView. 
In case of WinForms designer, we spawn a .NetCore process from devenv. Not sure if framework mismatch is the issue.

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -0,0 +1,232 @@
...
+                        if (stringBuilder.Length > maxLength)
+                        {
+                            return $"{stringBuilder.ToString().Substring(0, maxLength)}...(truncated)";
```
AArnott: Your `.ToString().Substring()` pattern would end up creating the very large string only to chop it, and all after you've already constructed a long stringbuilder.
You can optimize this by replacing `.ToString().Substring()` with just `.ToString(0, maxLength)`.

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -0,0 +1,232 @@
...
+                return stringBuilder.ToString();
+            }
+            else
+            {
+                string retVal = arguments.ToString();
+                return (retVal.Length > maxLength) ? $"{retVal.Substring(0, maxLength)}...(truncated)" : retVal;
+            }
+        }
+
+        internal static string GetErrorDetails(JsonRpcError error)
```
AArnott: similar: unless we have a need for error details, I'd rather they not be included in ETW tracing.
Ideally the ETW traces should only include a fixed amount of data, or known to be very small (e.g. request id and method name is fine for requests, and id for all responses + error code for errors).

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1108,12 +1116,25 @@ protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetNa
                 {
                     if (!request.IsResponseExpected)
                     {
+                        if (JsonRpcEventSource.Instance.IsEnabled())
+                        {
+                            JsonRpcEventSource.Instance.SendNotificationStart($"Request Method: \"{request.Method}\", Arguments: \"{JsonRpcEventSource.GetArgumentsString(request.Arguments)}\"");
```
AArnott: I'd like to drop arguments from these events. Then we can pass just the `request.Method` string itself as the argument and the event can be raised with `method` as a proper column in the ETW event that can be easily filtered for.

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            this.WriteEvent(CancelRequestStopEvent, message);
+        }
+
+        /// <summary>
+        /// Invoke Notification.
+        /// </summary>
+        /// <param name="message">Details of the Notification.</param>
+        [Event(InvokeNotificationStartEvent, Level = EventLevel.Verbose)]
+        public void InvokeNotificationStart(string message)
```
AArnott: start and stop events should be properly attributed as such so that PerfView recognizes the pairing. See https://github.com/microsoft/vs-threading/blob/90fc3eaff4add936f1a990de26c541a681605253/src/Microsoft.VisualStudio.Threading/ThreadingEventSource.cs#L86 for an example.

src/StreamJsonRpc/JsonRpc.cs
AArnott: That's probably because of missing attribute properties. See https://github.com/microsoft/vs-streamjsonrpc/pull/322#discussion_r314906719

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -507,10 +507,13 @@ public static T Attach<T>(Stream stream)
         public static T Attach<T>(Stream sendingStream, Stream receivingStream)
             where T : class
         {
+            string typeName = typeof(T).FullName;
```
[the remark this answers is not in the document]
Shyam-Gupta: Done.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Without arguments sometimes its difficult to know which method will get called. 
For instance, in WinForms, Method is mostly: "SendRequest". Its the argument that tells which endpoint will get hit. E.g: 'control/GetVisible', 'object/GetPropertyValue' etc.

AArnott: Hmm... ok fair point. I'm curious why you factored it that way. You can't have more precise method names? I guess since you've already identified that it's important to you, that's fine.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Thanks. Will do.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Ok. I will remove error message. It will only have Id and error code.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: That is how the WinForms designer implementation is. I guess it was done to have a single end point on the client side and club related operations for each object type.

I can limit the length of arguments to say 200 characters instead of 1024.

src/StreamJsonRpc/JsonRpc.cs
Shyam-Gupta: Required to identify the endpoint.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Thanks. Did that.

src/StreamJsonRpc/JsonRpc.cs
Shyam-Gupta: Tried but it didn't help.

src/StreamJsonRpc/JsonRpcEventSource.cs
Shyam-Gupta: Truncated arguments to 128 chars. 
Also removed arguments from InvokeMethodStop and SendRequestStop events as they can be correlated with their Start counterparts using request Id.

AArnott:
I took the liberty of squashing and rebasing your commits from master to v2.1 so that we can check this in for 16.3.

AArnott:
I've revised the ETW events somewhat to provide more data-friendly events. Here's how the perfview grid looks now. Note the columns dedicated for methods, args. We provide this string data only on the start event so the stop event can be leaner.

image

Shyam-Gupta:
Thanks Andrew for making the changes. Looks very good now.

PR #334, opened 2019-09-07
Add support for marshaling an IDuplexPipe or Stream as a JSON-RPC argument

### LINKED ISSUE #140: Support for out of band streams referenced in JSON-RPC messages
## Proposal

When a server method signature includes a `Stream`, it signifies that a side-channel should be established between the two parties. This is useful for efficient transfer of large binary blobs or any other one- or two-way communication that should take place outside the standard JSON-RPC protocol.

The method and transport used to create the side-channel is outside the scope of this spec, but may be established on the same underlying transport by leveraging a multiplexing stream (such as [the one implemented in Nerdbank.Streams](https://github.com/AArnott/Nerdbank.Streams/blob/master/doc/MultiplexingStream.md)) or as an entirely new connection.

For example, given a server method signature such as this:

```cs
async Task SaveJpgContent(string fileName, Stream stream);
```

A client may invoke it like this:

```cs
await jsonRpcClient.InvokeAsync("SaveJpgContent", fileName,  File.OpenRead(fileName));
```

The client library then takes the Stream that is passed to it, creates a side-channel, and sends the side-channel id to the server:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "TakeThisOfflineAsync",
  "params": [
    "some.jpg",
    15 // supposing the side-channel ID is 15
  ]
}
```

The server library would receive this request, find the matching method and notice that the second parameter is a Stream, and interpret the `15` in the request as a side-channel id and accept the connection to that channel. It would then provide the stream that represents that channel as the second argument when it actually invokes the server method locally.

At this point, the server method can read from the stream.

## Additional notes

For two-way communication, the `Stream` a client passes into the library would have to be prepared for the library to potentially read from *and* write to in order to propagate information from the side-channel. But this leaves the question of how the client itself communicates since both Read and Write methods must be used by the library. This might be resolved by passing in one of the Streams created by [`FullDuplexStream`](https://github.com/AArnott/Nerdbank.Streams/blob/master/doc/FullDuplexStream.md), and keeping the other Stream in the pair for communication.

## Open questions

1. Must the server complete using the stream before sending the JSON-RPC response message? I'm thinking not since I can't think of a reason why that would be a necessary restriction. But servers may certainly choose to complete all side-channel communication before returning from their async methods and sending the JSON-RPC response to the client.
1. Some libraries are not in strongly typed languages. How would a javascript implementation know to convert the `15` to a `Stream`? Perhaps this whole automatic feature can be provided only in a strongly-typed library, while maintaining wire compatibility with loosely-typed languages since they can still accept the side-channel by expecting `15` to be interpreted as its id.
1. How might we work the new System.IO.Pipelines API into these server methods?

AArnott:
Closes #140

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Buffers;
 using System.Text;
+using System.Threading.Tasks;
 using Nerdbank.Streams;
 using Newtonsoft.Json;
 using StreamJsonRpc;
@@ -110,6 +111,31 @@ public class JsonMessageFormatterTests : TestBase
         Assert.True(jsonMessage.TryGetArgumentByNameOrIndex(null, 0, typeof(string), out object value));
         Assert.IsType<string>(value);
         Assert.Equal("2019-01-29T03:37:28.4433841Z", value);
+    }
+
+    /// <summary>
+    /// Verifies that the <see cref="JsonMessageFormatter.MultiplexingStream"/> property
+    /// retains values it is set to, and accepts null.
+    /// </summary>
+    /// <remarks>
+    /// The rest of the multiplexing stream tests are defined in <see cref="DuplexPipeMarshalingTests"/>.
+    /// </remarks>
+    [Fact]
+    public async Task MultiplexingStream()
+    {
+        var formatter = new JsonMessageFormatter();
+        Assert.Null(formatter.MultiplexingStream);
+
+        Tuple<Nerdbank.FullDuplexStream, Nerdbank.FullDuplexStream> streams = Nerdbank.FullDuplexStream.CreateStreams();
+        MultiplexingStream[] mxStreams = await Task.WhenAll(
+            Nerdbank.Streams.MultiplexingStream.CreateAsync(streams.Item1, this.TimeoutToken),
+            Nerdbank.Streams.MultiplexingStream.CreateAsync(streams.Item2, this.TimeoutToken));
+
+        formatter.MultiplexingStream = mxStreams[0];
+        Assert.Same(mxStreams[0], formatter.MultiplexingStream);
+
+        formatter.MultiplexingStream = null;
+        Assert.Null(formatter.MultiplexingStream);
     }
 
     private static long MeasureLength(JsonRpcRequest msg, JsonMessageFormatter formatter)
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -30,7 +30,7 @@ namespace StreamJsonRpc
     /// <remarks>
     /// Each instance of this class may only be used with a single <see cref="JsonRpc" /> instance.
     /// </remarks>
-    public class JsonMessageFormatter : IJsonRpcAsyncMessageTextFormatter, IJsonRpcInstanceContainer
+    public class JsonMessageFormatter : IJsonRpcAsyncMessageTextFormatter, IJsonRpcInstanceContainer, IDisposable
     {
         /// <summary>
         /// The key into an <see cref="Exception.Data"/> dictionary whose value may be a <see cref="JToken"/> that failed deserialization.
@@ -76,6 +76,11 @@ namespace StreamJsonRpc
         /// <see cref="MessageFormatterProgressTracker"/> instance containing useful methods to help on the implementation of message formatters.
         /// </summary>
         private readonly MessageFormatterProgressTracker formatterProgressTracker = new MessageFormatterProgressTracker();
+
+        /// <summary>
+        /// The helper for marshaling pipes as RPC method arguments.
+        /// </summary>
+        private readonly MessageFormatterDuplexPipeTracker duplexPipeTracker = new MessageFormatterDuplexPipeTracker();
 
         /// <summary>
         /// The version of the JSON-RPC protocol being emulated by this instance.
@@ -124,6 +129,8 @@ namespace StreamJsonRpc
                 {
                     new JsonProgressServerConverter(this),
                     new JsonProgressClientConverter(this),
+                    new DuplexPipeConverter(this),
+                    new StreamConverter(this),
                 },
             };
         }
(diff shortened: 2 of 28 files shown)


src/StreamJsonRpc/Reflection/MessageFormatterDuplexPipeTracker.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            MultiplexingStream mxstream = this.GetMultiplexingStreamOrThrow();
+            if (token is null)
+            {
+                return null;
+            }
+
+            // In the case of multiple overloads, we might be called to convert a channel's token more than once.
+            // But we can only accept the channel once, so look up in a dictionary to see if we've already done this.
+            if (!this.openInboundChannels.TryGetValue(token.Value, out MultiplexingStream.Channel channel))
```
AArnott: In case the request gets rejected with no match, we need to be sure we close this channel rather than leave it open in our dictionary

AArnott: Fixed

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -677,5 +723,49 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+            private readonly JsonMessageFormatter jsonMessageFormatter;
+
+            public DuplexPipeConverter(JsonMessageFormatter jsonMessageFormatter)
+            {
+                this.jsonMessageFormatter = jsonMessageFormatter ?? throw new ArgumentNullException(nameof(jsonMessageFormatter));
+            }
+
+            public override IDuplexPipe ReadJson(JsonReader reader, Type objectType, IDuplexPipe existingValue, bool hasExistingValue, JsonSerializer serializer)
+            {
+                int? tokenId = JToken.Load(reader).Value<int?>();
```
IlyaBiryukov: Would it be better to serialize as an object, e.g. `{"id": id}` for future extensibility?

AArnott: I considered that, but I couldn't think of any extensibility we would need. This depends on MultiplexingStream, which only needs an integer to accept a channel. And if we ever switched away from MultiplexingStream, that would be a breaking change on both sides so we would have to define a new protocol for whatever would replace it anyway.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -677,5 +723,49 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
...
+
+            public override Stream ReadJson(JsonReader reader, Type objectType, Stream existingValue, bool hasExistingValue, JsonSerializer serializer)
+            {
+                int? tokenId = JToken.Load(reader).Value<int?>();
+                return this.jsonMessageFormatter.duplexPipeTracker.GetPipe(tokenId)?.AsStream();
+            }
+
+            public override void WriteJson(JsonWriter writer, Stream value, JsonSerializer serializer)
+            {
+                var token = this.jsonMessageFormatter.duplexPipeTracker.GetToken(value?.UsePipe());
```
IlyaBiryukov: Using a stream as a pipe loses its random access capability, so we can either not allow such streams to be passed through, or document this behavior.

How about one-directional streams, that can only read or write? Looks like write-only streams are also not supported, because `UsePipeReader` in `PipeExtensions.cs` in `Nerdbank.Streams` would throw.

src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
```diff
@@ -112,6 +113,24 @@ public void JTokenParserHonorsSettingsOnSerializer()
         Assert.Equal("2019-01-29T03:37:28.4433841Z", value);
     }
 
+    [Fact]
+    public async Task MultiplexingStream()
```
milopezc: Just out of curiosity what are we testing with this? The Assert methods seem to me to be only testing the setting of formatter.MultiplexingStream.

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: > Using a stream as a pipe loses its random access capability

What do you mean? `Stream.UsePipe()` returns an `IDuplexPipe` which can be read from and written to just like a duplex `Stream`. The only lossy conversion is from `IDuplexPipe` to `Stream`, since only `IDuplexPipe` lets you 'complete' one direction of communication and a `Stream` doesn't offer such an API.
And I *do* document that in the .md file I add in this PR.

> How about one-directional streams, that can only read or write? 

I support that too. Check out the unit tests I add. I specifically test with both read-only and write-only streams *and* "duplex" pipes. It all works.
I [updated Nerdbank.Streams' `UsePipe()` extension method](https://github.com/AArnott/Nerdbank.Streams/pull/115) to support one-directional streams last week to support this.

src/StreamJsonRpc.Tests/JsonMessageFormatterTests.cs
AArnott: I should add a comment. I am simply testing the `MultiplexingStream` property on the formatter -- that it can retain a value and accepts a `null` assignment as well.

All the functionality we expect due to setting the property is tested in another class. I'll mention this in the comment too.

src/StreamJsonRpc/JsonMessageFormatter.cs
IlyaBiryukov: >> Using a stream as a pipe loses its random access capability

> What do you mean?

Suppose some client passes a stream, and then wants to set `Position` on it. Will this screw up the service? Can the service use `Position` on the stream?

>> How about one-directional streams, that can only read or write?

> I support that too.
Cool. My local copy of `Nerdbank.Streams` doesn't have it, that's why I was asking.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -28,7 +28,7 @@ public class MessageFormatterProgressTracker
         /// <summary>
         /// Dictionary used to map the outbound request id to their progress id token so that the progress objects are cleaned after getting the final response.
         /// </summary>
-        private readonly Dictionary<long, object> requestProgressMap = new Dictionary<long, object>();
+        private readonly Dictionary<long, long> requestProgressMap = new Dictionary<long, long>();
```
milopezc: If this is changed to long, shouldn't the key in progressMap be changed to long as well?

AArnott: No. `progressMap` maps *incoming* messages while `requestProgressMap` maps `outgoing` messages. So while it is correct for `progressMap` to use `object` as the key since the ID could be a number or string, *this* map can correctly assume that the message ID is a `long` since we are always the author and always use longs.
So this was just a fix-up (unrelated to the rest of my changes) that I noticed I could make to yours while I was studying and copying from your class to make my own tracker class. :)

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: > Suppose some client passes a stream, and then wants to set Position on it. 

Network streams are not seekable. `Stream.CanSeek == false`. So you can't set `Position` on it -- it would throw `NotSupportedException`. Seek only makes sense on a stream representing a persistent thing like a file. But these streams are network streams -- they represent communication channels, and thus cannot seek, but you can read and write to them concurrently, which does *not* make sense for a file.

IMO it's confusing how in .NET a Stream can represent two totally different use cases like this. I think that is somewhat alleviated by the introduction of `IDuplexPipe` which is a type dedicated to just one of these (and does a better job).

PR #344, opened 2019-09-21
Add PipeReader/PipeWriter support for RPC method parameters

AArnott:


src/StreamJsonRpc/Reflection/MessageFormatterDuplexPipeTracker.cs
```diff
@@ -164,6 +188,44 @@ public IDuplexPipe GetPipe(int? token)
...
+        /// Creates a <see cref="PipeReader"/> from a given token as it is received at the RPC server as a method argument.
+        /// </summary>
+        /// <param name="token">The method argument, which was originally obtained by the client using the <see cref="GetToken(IDuplexPipe)"/> method.</param>
+        /// <returns>The <see cref="PipeReader"/> from the token; or <c>null</c> if <paramref name="token"/> was <c>null</c>.</returns>
+        /// <exception cref="InvalidOperationException">Thrown if the token does not match up with an out of band channel offered by the client.</exception>
+        /// <exception cref="NotSupportedException">Thrown if no <see cref="MultiplexingStream"/> was provided to the constructor.</exception>
+        public PipeReader GetPipeReader(int? token)
+        {
+            IDuplexPipe duplexPipe = this.GetPipe(token);
+            if (duplexPipe != null)
```
ikeras: For both of these methods, my assumption is that if duplexPipe is != null, then both the input and the output must also be non-null (at least after you complete the output or input). Is that right?

AArnott: An interesting question. I haven't been able to find docs to confirm whether `IDuplexPipe`'s two properties are meant to never be null. So in general it would probably be good defensive programming to defend against null before I `Complete()` either of these. But I suspect that these properties should never be null, and as I know these came from `MultiplexingStream` which never lets them be null, it's OK in this case.

Issue #288, opened 2019-06-04
Support for flexible named arg matching


AArnott:
The LSP spec always uses named arguments. It assumes that adding arguments is a non-breaking change. For StreamJsonRpc though, a client adding or removing a named argument *is* a breaking change for a server because our method overload selector works by matching every single named argument to every single method parameter. 

Today, StreamJsonRpc users that implement LSP servers do so by accepting a single `JToken` parameter. It may offer a preferable experience if server methods could actually use the named parameters in the method signature, without becoming brittle to LSP spec changes. 

The proposal here is to add an *option* to `JsonRpc` that allows it to "fuzzy match" target methods based on named arguments. When no named argument specifies a value for a method parameter, the default value for that parameter type is provided. When no parameter on the target method matches the name of a named argument, that argument is ignored.

AArnott:
The good part of this proposal (compared to #289) is that it remains largely true to the JSON-RPC spec with regard to its intended treatment of named arguments. The possible negative is that server methods have to change their signature (and the author must refer to the spec to see what parameters it might accept) in order to receive additional data, so it's less discoverable than a single strongly-typed object as described in #289.

dbaeumer:
I would actually go with this proposal since it matches the original idea of having named arguments. However we could never unfold them correctly in JS/TS (no reflection) but it would be the right thing to do.

>  When no named argument specifies a value for a method parameter, the default value for that parameter type is provided.

Could these be nullable types in C# and `null` be passed in the absence instead of the default value?

AArnott:
Yes, the default value for nullable types is `null`. So if a server method took an `int` parameter it would get `0` in the absence of an argument from the client. If that server method took `int?` instead, it would get `null` in the same circumstance, and could thus discern between a `0` and no value from the client.

dbaeumer:
Wasn't aware of this.

PR #347, opened 2019-10-17
Support for deserializing params *object* as single first parameter

### LINKED ISSUE #289: Support for deserializing params *object* as single first parameter
To improve LSP servers' interop with LSP clients when implemented with StreamJsonRpc, consider an option on `JsonRpc` that allows server methods to receive all named arguments via one deserialized type specified as the server method's first/only parameter. 

This could replace the one `JToken` parameter from LSP server methods with the anticipated type per the LSP spec that is already provided by an LSP protocol library. So instead of a server method boilerplate looking like this:

```cs
public async Task TextDocumentCompletion(JToken token, CancellationToken token) {
   var request = token.ToObject<CompletionParams>();
}
```

It could look like this:

```cs
public async Task TextDocumentCompletion(CompletionParams request, CancellationToken token) {
}
```

milopezc:
Closes #289

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -986,7 +986,7 @@ public abstract class JsonRpcTests : TestBase
         Assert.Equal(2, sum);
     }
 
-    [SkippableFact]
+    [Fact]
     public async Task InvokeWithSingleNonObjectParameter()
     {
         int sum = await this.clientRpc.InvokeWithParameterObjectAsync<int>(nameof(Server.MethodWithOneNonObjectParameter), new { x = 2 }, this.TimeoutToken);
@@ -1816,17 +1816,6 @@ public abstract class JsonRpcTests : TestBase
         public static int MethodWithInvalidProgressParameter(Progress<int> p)
         {
             return 1;
-        }
-
-        public static int MethodWithOnlyOneFuzzyParameter(int fuzzyX)
-        {
-            return fuzzyX * 2;
-        }
-
-        public static int MethodWithTwoFuzzyParameters(int fuzzyX, int fuzzyY)
-        {
-            int sum = fuzzyX + fuzzyY;
-            return sum;
         }
 
         public int? MethodReturnsNullableInt(int a) => a > 0 ? (int?)a : null;
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -2199,7 +2199,6 @@ public abstract class JsonRpcTests : TestBase
             set;
         }
 
-
         [DataMember]
         public IProgress<int> P
         {
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -942,72 +942,43 @@ public abstract class JsonRpcTests : TestBase
     }
 
     [SkippableFact]
-    public async Task InvokeWithSingleObjectParameter_SendingWithProgressFieldButServerDoesNotExpectIt()
+    public async Task InvokeWithSingleObjectParameter_SendingExpectedObject()
     {
         Skip.If(this.clientMessageFormatter is MessagePackFormatter, "Single object deserialization is not supported for MessagePack");
 
-        int sum = await this.clientRpc.InvokeWithParameterObjectAsync<int>(nameof(Server.MethodWithObjectWithFieldsParameter), new XAndYFieldsWithProgress { x = 2, y = 5, p = new Progress<int>() }, this.TimeoutToken);
+        int sum = await this.clientRpc.InvokeWithParameterObjectAsync<int>("test/MethodWithSingleObjectParameter", new XAndYFields { x = 2, y = 5 }, this.TimeoutToken);
         Assert.Equal(7, sum);
-    }
-
-    [SkippableFact]
-    public async Task InvokeWithSingleObjectParameter_SendingWithProgressFieldButServerDoesNotExpectItAndDefaultParam()
-    {
-        Skip.If(this.clientMessageFormatter is MessagePackFormatter, "Single object deserialization is not supported for MessagePack");
-
-        int sum = await this.clientRpc.InvokeWithParameterObjectAsync<int>(nameof(Server.MethodWithObjectWithFieldsParameter), new XAndYFieldsWithProgress { x = 2, p = new Progress<int>() }, this.TimeoutToken);
-        Assert.Equal(2, sum);
-    }
-
-    [SkippableFact]
-    public async Task InvokeWithSingleObjectParameter_SendingWithProgressPropertyButServerDoesNotExpectIt()
-    {
-        Skip.If(this.clientMessageFormatter is MessagePackFormatter, "Single object deserialization is not supported for MessagePack");
-
-        int sum = await this.clientRpc.InvokeWithParameterObjectAsync<int>(nameof(Server.MethodWithObjectWithPropertiesParameter), new XAndYPropertiesWithProgress { X = 2, Y = 5,  P = new Progress<int>() }, this.TimeoutToken);
-        Assert.Equal(7, sum);
-    }
-
-    [SkippableFact]
-    public async Task InvokeWithSingleObjectParameter_SendingWithProgressPropertyButServerDoesNotExpectItAndDefaultParam()
-    {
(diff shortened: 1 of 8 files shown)


src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1657,6 +1765,27 @@ public static int MethodWithDefaultParameter(int x, int y = 10)
...
+        public static int MethodWithObjectWithPropertiesParameter(XAndYProperties props)
+        {
+            return props.X + props.Y;
+        }
+
+        public static int MethodWithObjectWithProgressParameter(XAndYPropertiesWithProgress props)
+        {
+            props.P.Report(props.X + props.Y);
+            return props.X + props.Y;
+        }
```
AArnott: Where are the tests for methods that take these single parameters plus the (specially treated) `CancellationToken` parameter?

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -583,17 +583,55 @@ internal JsonRpcRequest(JsonMessageFormatter formatter)
...
+                        }
+
+                        typedArguments[0] = obj;
+                        return ArgumentMatchResult.Success;
                     }
 
-                    typedArguments[0] = obj;
-                    return ArgumentMatchResult.Success;
+                    // Check if it is constructable with default constructor
+                    if (parameters[0].ParameterType.GetConstructor(Type.EmptyTypes) != null && !parameters[0].ParameterType.IsAbstract)
```
AArnott: I think this is a breaking change, since any method with just one parameter whose type as a default constructor would not be interpreted differently yet it was a valid scenario before this change. It also deviates from the JSON-RPC spec. Given those, I feel like this should be an explicit opt-in. Our `JsonRpcMethodAttribute` might be the best way to do this. Suppose we add a property to that attribute to indicate that we want the method's one parameter to receive the deserialized result of the entire parameter object? 
The server might also return an error to the client if it doesn't use parameter object syntax in this case, since an array of values can't be reliably matched to named members on the CLR object.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -583,17 +583,55 @@ internal JsonRpcRequest(JsonMessageFormatter formatter)
...
+                                {
+                                    fieldInfo.SetValue(obj, value.ToObject(fieldInfo.FieldType, this.formatter.JsonSerializer));
+                                    match = true;
+                                }
+                            }
+                        }
+
+                        typedArguments[0] = obj;
+
+                        // If one of the arguments matched the parameters we return ArgumentMatchResult.Success
```
AArnott: This criteria of allowing just one parameter/property name match sounds like you're mixing in #288 with #289. So if nothing else I think your PR description should call out that it closes #288 as well. But IMO this should be its own opt-in. Perhaps another `JsonRpcMethodAttribute` property that you add to `AllowFuzzyMatching`.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -583,17 +583,55 @@ internal JsonRpcRequest(JsonMessageFormatter formatter)
...
+                    {
+                        // If the method is expecting only one parameter we should try to create the expected object with the given arguments
+                        var obj = Activator.CreateInstance(parameters[0].ParameterType);
+                        bool match = false;
+
+                        // Loop value-key pairs from arguments
+                        foreach (KeyValuePair<string, object> property in this.NamedArguments)
+                        {
+                            // If argument name matches property name, assign the value to that property
+                            PropertyInfo propertyInfo = parameters[0].ParameterType.GetProperty(property.Key);
```
AArnott: I believe for both `GetProperty` and `GetField` you should constrain the discovery using `BindingFlags` to `Public | Instance` members.

src/StreamJsonRpc/JsonMessageFormatter.cs
milopezc: I see, so for #289 should it be something similar to what we do for JToken, only difference would be that we deserialize it to the specific type before assigning to typedArguments[0] right? Something like:
`typedArguments[0] = obj.ToObject(parameters[0].ParameterType);`

---
In reply to: [336723042](https://github.com/microsoft/vs-streamjsonrpc/pull/347#discussion_r336723042) [](ancestors = 336723042)

AArnott: Ya, that sounds about right.
One trick may be though is determining when to allow this. We don't currently tunnel properties on `JsonRpcMethodAttribute` to the formatter, which may be required if you go with the way I was proposing.

milopezc: I refactored the logic on the latest update

src/StreamJsonRpc/JsonMessageFormatter.cs
milopezc: I added a property on `JsonMethodAttribute` to indicate whether the method supports this or not.

src/StreamJsonRpc.Tests/JsonRpcTests.cs
milopezc: Test added in latest update.

src/StreamJsonRpc/PublicAPI.Shipped.txt
```diff
@@ -41,7 +41,7 @@ StreamJsonRpc.JsonRpc.InvokeAsync<TResult>(string targetName, object argument) -
 StreamJsonRpc.JsonRpc.InvokeWithParameterObjectAsync<TResult>(string targetName, object argument = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task<TResult>
 StreamJsonRpc.JsonRpc.NotifyWithParameterObjectAsync(string targetName, object argument = null) -> System.Threading.Tasks.Task
 StreamJsonRpc.JsonRpcMethodAttribute
-StreamJsonRpc.JsonRpcMethodAttribute.JsonRpcMethodAttribute(string name) -> void
+StreamJsonRpc.JsonRpcMethodAttribute.JsonRpcMethodAttribute(string name, bool useSingleObjectParameterDeserialization = false) -> void
```
AArnott: This is a breaking API change. Instead of changing the existing method, add an overload that takes both parameters.

AArnott: I pushed a commit to take care of it. But any time you make a change that impacts PublicAPI.Shipped.txt you should carefully consider the breaking API impact.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -54,14 +54,14 @@ public class JsonRpc : IDisposableObservable
         /// A map of outbound calls awaiting responses.
         /// Lock the <see cref="dispatcherMapLock"/> object for all access to this member.
         /// </summary>
-        private readonly Dictionary<RequestId, OutstandingCallData> resultDispatcherMap = new Dictionary<RequestId, OutstandingCallData>();
+        private readonly Dictionary<long, OutstandingCallData> resultDispatcherMap = new Dictionary<long, OutstandingCallData>();
 
         /// <summary>
```
AArnott: Being internal means this magic trick isn't reproducible by other (external) formatters. I like that it's your default, but I'd like to explore whether we can make the functionality available publicly. 
Your method as it stands is very focused on your feature. If however we change it to simply return the `JsonRpcMethodAttribute` for *any* RPC method, then it will likely be useful not just for this but other future features, which makes me feel better about making it public.
I'm going to play around with this idea and may push a commit to your PR to see what you think, if that's OK with you.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1042,6 +1042,32 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
+        /// <summary>
+        /// Tries to obtain the <see cref="MethodSignatureAndTarget"/> info of a method.
+        /// </summary>
+        internal bool MethodSupportsSingleParameterObjectDeserialization(string methodName, Type parameterType)
```
AArnott: Being internal means this magic trick isn't reproducible by other (external) formatters. I like that it's your default, but I'd like to explore whether we can make the functionality available publicly. 
Your method as it stands is very focused on your feature. If however we change it to simply return the `JsonRpcMethodAttribute` for *any* RPC method, then it will likely be useful not just for this but other future features, which makes me feel better about making it public.
I'm going to play around with this idea and may push a commit to your PR to see what you think, if that's OK with you.

AArnott:
@milopezc Now that I've pushed a few commits to this PR, can you review as well? If you approve too, we'll complete this.

PR #348, opened 2019-10-19
Add support for custom (string based) request IDs

### LINKED ISSUE #329: serialize the id field as a string
Would the vs-streamjsonrpc maintainers consider adding an option/setting to serialize the **id** field as a string?

```
{
    "jsonrpc": "2.0",
    "method": "ExampleMethod",
    "params": {},
    "id": "pt-1"
}
```

---
Here is the reasoning behind this request:

I am connecting to a terminal via jsonrpc. The terminal follows the JSON/RPC specifications, but defines the following additional constraint:

> The id field must be a string. The suggested format is "x-<num>" where x is the endpoint short name (e.g. pt for payment terminal) and <num> is a counter starting from 1 and increasing after every method request sent by that party. Example: "pt-1"

Right now the library serializes the id generated by the invoke* methods as integers. It was easy enough to fork the streamjsonrpc library to convert that id to a string. And that was enough to fullfil the terminal additional constraint. 
```
  protected async Task<TResult> InvokeCoreAsync<TResult>(long? id, string targetName, IReadOnlyList<object> arguments, CancellationToken cancellationToken, bool isParameterObject)
        {
            var request = new JsonRpcRequest
            {
                Id = id.toString(), // <--- toString() added here.
                Method = targetName,
         };
```
Would the vs-streamjsonrpc maintainers consider adding an option/setting to serialize the id field as a string? Optionally, providing the user an option to use a define an id_prefix would be a nice addition.

AArnott:
This adds a virtual method to the JsonRpc class that allows advanced scenarios that require a specially customized request ID format. It creates a `RequestId` struct to encapsulate the string-or-integer that serves as the request ID and removes all assumptions from our code that our own locally generated request ID is an integer.

This makes some breaking public API changes, which I would typically avoid within a major release. Feel free to push back if you think this is a problem. I am thinking it's not as I've limited them to just the very new formatter-helper classes which are only used within custom formatter implementations of which there are few or none outside this repo at this point AFAIK.

Closes #329

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -83,6 +83,18 @@ namespace StreamJsonRpc
         private readonly MessageFormatterDuplexPipeTracker duplexPipeTracker = new MessageFormatterDuplexPipeTracker();
 
         /// <summary>
+        /// A value indicating whether a request where <see cref="Protocol.JsonRpcRequest.RequestId"/> is a <see cref="string"/>
+        /// has been transmitted.
+        /// </summary>
+        /// <remarks>
+        /// This is useful to detect whether <see cref="JsonRpc"/> is operating in its default mode of producing
+        /// integer-based values for <see cref="Protocol.JsonRpcRequest.RequestId"/>, which informs us whether we should
+        /// type coerce strings in response messages back to integers to accomodate JSON-RPC servers
+        /// that improperly convert our integers to strings.
+        /// </remarks>
+        private bool observedTransmittedRequestWithStringId;
+
+        /// <summary>
         /// The version of the JSON-RPC protocol being emulated by this instance.
         /// </summary>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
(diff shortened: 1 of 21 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -535,15 +537,16 @@ private JsonRpcResult ReadResult(JToken json)
         {
             Requires.NotNull(json, nameof(json));
 
-            JToken id = json["id"];
+            RequestId id = this.NormalizeId(json["id"].ToObject<RequestId>());
```
milopezc: Don't we need a null check here just as in ReadRequest?

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -552,15 +555,15 @@ private JsonRpcError ReadError(JToken json)
...
```
milopezc: Same as above, isn't null check here necessary?

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: No. Results *always* have id properties.

Issue #246, opened 2019-03-07
Stabilize Dispose patterns


AArnott:
Customers have reported a few times that they want to Dispose of JsonRpc (or indirectly its `MessageHandlerBase`):
1. from their `Disconnected` event handler
1. while outbound messages are still in progress or perhaps even in transmission.

We know that disposing the `MessageHandlerBase` while messages are being transmitted is very problematic (e.g. https://github.com/dotnet/roslyn/issues/18563).
And I suspect disposing from our event handler has undefined behavior.

1. I think we should add a test for the disposal from our event handler if we can make that work (for user convenience).
1. We should stop disposing our `AsyncSemaphore` in `MessageHandlerBase.Dispose`. There are no native resources to release, and this would avoid random exception types being thrown on our transmission thread, sometimes leading to crashes. Instead, we can set a boolean field so that any *future* transmission request is denied with an `ObjectDisposedException`.

Also: when is `ObjectDisposedException` thrown compared to `ConnectionLostException`? Even `OperationCanceledException` can be thrown back to pending requests due to a disconnection.

AArnott:
I've re-opened https://github.com/Microsoft/vs-threading/pull/105 where we might place the fix for no. 2 above.

AArnott:
No. 2 above is indeed resolved by https://github.com/microsoft/vs-threading/pull/493

tmat:
This would be very useful. We need to get to a state when remote calls can only throw 4 types of exceptions that correspond to
1) Connection issue (connection dropped for any reason)
2) Serialization issue - bug in serialization of arguments (types are not serializable, etc.)
3) Remote exception - an exception was thrown by the callee
4) Cancelation

PR #352, opened 2019-11-02
Only complete PipeReader/PipeWriter after operations complete

### LINKED ISSUE #350: HeaderDelimitedMessageHandler.Write throws when in race with disposal
If the `PipeMessageHandler` is writing (or reading) while being disposed, it can create a memory corruption and/or exception situation when the PipeWriter or PipeReader or their buffers are accessed after being completed.

VS has recently shown a tendency to fail with an infobar with this error:

```System.InvalidOperationException: Writing is not allowed after writer was completed.
 at System.IO.Pipelines.Pipe.GetMemory(Int32 sizeHint)
 at System.IO.Pipelines.Pipe.DefaultPipeWriter.GetMemory(Int32 sizeHint)
 at StreamJsonRpc.HeaderDelimitedMessageHandler.Write(JsonRpcMessage content, CancellationToken cancellationToken)
 at StreamJsonRpc.PipeMessageHandler.WriteCoreAsync(JsonRpcMessage content, CancellationToken cancellationToken)
 at StreamJsonRpc.MessageHandlerBase.<WriteAsync>d__15.MoveNext() --- End of stack trace from previous location where exception was thrown ---
 at StreamJsonRpc.JsonRpc.<TransmitAsync>d__118.MoveNext() --- End of stack trace from previous location where exception was thrown ---
 at System.Threading.Tasks.ValueTask.ThrowIfCompletedUnsuccessfully()
 at StreamJsonRpc.JsonRpc.<InvokeCoreAsync>d__102.MoveNext() --- End of stack trace from previous location where exception was thrown ---
 at StreamJsonRpc.JsonRpc.<InvokeCoreAsync>d__96`1.MoveNext() --- End of stack trace from previous location where exception was thrown ---
 at Microsoft.VisualStudio.CodeLens.Proxy.CodeLensHubClient.<>c__DisplayClass22_0`1.<<InvokeRpcAsync>b__0>d.MoveNext() --- End of stack trace from previous location where exception was thrown ---
 at Microsoft.VisualStudio.Text.Utilities.GuardedOperations.<CallExtensionPointAsync>d__36`1.MoveNext() --- End of stack trace from previous location where exception was thrown ---
 at Microsoft.VisualStudio.Telemetry.WindowsErrorReporting.WatsonReport.GetClrWatsonExceptionInfo(Exception exceptionObject)
```

AArnott:
We must not allow a race to exist between PipeReader/PipeWriter use and calling Complete() on them. This not only produces exceptions like the one in #350 but it opens up the possibility of accessing recycled buffers.

Fixes #350

### DIFF
--- a/src/StreamJsonRpc/MessageHandlerBase.cs
+++ b/src/StreamJsonRpc/MessageHandlerBase.cs
@@ -36,6 +36,16 @@ namespace StreamJsonRpc
         private readonly AsyncSemaphore sendingSemaphore = new AsyncSemaphore(1);
 
         /// <summary>
+        /// The sync object to lock when inspecting and mutating the <see cref="state"/> field.
+        /// </summary>
+        private readonly object syncObject = new object();
+
+        /// <summary>
+        /// A value indicating whether the <see cref="ReadAsync(CancellationToken)"/> method is in progress.
+        /// </summary>
+        private MessageHandlerState state;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="MessageHandlerBase"/> class.
         /// </summary>
         /// <param name="formatter">The formatter used to serialize messages.</param>
@@ -43,6 +53,30 @@ namespace StreamJsonRpc
         {
             Requires.NotNull(formatter, nameof(formatter));
             this.Formatter = formatter;
+        }
+
+        [Flags]
+        private enum MessageHandlerState
+        {
+            /// <summary>
+            /// No flags.
+            /// </summary>
+            None = 0,
+
+            /// <summary>
+            /// Indicates that the <see cref="WriteAsync(JsonRpcMessage, CancellationToken)"/> method is in running.
+            /// </summary>
+            Writing = 0x1,
+
+            /// <summary>
+            /// Indicates that the <see cref="ReadAsync(CancellationToken)"/> method is in running.
+            /// </summary>
+            Reading = 0x4,
+
+            /// <summary>
+            /// Indicates that the <see cref="Dispose(bool)"/> method has been called.
+            /// </summary>
+            DisposeVirtualMethodInvoked = 0x10,
         }
 
         /// <summary>
@@ -85,6 +119,7 @@ namespace StreamJsonRpc
             Verify.Operation(this.CanRead, "No receiving stream.");
             cancellationToken.ThrowIfCancellationRequested();
             Verify.NotDisposed(this);
+            this.SetState(MessageHandlerState.Reading);
 
             try
             {
@@ -102,6 +137,13 @@ namespace StreamJsonRpc
                 // If already canceled, throw that instead of ObjectDisposedException.
                 cancellationToken.ThrowIfCancellationRequested();
                 throw;
+            }
+            finally
+            {
+                if (this.CheckIfDisposalAppropriate(MessageHandlerState.Reading))
+                {
+                    this.Dispose(true);
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/StreamJsonRpc/MessageHandlerBase.cs
```diff
@@ -35,6 +35,16 @@ public abstract class MessageHandlerBase : IJsonRpcMessageHandler, IDisposableOb
         /// </summary>
         private readonly AsyncSemaphore sendingSemaphore = new AsyncSemaphore(1);
 
+        /// <summary>
+        /// The sync object.
```
jepetty: Small nit: could we add something about this locking/protecting specifically the state of the MessageHandler?

AArnott: Good idea.

davidfowl:
Why dispose in Read and Write itself instead of Dispose?

AArnott:
@davidfowl We dispose in all 3 places (read, write, dispose), whichever comes *last*. The reason is because we auto-dispose when the connection drops, or whenever the user calls Dispose explicitly. But if we're in the middle of a write/read operation, that's the "use after dispose" case you warned me about and that we're seeing exceptions regarding. So we can't Complete during a read/write operation, and our Dispose is synchronous, so the best we can do is defer completion until after the read/write operations complete.

davidfowl:
>  The reason is because we auto-dispose when the connection drops, or whenever the user calls Dispose explicitly.

Don't auto-dispose when the connection drops. Only when they call dispose, that's how you know they are done with the handler.

AArnott:
> Don't auto-dispose when the connection drops. Only when they call dispose, that's how you know they are done with the handler.

This is prior shipped behavior, so it would be difficult to change. But anyway, the fact that they call `Dispose` does not guarantee that we're done reading, because they really have no way of knowing whether a message was coming in just as they decided to dispose.

davidfowl:
> This is prior shipped behavior, so it would be difficult to change. 

What's the observed behavior from the caller's end? Can't that be preserved?

> But anyway, the fact that they call Dispose does not guarantee that we're done reading, because they really have no way of knowing whether a message was coming in just as they decided to dispose.

But that code would throw an object disposed exception, because you would cancel any pending operations, then dispose.

AArnott:
We do have #246 to better define exactly what the disposal behaviors and expectations are.

From this completed PR we already have a failing test in the VS repo because the test verified that the underlying pipe had been closed immediately after JsonRpc.Dispose() returned. So I have to go back and fix that, at least. I'd rather not return from Dispose() before *all* async/concurrent code has stopped executing, since anything less can result in crashes when the AppDomain is torn down if that code is still active.

> But that code would throw an object disposed exception, because you would cancel any pending operations, then dispose.

It's not enough to cancel then dispose, I think. I have to wait for the canceled work to complete. Otherwise I might still be disposing the PipeReader/PipeWriter while it's being accessed on another thread but before that thread has observed the cancellation request.

davidfowl:
Fair, this is the safest approach then (especially since there's no dispose async)

AArnott:
Thanks for your review.

However this PR is faulty after all, as it blocked a VS insertion due to a failing test, since Dispose() returned before all the async cleanup was done. So I have a change nearly ready for another PR that (wait for it...) adds a DisposeAsync() method. Since you said "especially since there's no dispose async", I'm very interested in hearing your concerns about that approach. I'll include you on the PR when it's ready so you can see more specifically what I'm talking about.

PR #357, opened 2019-11-04
Enable nullable ref types

AArnott:
This enables the C# 8 feature called "nullable reference types" that helps us internally as well as our users to avoid `NullReferenceException` and otherwise understand where we may accept or return null values in our API.


src/StreamJsonRpc.Tests/FullDuplexStream.cs
```diff
@@ -32,13 +30,15 @@ public class FullDuplexStream : Stream
         /// <summary>
         /// The stream to write to.
         /// </summary>
+#pragma warning disable CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
         private FullDuplexStream other;
+#pragma warning restore CS8618 // Non-nullable field is uninitialized. Consider declaring as nullable.
```
sharwell: ❔ Why is this one excluded?

src/StreamJsonRpc.Tests/JsonRpcClient20InteropTests.cs
```diff
@@ -301,7 +301,7 @@ public async Task ErrorResponseIncludesCallstack()
         };
         this.Send(errorObject);
         var ex = await Assert.ThrowsAsync<RemoteInvocationException>(() => requestTask);
-        var commonErrorData = ((JToken)ex.ErrorData).ToObject<CommonErrorData>();
+        var commonErrorData = ((JToken?)ex.ErrorData)!.ToObject<CommonErrorData>();
```
sharwell: 💭 I would have taken the simpler route for this one

```suggestion
        var commonErrorData = ((JToken)ex.ErrorData!).ToObject<CommonErrorData>();
```

src/StreamJsonRpc.Tests/JsonRpcCustomRequestIdTests.cs
```diff
@@ -60,7 +60,7 @@ public JsonRpcWithStringIds(IJsonRpcMessageHandler handler)
 
         protected override RequestId CreateNewRequestId()
         {
-            return new RequestId("prefix-" + base.CreateNewRequestId().Number.Value.ToString());
+            return new RequestId("prefix-" + base.CreateNewRequestId().Number?.ToString());
```
sharwell: 💭 Why have the explicit `ToString()`?

```suggestion
            return new RequestId("prefix-" + base.CreateNewRequestId().Number);
```

src/StreamJsonRpc.Tests/MessagePackFormatter.cs
```diff
@@ -108,20 +108,20 @@ public void Serialize(IBufferWriter<byte> contentBuffer, JsonRpcMessage message)
         /// doesn't expose a simple way of doing this, so we'd have to emulate it by supporting both
         /// <see cref="DataMemberAttribute.Name"/> and <see cref="KeyAttribute.StringKey"/> handling.
         /// </remarks>
-        private static Dictionary<string, object> GetParamsObjectDictionary(object paramsObject)
+        private static Dictionary<string, object?>? GetParamsObjectDictionary(object paramsObject)
```
sharwell:
```suggestion
        [return: NotNullIfNotNull("paramsObject")]
        private static Dictionary<string, object?>? GetParamsObjectDictionary(object? paramsObject)
```

src/StreamJsonRpc.Tests/PerfTests.cs
```diff
@@ -151,7 +151,9 @@ public async Task BurstInvokeMessages(Stream serverStream, Stream clientStream)
         }
     }
 
+#pragma warning disable SA1414 // Tuple types in signatures should have element names
```
sharwell: 💭 This one seems like an easy fix to just add the names

src/shipping.ruleset
```diff
@@ -84,6 +84,7 @@
     <Rule Id="SA1639" Action="Hidden" />
     <Rule Id="SA1640" Action="Hidden" />
     <Rule Id="SA1641" Action="Hidden" />
+    <Rule Id="SA1316" Action="None" />
```
sharwell: 📝 This rule is configurable in **stylecop.json** (camelCase or PascalCase).

src/StreamJsonRpc.Tests/FullDuplexStream.cs
AArnott: This field creates a circular reference loop with another instance of this type. So it's impossible for this field to be set in the constructor.

src/StreamJsonRpc.Tests/PerfTests.cs
AArnott: Easy fix, but IMO inappropriate since there are literally no semantic names we could add over the ones that ValueTuple provides. We just have two streams.

src/shipping.ruleset
AArnott: Awesome. Because I simply disagree with the default camelCase, since they are exposed as public properties, which generally are supposed to be PaslcalCased.

PR #359, opened 2019-11-06
Provide deterministic way to await for async disposal

AArnott:
This fixes up a regression from #352 that came from the fact that `JsonRpc.Dispose()` would return prior to full disposal of the `IJsonRpcMessageHandler` behind it. A VS unit test caught this when it asserted the named pipe was closed after Dispose() and it started failing after my last change.

The plan here is still that Dispose() will not block, but the `JsonRpc.Completion` property will now capture the async closure of the message handler where it didn't before.

Part of full fix for #350


davidfowl:
Too much continue with in this PR. Needs more async await

AArnott:
Thanks for reviewing, @davidfowl. I prefer async/await, but I used `ContinueWith` here because I didn't want to define separate methods for each of these one-line delegates, and seeing how they are all scheduled to depend on each other is clearer with 3 consecutive ContinueWith lines than spread out across several methods that would all start with "await something else".

davidfowl:
😢

PR #367, opened 2019-11-11
Add built-in MessagePack formatter

### LINKED ISSUE #365: StreamJsonRpc.RequestId is not MessagePack formattable by default.
When using StreamJsonRpc with MessagePack, the RequestId object cannot be serialized using the default formatters.

You get an error/stack like this when attempting to do so:

```
System.TypeInitializationException: The type initializer for 'FormatterCache`1' threw an exception. ---> MessagePack.Internal.MessagePackDynamicObjectResolverException: can't find matched constructor. type:StreamJsonRpc.RequestId
   at MessagePack.Internal.ObjectSerializationInfo.CreateOrNull(Type type, Boolean forceStringKey, Boolean contractless, Boolean allowPrivate)
   at MessagePack.Internal.DynamicObjectTypeBuilder.BuildFormatterToDynamicMethod(Type type, Boolean forceStringKey, Boolean contractless, Boolean allowPrivate)
   at MessagePack.Resolvers.DynamicContractlessObjectResolverAllowPrivate.FormatterCache`1..cctor()
```

AArnott:
Adds `MessagePackFormatter` as a built-in formatter, and enhances the tests to cover more scenarios than the sample had.

Closes #365

### DIFF
--- a/src/StreamJsonRpc/MessagePackFormatter.cs
+++ b/src/StreamJsonRpc/MessagePackFormatter.cs
@@ -46,15 +46,9 @@ namespace StreamJsonRpc
         private MessagePackSerializerOptions userDataSerializationOptions = MessagePackSerializerOptions.Standard;
 
         /// <summary>
-        /// The <see cref="IFormatterResolver"/> that was fed into the <see cref="GetWrappersForUserSuppliedResolver(IFormatterResolver, out MessagePackSerializerOptions, out IMessagePackFormatter{object?})"/> method.
-        /// Backing field for the <see cref="Resolver"/> property.
-        /// </summary>
-        private IFormatterResolver userSuppliedResolver;
-
-        /// <summary>
         /// The formatter to use when serializing user data that we only see typed as <see cref="object"/>.
         /// </summary>
-        private IMessagePackFormatter<object?> dynamicObjectTypeFormatterForUserSuppliedResolver;
+        private DynamicObjectTypeFallbackFormatter dynamicObjectTypeFormatterForUserSuppliedResolver;
 
         /// <summary>
         /// Backing field for the <see cref="IJsonRpcInstanceContainer.Rpc"/> property.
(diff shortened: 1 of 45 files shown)


src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -0,0 +1,1235 @@
...
+            if (message is IJsonRpcMessageWithId msgWithId && (message is Protocol.JsonRpcResult || message is Protocol.JsonRpcError))
+            {
+                this.duplexPipeTracker.OnResponseSent(msgWithId.RequestId, successful: msgWithId is Protocol.JsonRpcResult);
+            }
+
+            if (message is Protocol.JsonRpcRequest request && request.Arguments != null && request.ArgumentsList == null && !(request.Arguments is IReadOnlyDictionary<string, object?>))
+            {
+                // This request contains named arguments, but not using a standard dictionary. Convert it to a dictionary so that
+                // the parameters can be matched to the method we're invoking.
+                request.Arguments = GetParamsObjectDictionary(request.Arguments);
```
peterwald: In the TestWindow, we have always used `new object[]{p0,p1}` for our arguments. It that using named arguments would be both larger and much slower. Is that your experience, and should we note that in the README?

AArnott: Named arguments do require some additional machinery, it's true. But on the whole of an RPC communication I'm not aware of significant perf issues that are unique to named args. It's something we can measure.
FWIW I find positional args generally more familiar than named args, as virtually every language uses positional args rather than named args by default. But we support named args because some servers _require_ them (e.g. LSP). So rather than document a recommendation, I think the best thing we can do is support both as well as possible and let folks use what they need to.
Also in light of #288 named args might become more preferred as they may allow for greater version compatibility across method signature changes, just FYI.

src/StreamJsonRpc/Reflection/MessageFormatterDuplexPipeTracker.cs
```diff
@@ -92,6 +92,7 @@ public MessageFormatterDuplexPipeTracker()
         /// When the response is received, a call should always be made to <see cref="OnResponseReceived(RequestId, bool)"/>.
         /// </remarks>
         /// <exception cref="NotSupportedException">Thrown if no <see cref="MultiplexingStream"/> was provided to the constructor.</exception>
+        [return: NotNullIfNotNull("token")]
```
karpinsn: Shouldn't this be duplexPipe? Similar for GetToken below as well

AArnott: Yes. I'll fix that. That seems like something C# ought to have an analyzer to catch. @sharwell, are you aware of anything on the backlog for this?

sharwell: I believe it's item 11 in #36073

PR #385, opened 2019-11-27
Replace specialized tracker code in formatters with general APIs

AArnott:
Each tracker came with a burden on each formatter that used it that it had to be called at special times and in special ways. This change leverages some recently introduced interfaces so the trackers can take care of most needs themselves, simplifying the job of the formatters.

This includes breaking API changes to the trackers, but we've already made such changes in 2.3, and on the same basis that we aren't aware of any external users of it, it seems like a reasonable opportunity to simplify it.

The `IProgress<T>` tracker still has one special requirement because of when cleanup happens relative to the $/progress notifications, sadly. When we treat $/progress as just an ordinary method call it can be processed *after* the actual result from the method is received and its callback invoked, which creates a "lossy" progress report because any that are processed on the client after the result callback get dropped. So I ensure $/progress reporting happens directly from the formatter (as it was before) to avoid a significant change in behavior.

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -73,15 +73,23 @@ namespace StreamJsonRpc
         private readonly SequenceTextReader sequenceTextReader = new SequenceTextReader();
 
         /// <summary>
+        /// Backing field for the <see cref="MultiplexingStream"/> property.
+        /// </summary>
+        private MultiplexingStream? multiplexingStream;
+
+        /// <summary>
         /// <see cref="MessageFormatterProgressTracker"/> instance containing useful methods to help on the implementation of message formatters.
         /// </summary>
-        private readonly MessageFormatterProgressTracker formatterProgressTracker = new MessageFormatterProgressTracker();
+        private MessageFormatterProgressTracker? formatterProgressTracker;
 
         /// <summary>
         /// The helper for marshaling pipes as RPC method arguments.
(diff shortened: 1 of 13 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -913,12 +939,12 @@ public DuplexPipeConverter(JsonMessageFormatter jsonMessageFormatter)
...
             {
                 int? tokenId = JToken.Load(reader).Value<int?>();
-                return this.jsonMessageFormatter.duplexPipeTracker.GetPipe(tokenId);
+                return this.jsonMessageFormatter.duplexPipeTracker!.GetPipe(tokenId);
             }
 
             public override void WriteJson(JsonWriter writer, IDuplexPipe? value, JsonSerializer serializer)
             {
-                var token = this.jsonMessageFormatter.duplexPipeTracker.GetToken(value);
+                var token = this.jsonMessageFormatter.duplexPipeTracker!.GetToken(value);
```
milopezc: Just out of curiosity, how does the ! operator work? I couldn't find anything online about it. (Just trying to learn something here)

AArnott: I should have used my new `DuplexPipeTracker` property here. I'll fix that in a follow-up PR.

The `!` is the ["null forgiving operator"](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving).

PR #391, opened 2019-12-02
Workaround issues on mono and get tests running

### LINKED ISSUE #387: Getting a StreamJsonRpc.RemoteInvocationException when invoking a server->client RPC on Mono.
Minimal source code that reproduces this error: https://github.com/ILikePizza555/StreamJsonRPCTest

This happens on MacOS on the latest version of Mono (6.4.0.198).

Output of the project is

```
Entered RequestDataFromServerRPC
Entered RequestDataFromClientRPC
RPC Exception: StreamJsonRpc.RemoteInvocationException: Attempted to access an element as a type incompatible with the array.
  at StreamJsonRpc.JsonRpc.InvokeCoreAsync[TResult] (System.Nullable`1[T] id, System.String targetName, System.Collections.Generic.IReadOnlyList`1[T] arguments, System.Threading.CancellationToken cancellationToken, System.Boolean isParameterObject) [0x0018d] in <ab9df841741b4f729c437e4366755293>:0 
  at SJRTest.Server.RequestDataFromClientRPC (System.String key, System.Threading.CancellationToken ct) [0x0009b] in <33bed87fd5be4d728d3def5bfad1c0af>:0 
Result: 
```

Obviously, the expected behavior is to not have the exception. The fact that the application does not print out `Entered DictionaryDataCallback` indicates that this exception is thrown before the method is called.

AArnott:
Add a workaround to a mono bug so StreamJsonRpc can run on mono (and support VS 4 Mac).
This also gets all our unit tests running on mono on Azure Pipelines except for the `PerfTests` class, which causes mono to crash.

Fixes #387

### DIFF
--- a/azure-pipelines/dotnet.yml
+++ b/azure-pipelines/dotnet.yml
@@ -43,6 +43,10 @@ steps:
     testRunTitle: netcoreapp2.2-$(Agent.JobName)
     workingDirectory: src
 
+- powershell: mono $(NUGET_PACKAGES)/xunit.runner.console/2.4.1/tools/net472/xunit.console.exe bin/StreamJsonRpc.Tests/$(BuildConfiguration)/net472/StreamJsonRpc.Tests.dll -vsts -notrait "TestCategory=FailsInCloudTest" -noclass PerfTests
+  displayName: Run tests on mono
+  condition: ne(variables['Agent.OS'], 'Windows_NT') # The Windows Hosted agent doesn't include mono
+
 # We have to artifically run this script so that the extra .nupkg is produced for variables/InsertConfigValues.ps1 to notice.
 - powershell: azure-pipelines\artifacts\VSInsertion.ps1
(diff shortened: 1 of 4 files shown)


azure-pipelines/dotnet.yml
```diff
@@ -43,6 +43,10 @@ steps:
     testRunTitle: netcoreapp2.2-$(Agent.JobName)
     workingDirectory: src
 
+- powershell: mono $(NUGET_PACKAGES)/xunit.runner.console/2.4.1/tools/net472/xunit.console.exe bin/StreamJsonRpc.Tests/$(BuildConfiguration)/net472/StreamJsonRpc.Tests.dll -vsts -notrait "TestCategory=FailsInCloudTest" -noclass PerfTests
```
AArnott: `-noclass PerfTests` skips tests that crash the test runner on mono due to https://github.com/mono/mono/issues/17994

PR #395, opened 2019-12-04
Include RPC server stack trace in RemoteInvocationException.ToString()

### LINKED ISSUE #369: RemoteInvocationException.ToString() return value should include remote callstack
If an exception occurs remotely and is communicated back across the RPC boundary, you get a RemoteInvocationException.  Calling ToString on it returns a string that includes only the local callstack.  It would be very helpful if the callstack could be made to include the remote portion of the callstack (which is available in the ErrorData property) as well.

AArnott:
It isn't always possible to do, but when the server provides a `CommonErrorData` object, we certainly can do it.
I endeavored to copy the string format that .NET uses for its own exception formatting, including for inner exceptions.

Closes #369

### DIFF
--- a/src/StreamJsonRpc/Exceptions/RemoteInvocationException.cs
+++ b/src/StreamJsonRpc/Exceptions/RemoteInvocationException.cs
@@ -4,6 +4,8 @@
 namespace StreamJsonRpc
 {
     using System;
+    using System.IO;
+    using System.Text;
     using Newtonsoft.Json.Linq;
     using StreamJsonRpc.Protocol;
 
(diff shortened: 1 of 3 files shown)


src/StreamJsonRpc/Exceptions/RemoteInvocationException.cs
```diff
@@ -83,5 +85,56 @@ public RemoteInvocationException(string? message, int errorCode, object? errorDa
         /// The default implementation of this method produces a <see cref="CommonErrorData"/> object.
         /// </remarks>
         public object? DeserializedErrorData { get; }
+
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            string result = base.ToString();
```
JoshuaBStevens: You could avoid calling base.ToString (for efficiency) if you're going to overwrite the result anyway.

src/StreamJsonRpc/Exceptions/RemoteInvocationException.cs
```diff
@@ -83,5 +85,56 @@ public RemoteInvocationException(string? message, int errorCode, object? errorDa
...
+
+            CommonErrorData? errorData = this.DeserializedErrorData as CommonErrorData;
+            if (errorData != null)
+            {
+                var builder = new StringBuilder(base.ToString());
```
JoshuaBStevens: You can use "result" instead of "base.ToString()" here.

src/StreamJsonRpc/Exceptions/RemoteInvocationException.cs
```diff
@@ -83,5 +85,56 @@ public RemoteInvocationException(string? message, int errorCode, object? errorDa
...
         /// </remarks>
         public object? DeserializedErrorData { get; }
+
+        /// <inheritdoc/>
+        public override string ToString()
+        {
+            string result = base.ToString();
+
+            CommonErrorData? errorData = this.DeserializedErrorData as CommonErrorData;
+            if (errorData != null)
```
JoshuaBStevens: These two lines could be written as "if (this.DeserializedErrorData is CommonErrorData errorData)", couldn't they?

src/StreamJsonRpc/Exceptions/RemoteInvocationException.cs
milopezc: This feels right to me in terms of readability. If the ToString() method affects the performance considerably then we should go with this, but I would suggest changing the name of the variable to baseString or something similar.

PR #425, opened 2020-03-02
Update Mac OS hosted version after AzDo warning

jepetty:
The image version we're using for Mac OS is going to be deprecated on March 23rd 2020. This updates to the newest hosted version of that image.

### DIFF
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -31,7 +31,7 @@ jobs:
 
 - job: macOS
   pool:
-    vmImage: macOS 10.13
+    vmImage: macOS 10.15
   steps:
   - checkout: self
     clean: true
--- a/azure-pipelines/build.yml
+++ b/azure-pipelines/build.yml
@@ -31,7 +31,7 @@ jobs:
 
 - job: macOS
   pool:
-    vmImage: macOS 10.15
+    vmImage: macOS-10.15
   steps:
   - checkout: self
     clean: true


azure-pipelines/build.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
 - job: macOS
   pool:
-    vmImage: macOS 10.13
+    vmImage: macOS 10.15
```
AArnott: You need to add a hyphen, [like this](https://github.com/AArnott/Library.Template/commit/ba764031a80c365dbf0182c67fda295fd89aaa5b#diff-6f48d6e0be33ef564632a0a577eb3504).

AArnott:
Also, we should get this fixed in the `v1.5` branch since we still support that version.

azure-pipelines/build.yml
jepetty: Good catch! Thanks

jepetty:
I can start another PR with the same branch to fix 1.5, if that would be the appropriate flow to get the fix in?

AArnott:
You can rebase this commit and then edit this PR's target branch to point at v1.5, or create a new PR for v1.5. 
My primary goal would be to author one commit based on v1.5 that fixes it and then merge that forward to each newer servicing branch till we reach master. That way, just like any other fix to a servicing branch, we can track the commit and see that the fix has been applied to all branches. If we just invent new commits on each branch, it's much harder to track where a fix goes.
I'm pretty sure there will be merge conflicts merging v1.5 to v2.x with this change since the build authoring changed, but we can resolve that conflict by re-applying the right fix from scratch and putting that into the merge commit so it still looks like the same fix everywhere.

AArnott:
I've made the change.

PR #433, opened 2020-03-19
fix a small typo

tebeco:
Spotted a small typo while trying to understand how `MessagePackSerializerOptions` are used (from https://github.com/dotnet/aspnetcore/issues/18569#issuecomment-601330290)

I does not look like a breaking change since it's `private`

(Still a draft since 2 test are failling locally)

I'm not sure why this one fails :
(the `WeakReference` is still alive after collection)
```
    [SkippableFact]
    [Trait("GC", "")]
    [Trait("TestCategory", "FailsInCloudTest")] // Test showing unstability on Azure Pipelines, but always succeeds locally.
    [Trait("FailsOnMono", "true")]
    public async Task InvokeWithProgressParameter_NoMemoryLeakConfirm()
    {
        Skip.If(IsRunningUnderLiveUnitTest);
        WeakReference weakRef = await this.InvokeWithProgressParameter_NoMemoryLeakConfirm_Helper();
        GC.Collect();
        Assert.False(weakRef.IsAlive);
    }
```

### DIFF
--- a/src/StreamJsonRpc/MessagePackFormatter.cs
+++ b/src/StreamJsonRpc/MessagePackFormatter.cs
@@ -116,7 +116,7 @@ namespace StreamJsonRpc
             this.pipeFormatterResolver = new PipeFormatterResolver(this);
 
             // Set up default user data resolver.
-            this.userDataSerializationOptions = this.MassageUserDataOptions(StandardResolverAllowPrivate.Options);
+            this.userDataSerializationOptions = this.MessageUserDataOptions(StandardResolverAllowPrivate.Options);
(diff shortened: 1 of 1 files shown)


AArnott:
Our memory leak tests often fail, I'm afraid. Too much noise in the system. You can disregard them.

AArnott:
But this isn't a typo. We *massage* the user value (as in, we rub it gently to make it what we want). It's not supposed to be "message".

tebeco:
all i can do now if both laugh and "well, ok" :D

PR #446, opened 2020-04-05
Added UseSingleObjectParameterDeserialization option to JsonRpcTargetOptions

### LINKED ISSUE #444: UseSingleObjectParameterDeserialization without adding a attribute to all methods
Hi,

If i want to customize the name mapping between incoming calls and my target methods, i can pass a transform function to JsonRpcTargetOptions so i don't need to add a JsonRpcMethod attribute to all my methods. But if want to use the single object parameter deserialization for all my methods, there is no such option. Is there any way i can centralize this too?

Otherwise, would it be possible to add the UseSingleObjectParameterDeserialization feature to JsonRpcTargetOptions ?

daniel-vetter:
I added the the `UseSingleObjectParameterDeserialization` property to the target options. When the property is set to true, the `GetRequestMethodToClrMethodMap` method will add a `JsonRpcMethodAttribute` to the `MethodSignatureAndTarget`. I don't know if this is the correct approach because `GetJsonRpcMethodAttribute` will now return a attribute, that did not exist on the original method. But it seems like `AddLocalRpcMethod` is also creating fake attributes, so i think it should be ok.

resolves #444

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -68,6 +68,8 @@ public abstract class JsonRpcTests : TestBase
 
         [JsonRpcMethod(UseSingleObjectParameterDeserialization = true)]
         int InstanceMethodWithSingleObjectParameterAndCancellationToken(XAndYFields fields, CancellationToken token);
+
+        int InstanceMethodWithSingleObjectParameterButNoAttribute(XAndYFields fields);
     }
 
     [Fact]
@@ -471,6 +473,20 @@ public abstract class JsonRpcTests : TestBase
     {
         int result = await this.clientRpc.InvokeAsync<int>(nameof(Server.MethodThatMayEndIn));
         Assert.Equal(5, result);
+    }
+
+    [Fact]
+    public async Task CanCallMethodWithAttributeOmittingAsyncSuffix()
+    {
+        int result = await this.clientRpc.InvokeAsync<int>("MethodWithAttributeThatEndsIn");
+        Assert.Equal(6, result);
+    }
+
+    [Fact]
+    public async Task CanCallMethodWithAttributeWithoutOmittingAsyncSuffix()
+    {
+        int result = await this.clientRpc.InvokeAsync<int>(nameof(Server.MethodWithAttributeThatEndsInAsync));
+        Assert.Equal(6, result);
     }
 
     [Fact]
@@ -991,6 +1007,17 @@ public abstract class JsonRpcTests : TestBase
     {
         int sum = await this.clientRpc.InvokeWithParameterObjectAsync<int>(nameof(IServer.InstanceMethodWithSingleObjectParameterAndCancellationToken), new XAndYFields { x = 2, y = 5 }, this.TimeoutToken);
         Assert.Equal(7, sum);
+    }
+
+    [Fact]
+    public async Task AddLocalRpcTarget_UseSingleObjectParameterDeserialization()
+    {
+        var streams = FullDuplexStream.CreatePair();
+        var rpc = new JsonRpc(streams.Item1, streams.Item2);
+        rpc.AddLocalRpcTarget(new Server(), new JsonRpcTargetOptions { UseSingleObjectParameterDeserialization = true });
+        rpc.StartListening();
+
+        Assert.Equal(3, await rpc.InvokeWithParameterObjectAsync<int>(nameof(IServer.InstanceMethodWithSingleObjectParameterButNoAttribute), new XAndYFields { x = 1, y = 2 }));
     }
 
     [Fact]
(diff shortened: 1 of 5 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1419,6 +1419,11 @@ protected virtual void OnRequestTransmissionAborted(JsonRpcRequest request)
 
                     JsonRpcMethodAttribute? attribute = mapping.FindAttribute(method);
 
+                    if (attribute == null && options.UseSingleObjectParameterDeserialization)
+                    {
+                        attribute = new JsonRpcMethodAttribute(null) { UseSingleObjectParameterDeserialization = true };
```
AArnott, line 1424: Synthesizing an attribute is fine, but the way you're doing it you break the code on line 1438 that checks whether `attribute == null` to do other default things.
Maybe we need a boolean local variable to track whether the attribute originally existed.

src/StreamJsonRpc/JsonRpcTargetOptions.cs
```diff
@@ -39,6 +39,11 @@ public class JsonRpcTargetOptions
         /// </remarks>
         public bool AllowNonPublicInvocation { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets a value indicating whether JSON-RPC named arguments should all be deserialized into this method's first parameter.
```
AArnott, line 43:
```suggestion
        /// Gets or sets a value indicating whether JSON-RPC named arguments should all be deserialized into the RPC method's first parameter.
```

src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -993,6 +995,17 @@ public async Task InvokeWithSingleObjectParameter_SendingExpectedObjectAndCancel
...
 
+    [Fact]
+    public async Task AddLocalRpcTarget_UseSingleObjectParameterDeserialization()
+    {
+        var streams = FullDuplexStream.CreatePair();
+        var rpc = new JsonRpc(streams.Item1, streams.Item2);
+        rpc.AddLocalRpcTarget(new Server(), new JsonRpcTargetOptions { UseSingleObjectParameterDeserialization = true });
+        rpc.StartListening();
+
+        Assert.Equal(3, await rpc.InvokeAsync<int>(nameof(IServer.InstanceMethodWithSingleObjectParameterButNoAttribute), new XAndYFields { x = 1, y = 2 }));
```
AArnott, line 1006: Calling `InvokeAsync` isn't using named arguments at all, so I don't think it's actually exercising your change. Does this test fail without your product changes (or when you don't set your new property to true on Line 1003)? I'm guessing it would pass.

You should call `InvokeWithParameterObjectAsync` here.

src/StreamJsonRpc/JsonRpc.cs
daniel-vetter: Are you sure we should check for the existence of the attribute (line 1438) and not the explicit name inside the attribute itself? If i add an attribute (with name = null) to a method, should it still be possible to call a method by omitting the async suffix?

src/StreamJsonRpc.Tests/JsonRpcTests.cs
daniel-vetter: fixed

src/StreamJsonRpc/JsonRpc.cs
AArnott: What would you check within the existing attribute itself? The property you added is just a `bool` property, so it will always be `false` unless it was set to `true` explicitly.
I think your current implementation that only sets the value to `true` if the attribute is entirely absent makes sense. Your original problem statement was that you wanted to avoid adding the attribute to every method, which is fair. But if the attribute is *already* on it, considering that method to be explicitly defined and not inherit defaults from elsewhere is a nice, simple model.

daniel-vetter: My original issue is solved. My point is not related to my original issue (but kind of to the solution of this bug).

Lets assume somebody defines this method:
```csharp
public async Task<string> DoSomethingAsync()
```
you can now call this method with an rpc call `DoSomething` and `DoSomethingAsync` (is this correct?)

Later, somebody adds an single object parameter like this:
```csharp
[JsonRpcMethod(UseSingleObjectParameterDeserialization = true)]
public async Task<string> DoSomethingAsync(SomeInputData input)
```
Since there is an attribute now, you can no longer call this method via `DoSomething` (even though the developer did change anything about the naming). Is this intended?

My solution would be to change line 1438 from this:
```csharp
if (attribute == null && method.Name.EndsWith(ImpliedMethodNameAsyncSuffix, StringComparison.Ordinal))
```
to this:
```csharp
if (attribute?.Name == null && method.Name.EndsWith(ImpliedMethodNameAsyncSuffix, StringComparison.Ordinal))
```

Let me know if i am understanding something wrong.

AArnott:
@daniel-vetter As there are 6 commits for this relatively small change, do you mind if I squash when completing this PR? Or would you like to squash your own commits and do a force push to your branch before I complete it?

daniel-vetter:
@AArnott I squashed everything. You can complete the PR.

PR #454, opened 2020-04-13
add richnav

noellelc:


azure-pipelines.yml
```diff
@@ -21,3 +21,6 @@ variables:
 
 jobs:
 - template: azure-pipelines/build.yml
+- template: azure-pipelines/richnav.yml
+  parameters:
+    windowsPool: Hosted Windows 2019 with VS2019
```
AArnott, line 26: Can we remove the parameter here since it only runs in one environment?

AArnott:
Can you squash your commits please?

azure-pipelines.yml
noellelc: Yes.

PR #459, opened 2020-04-24
Trace failures to report progress

AArnott:


src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -267,7 +269,12 @@ public JsonProgress(JsonRpc rpc, object token)
             /// <param name="value">The typed value that will be send in the notification to be reported by the original <see cref="IProgress{T}"/> instance.</param>
             public void Report(T value)
             {
-                this.rpc.NotifyAsync(ProgressRequestSpecialMethod, this.token, value).Forget();
+                this.rpc.NotifyAsync(ProgressRequestSpecialMethod, this.token, value).ContinueWith(
+                    (t, s) => ((JsonRpc)s).TraceSource.TraceEvent(System.Diagnostics.TraceEventType.Error, (int)JsonRpc.TraceEvents.ProgressNotificationError, "Failed to send progress update. {0}", t.Exception.InnerException ?? t.Exception),
```
milopezc, line 273: Doesn't the message "Failed to send progress update. {0}" need to be localized?

AArnott: Good question. But no. None of our trace messages are localizable at the moment. And this is by design, since we don't expect customers to look at these logs. They're for us, and we want to be able to read them no matter what language VS is running in when they are produced.

PR #467, opened 2020-05-20
Add JsonRpc.DispatchCompletion property

### LINKED ISSUE #466: Technique for waiting for work to complete?
Right now the `JsonRpc.Completion` task only continues on _disconnection_, not necessarily when all the tasks have been processed and completed (or even started).  Is there a means of knowing when the work is truly complete? On disconnection I'd like to wait to complete all the work before exiting / reporting back. 

I've tried multiple means of leveraging SynchronizationContexts but awaiting the `Completion` or a provided `SynchronizationContext` always continues on as soon as there's an await in the target method, and sometimes even before the target method even starts running.  Using a Single-threaded context proves to be problematic as well, as you can easily await the context after a message is read but _before_ it's actually dispatched to run, blocking it from doing anything.

What I actually want to achieve is routing RPC messages through IoT hub to Event Grid and then back down to the IoT device.  From the device side this works well - I've got a handler I can push received messages to and send messages with - but from the service side the function always exits before work is complete, which will cause problems down the road.

Maybe this just isn't the best choice of library for what I'd like to do right now, but it seems reasonable to need a want to waiting until all the work is complete.

AArnott:
This property returns a Task which completes whenever there are no local RPC methods currently running.
This allows shutdown code to not only wait for disconnection, but also for any trailing local RPC methods to finish running, if any.

Closes #466

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -1611,6 +1611,38 @@ public abstract class JsonRpcTests : TestBase
     }
 
     [Fact]
+    public async Task DispatchCompletion_RepeatedlyRefreshesDuringConnection()
+    {
+        for (int i = 0; i < 3; i++)
+        {
+            Assert.True(this.serverRpc.DispatchCompletion.IsCompleted);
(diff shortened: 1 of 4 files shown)


jlorich:
👍 Looks good from me end.  `await _jsonRpc.Completion` + `await _jsonRpc.DispatchCompletion` will let me safely hold the function until everything is wrapped up.

Thanks for the incredibly quick response on this!

Issue #471, opened 2020-05-29
If id property is missing in result or error message, code should fail gracefully


NielsUll:
I am interfacing with a device, which does not support `cancelRequest`. 
It sends the following invalid error reply - the id field is missing.

```json
   {"error": {"message": "'CommonRpcClient' object has no attribute '$/cancelRequest'", "data": 
     "'CommonRpcClient' object has no attribute '$/cancelRequest'", "code": -32000}, "jsonrpc": "2.0"}
```

This causes a null reference exception in ReadError:
https://github.com/microsoft/vs-streamjsonrpc/blob/7bba74ba8a9082f88acd2a1b78e361e4856dd705/src/StreamJsonRpc/JsonMessageFormatter.cs#L646

AArnott:
Your device is not conforming to JSON-RPC 2.0, which [forbids returning any response to a notification](https://www.jsonrpc.org/specification#notification) (emphasis added):

> A Notification is a Request object without an "id" member. A Request object that is a Notification signifies the Client's lack of interest in the corresponding Response object, and as such no Response object needs to be returned to the client. **The Server MUST NOT reply to a Notification**, including those that are within a batch request.

I get that you want to interop with it anyway, but please do report a bug to whoever is responsible for the software running on that device.

I'm also curious whether simply not using a `CancellationToken` when communicating with this device would be an adequate workaround for you.

AArnott:
Another bug in the device's implementation of JSON-RPC is its violation of [this requirement of all response objects](https://www.jsonrpc.org/specification#response_object):

> id
**This member is REQUIRED.**
It MUST be the same as the value of the id member in the Request Object.
**If there was an error in detecting the id in the Request object (e.g. Parse error/Invalid Request), it MUST be Null.**

When the device receives a message where it fails to parse the ID and wants to complain about it to the client, it should do so with an error message that _does_ contain a null `id` field. So that's another bug you can send the way of the device's developer.

Thanks for your report and PR by the way. I agree, StreamJsonRpc should not throw NRE in this case.

NielsUll:
Thanks – I will try that. I will also notify the implementors of the device.

PR #472, opened 2020-05-29
Fail gracefully if json-rpc 2.0 error or result is missing id

NielsUll:
Fix for issue #471


AArnott:
Looks like we have a problem with our CI/PR build. Let me fix that and get back to you.

PR #475, opened 2020-06-09
Allow roll-forward policy within the same feature band

kzu:
I think an improved dev flow is to allow newer SDK versions 
within the same feature band, which is typically a quite safe 
way to guarantee backwards compatibility while gaining 
important patches and minor features automatically if the 
dev box happens to have a newer version (i.e. I have 3.1.301
but it currently cannot be used with the existing policy in 
this repo).

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "sdk": {
     "version": "3.1.100",
-    "rollForward": "latestPatch",
+    "rollForward": "feature",
     "allowPrerelease": false
   }
 }


AArnott:
No. .NET SDK breaks builds too often. Twice this year so far at least.

.NET SDK breaks people even in patch releases.
Here's one recent one:
https://github.com/dotnet/sdk/issues/11274

There are others (that I didn't file but impacted me). It's easy to see a list of many of them:
https://github.com/dotnet/sdk/issues?q=is%3Aissue+is%3Aclosed+breaks+

Historically this has happened even before .NET Core. SDKs are just danged hard to both innovate as well as keep every past use working as it once did.

And even if there weren't breaks, new warnings from compilers or other toolsets will break the build because we have TreatWarningsAsErrors: true set in our pipeline. 

I'd like to encourage the SDK to innovate, even at the cost of sometimes breaking me or others. But to do that without causing fire drills for everyone who is impacted, global.json pinning of the SDK version is required so that the repo owners can set the schedule to update and test the new SDK for each repo.

kzu:
Makes sense. Oh well, on to installing another few SDKs to satisfy global.jsons :)

PR #480, opened 2020-06-23
Support streams as a return type

### LINKED ISSUE #424: Support returning a Stream from an RPC call
We are attempting to build an API that provides and accepts streams from a remote client. Sadly we can only use streams as parameters in the current implementation of streamjsonrpc. 

This means that any consumers of our service will have to create an intermediary stream which isn't user friendly. 
We can circumvent this by asking consumers to create a Duplex Pipe but it seems unnecessary to take that step if we can provide a stream from the API using streamjsonrpc.

javierdlg:
This PR enables the ability to return a stream as a server method to the client. 
The following corner cases have been added as tests:
- Support returning a stream from a server method
- Verify that a stream can be disposed after the client drops it
- Verify that custom return types with an inner stream also get a multiplexing channel
- Requesting a stream from a notification logs a warning and disposes of such stream immediately 

Closes #424

### DIFF
--- a/doc/oob_streams.md
+++ b/doc/oob_streams.md
@@ -47,14 +47,10 @@ The server may also reply with stream or pipe:
 ```cs
 public async Task<Stream> SendLargeFileAsync()
 {
-   var streamPair = FullDuplexPipe.CreatePair();
-   
-   // Use streamPair.Item1
+   // Open a stream on the server-side and forward it to the client.
+   var stream = new FileStream("pathToServerFile", FileMode.Open);
 
-   // Dispose of the stream to notify the client we are done sending data
-   streamPair.Item1.Dispose();
-
-   return streamPair.Item2;
+   return stream;
 }
 ```
 
--- a/doc/oob_streams.md
+++ b/doc/oob_streams.md
@@ -45,12 +45,9 @@ public async Task TakeLargeFileAsync(IDuplexPipe pipe)
 The server may also reply with stream or pipe:
 
 ```cs
-public async Task<Stream> SendLargeFileAsync()
-{
-   // Open a stream on the server-side and forward it to the client.
-   var stream = new FileStream("pathToServerFile", FileMode.Open);
-
-   return stream;
+public Stream GetFile(string path) {
+    // Validate that the client should be granted access to the requested file.
+    return new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, useAsync: true);
 }
 ```
 
--- a/doc/oob_streams.md
+++ b/doc/oob_streams.md
@@ -2,7 +2,7 @@
 
 JSON-RPC is great for invoking methods and passing regular data types as arguments.
 When you want to pass binary data or stream a great deal of text without encoding as a very large JSON message,
-StreamJsonRpc gives you an option to pass `Stream`, `IDuplexPipe`, `PipeReader` or `PipeWriter` as an argument to an RPC method.
+StreamJsonRpc gives you an option to pass `Stream`, `IDuplexPipe`, `PipeReader` or `PipeWriter` as an argument or as a return type for an RPC method.
 
 The content of the `Stream` or `IDuplexPipe` is transmitted out of band of the JSON-RPC channel so that no extra encoding is required. This out of band channel is provisioned from a [`MultiplexingStream`](https://github.com/AArnott/Nerdbank.Streams/blob/master/doc/MultiplexingStream.md) that can optionally be provided to the `JsonMessageFormatter` (or other formatters that support this feature). The `JsonRpc` connection itself is expected to be one of the channels in this `MultiplexingStream`.
 This can be configured like this (creation of the `MultiplexingStream` is out of scope of this topic):
@@ -42,14 +42,28 @@ public async Task TakeLargeFileAsync(IDuplexPipe pipe)
 }
 ```
 
+The server may also reply with stream or pipe:
+
+```cs
+public async Task<Stream> SendLargeFileAsync()
+{
+   var streamPair = FullDuplexPipe.CreatePair();
+   
+   // Use streamPair.Item1
+
+   // Dispose of the stream to notify the client we are done sending data
+   streamPair.Item1.Dispose();
+
+   return streamPair.Item2;
+}
+```
+
 ## Rules
 
 Passing out of band streams/pipes along JSON-RPC messages requires care be taken to avoid leaving
 abandoned `MultiplexingStream` channels active and consuming resources in corner cases.
 To facilitate this, the following rules apply:
-
-1. The `IDuplexPipe` always originates on the client and is passed as an argument to the server.
-   Servers are not allowed to return `IDuplexPipe` to clients because the server would have no feedback if the client dropped it, leaking resources.
+ 
 1. The client can only send an `IDuplexPipe` in a request (that expects a response).
    Notifications would not provide the client with feedback that the server dropped it, leaking resources.
 1. The client will immediately terminate the `IDuplexPipe` if the server returns ANY error in response to the request, since the server may not be aware of the `IDuplexPipe`.
--- a/src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
+++ b/src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
@@ -309,12 +309,13 @@ public abstract class DuplexPipeMarshalingTests : TestBase, IAsyncLifetime
     public async Task ClientCanWriteAndReadFromTwoWayStream()
     {
         var remoteStream = await this.clientRpc.InvokeAsync<Stream>(nameof(Server.ServerMethodThatWritesAndReadsFromTwoWayStream));
+        Assumes.Present(remoteStream);
 
         var readOnlyStream = new StreamReader(remoteStream);
         var writeOnlyStream = new StreamWriter(remoteStream);
 
         // Read server message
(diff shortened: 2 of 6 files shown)


src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
```diff
@@ -786,6 +832,39 @@ public Task AcceptPipeAndChatLater(bool writeOnOdd, IDuplexPipe pipe, Cancellati
...
+            streamPair.Item1.Flush();
+
+            return stream;
+        }
+
+        public Stream ServerMethodThatReturnsAndDisposesStream()
+        {
+            var stream = new SimplexStream();
+
+            this.StreamDisposedTask = Task.Run(() => stream.Dispose());
```
AArnott, line 864: This introduces a random timing where `stream.Dispose()` may occur anywhere from immediately (before the `return` statement) to arbitrarily later (or when this task is awaited). 
Which ordering do you intend to test? Let's rewrite the test to enforce one particular ordering for a more stable test.

src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
```diff
@@ -786,6 +832,39 @@ public Task AcceptPipeAndChatLater(bool writeOnOdd, IDuplexPipe pipe, Cancellati
         public Task RejectCall(IDuplexPipe pipe) => Task.FromException(new InvalidOperationException("Expected test exception."));
 
         public object ReturnPipeAsObject() => FullDuplexStream.CreatePipePair().Item1;
+
+        public Stream ServerMethodThatReturnsTwoWayStream()
+        {
+            var streamPair = FullDuplexStream.CreatePair();
```
AArnott, line 838: These two server methods aren't demonstrating 2-way communication with the client. I suggest you replace this with a `MemoryStream` which you simply initialize and set Position to 0 before returning it. That would resemble something closer to the typical use case of opening a `FileStream` and returning it.

Demonstrating a two way stream (and/or a stream that is one-way but one the client uses to write to instead of read from) would make for another interesting test. Not that it would exercise unique code but it demonstrates a use case that is possible.

AArnott: And speaking of documenting interesting use cases, please update the `oob_streams.md` doc to explain that `Stream` can be returned (I think the doc says it cannot be) and also demonstrates some uses for it.

src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
```diff
@@ -292,17 +294,59 @@ public async Task ClientCanSendWriteOnlyStreamToServer(bool orderedArguments)
...
+    /// </summary>
+    [Fact]
+    public async Task ClientCanRequestStreamAndDropIt()
+    {
+        byte[] buffer = new byte[32];
+        await this.clientRpc.InvokeAsync(nameof(Server.ServerMethodThatReturnsAndDisposesStream));
+
+        this.serverRpc.Dispose();
+
+        await WhenAllSucceedOrAnyFault(this.server.StreamDisposedTask!);
```
AArnott, line 335: I'm not sure what you're trying to show here. Of course this will complete because your server method scheduled its disposal before the client even had a chance to read from it.
I wonder if you were trying to test what I was talking about where closing the JsonRpc connection is supposed to dispose all active streams. I don't see a test here that covers that scenario. In that case you'd remove all explicit Stream disposals, but go ahead and dispose the client *or* server's `JsonRpc` object. Then verify that both ends of the `Stream` did in fact get disposed of.
The `JsonRpcClosesStreamAfterDisconnectedEvent` test looks like it may be helpful to at least look at here.

src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
```diff
@@ -786,6 +832,39 @@ public Task AcceptPipeAndChatLater(bool writeOnOdd, IDuplexPipe pipe, Cancellati
...
+
+        public Stream ServerMethodThatReturnsTwoWayStream()
+        {
+            var streamPair = FullDuplexStream.CreatePair();
+            var bytes = Encoding.UTF8.GetBytes("Streamed bits!");
+
+            streamPair.Item1.Write(bytes, 0, bytes.Length);
+            streamPair.Item1.Flush();
+
+            return streamPair.Item2;
```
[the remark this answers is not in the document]
javierdlg: Would disposing of one side of the FullDuplexPipe pair not also close the other side? Or does it wait until the other side reaches the end of the stream before disposing?

src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
javierdlg: This method intended to test returning a stream that the client drops, and it can still be disposed by the server. I'd say that it would be best for the disposal to happen, after we return from this method.
Could I instead save the task under `StreamDisposedTask` and then run it on the client side to guarantee that it runs after the server returns the stream?

src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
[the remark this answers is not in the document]
AArnott: No, disposing one stream does not dispose its associated Stream. It merely indicates that writing is done.
Streams do not have an `IsCompleted` property on them.

src/StreamJsonRpc.Tests/DuplexPipeMarshalingTests.cs
AArnott: Yes, that approach would work.

javierdlg: I opted for instead testing for stream disposal on RPC disconnect as you suggested in a comment above.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1870,6 +1870,17 @@ private async ValueTask<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcRequ
                             return CreateCancellationResponse(request);
                         }
 
+                        // Dispose of the result if it was requested via a notification and is a stream.
+                        if (!request.IsResponseExpected && result is Stream stream)
```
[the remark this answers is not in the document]
javierdlg: I looked into `MessageFormatterDuplexPipeTracker.cs` but I can't see where it checks for a notification, and from my tests it doesn't even reach the GetToken or GetPipe methods in that file. Did you mean another place?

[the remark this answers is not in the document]
javierdlg: Quick ping on this question.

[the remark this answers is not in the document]
AArnott: It's here, in the `GetToken` method:

```cs
            if (this.formatterState.SerializingMessageWithId.IsEmpty)
            {
                throw new NotSupportedException(Resources.MarshaledObjectInNotificationError);
            }
```

That already throws, so you'd never even reach this code that you're adding to `JsonRpc.cs`.
When I remove this code, all the tests still pass as well.

doc/oob_streams.md
```diff
@@ -42,14 +42,28 @@ public async Task TakeLargeFileAsync(IDuplexPipe pipe)
...
+public async Task<Stream> SendLargeFileAsync()
+{
+   var streamPair = FullDuplexPipe.CreatePair();
+   
+   // Use streamPair.Item1
+
+   // Dispose of the stream to notify the client we are done sending data
+   streamPair.Item1.Dispose();
+
+   return streamPair.Item2;
```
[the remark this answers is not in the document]
javierdlg: I have to agree with you there. I instead replaced this with a FileStream that then gets forwarded to the client since that will definitely be a more common case than a two-way communication.

doc/oob_streams.md
```diff
@@ -47,14 +47,10 @@ The server may also reply with stream or pipe:
 ```cs
 public async Task<Stream> SendLargeFileAsync()
 {
-   var streamPair = FullDuplexPipe.CreatePair();
-   
-   // Use streamPair.Item1
+   // Open a stream on the server-side and forward it to the client.
+   var stream = new FileStream("pathToServerFile", FileMode.Open);
```
AArnott, line 51: Since the stream is used for async I/O, it's a significant perf improvement to open the stream stating that async I/O will be used, [according to the docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.-ctor?view=netcore-3.1#System_IO_FileStream__ctor_System_String_System_IO_FileMode_System_IO_FileAccess_System_IO_FileShare_System_Int32_System_Boolean_):

> If the application is designed to take advantage of asynchronous I/O, set the useAsync parameter to true. Using asynchronous I/O correctly can speed up applications by as much as a factor of 10, but using it without redesigning the application for asynchronous I/O can decrease performance by as much as a factor of 10.

So I suggest we update the sample to:

```cs
public Stream GetFile(string path) {
    // Validate that the client should be granted access to the requested file.
    return new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 4096, useAsync: true);
}
```

I realize this is just a sample in a doc, but as it will likely serve as a reference to folks actually writing such code, let's show them how to do it in a performant way.

javierdlg: I very much agree, docs set a precedent for newcomers and the better it's written the better the average project that uses it will be. Fixed!

PR #484, opened 2020-07-02
LOC CHECKIN | Microsoft/vs-streamjsonrpc master | 20200702.

danyeh:
Translation of new string.


AArnott:
Thanks. I cherry-picked your commit to our v2.5 branch as 813fd29 since that's what we're stabilizing at this point.

PR #488, opened 2020-07-14
Add TransmissionCompletedSize ETW event

MokoSan:
Updated implementations of IJsonRpcMessageHandler to emit a new event including the size of payload. 

These implementations include those of: 
1. HeaderDelimitedMessageHandler.cs
2. LengthHeaderMessageHandler.cs

### DIFF
--- a/src/StreamJsonRpc/NewLineDelimitedMessageHandler.cs
+++ b/src/StreamJsonRpc/NewLineDelimitedMessageHandler.cs
@@ -113,11 +113,6 @@ namespace StreamJsonRpc
             cancellationToken.ThrowIfCancellationRequested();
             this.Formatter.Serialize(this.Writer, content);
             this.Writer.Write(this.newLineBytes.Span);
-
-            if (JsonRpcEventSource.Instance.IsEnabled(System.Diagnostics.Tracing.EventLevel.Informational, System.Diagnostics.Tracing.EventKeywords.None))
-            {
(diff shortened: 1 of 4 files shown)


src/StreamJsonRpc/NewLineDelimitedMessageHandler.cs
```diff
@@ -113,6 +113,11 @@ protected override void Write(JsonRpcMessage content, CancellationToken cancella
             cancellationToken.ThrowIfCancellationRequested();
             this.Formatter.Serialize(this.Writer, content);
             this.Writer.Write(this.newLineBytes.Span);
+
+            if (JsonRpcEventSource.Instance.IsEnabled(System.Diagnostics.Tracing.EventLevel.Informational, System.Diagnostics.Tracing.EventKeywords.None))
+            {
+                JsonRpcEventSource.Instance.TransmissionCompletedSize(this.newLineBytes.Length);
```
AArnott, line 119: This one will always raise an event with `1` or `2`. Unlike the other two handlers, this one does not buffer the content first. It just writes it straight out, so there's nothing tracking length of the message.
We *could* fix this (without introducing an otherwise unnecessary intermediate buffer) by creating an `IBufferWriter<byte>` wrapper that acts as a pass through to another one, but sums up all calls to `Advance(int)` that we could retrieve (and reset) at this point. 
But we aren't actually using this handler in VS at all, so consider simply reverting this change and calling it good. The other two handlers are the ones we used and your changes look good.

PR #498, opened 2020-07-16
Fix mis-reported cancellation of message transmission

### LINKED ISSUE #489: Race condition in InvokeWithCancellationAsync
Cancellation of `JsonRpc.InvokeWithCancellationAsync` might result in error RemoteProtocolViolation: A response was received without a request having been sent.

Related VS Feedback issue: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1156256

Repro:

1) Install VS 16.7.0 Preview 4.0 [30312.25.d16.7] (or any 16.7 P4 build before 7/14).
2) Set `ServiceHubTraceLevel=All`
3) Clone https://github.com/dotnet/roslyn and open Roslyn.sln
4) Make some changes, use IDE features such as Go To Definition, Find all References, etc.
5) After a while (a few minutes) VS displays gold-bar reporting JSON RPC exception and asking to restart Roslyn's ServiceHub process.
5) ServiceHub log `*-ManagedLanguage.IDE.RemoteHostClient-*.log` contains something like:

```
07/14/2020 10:52:29 Pacific Standard Time: Information : 6 :Received request "3943" for method "ReportTodoCommentDataAsync".

07/14/2020 10:52:29 Pacific Standard Time: Critical : 13 :Connection closing (RemoteProtocolViolation: A response was received without a request having been sent.). 

07/14/2020 10:52:29 Pacific Standard Time: Information : 9 :Invoking Microsoft.VisualStudio.LanguageServices.Implementation.TodoComments.VisualStudioTodoCommentsService.ReportTodoCommentDataAsync(DocumentId, ImmutableArray`1, CancellationToken)
07/14/2020 10:52:29 Pacific Standard Time: Information : 8 :{"id":3943}
07/14/2020 10:52:29 Pacific Standard Time: Verbose : 8 :Sent: {
  "jsonrpc": "2.0",
  "id": 3943,
  "result": null
}

07/14/2020 10:52:29 Pacific Standard Time: Error : 3 : [devenv:30604] Stream disconnected unexpectedly:  RemoteProtocolViolation, 'A response was received without a request having been sent.', LastMessage: , Exception: 

07/14/2020 10:52:29 Pacific Standard Time: Error : 18 :Exception thrown while transmitting message: System.OperationCanceledException: The operation was canceled.
   at System.Threading.CancellationToken.ThrowOperationCanceledException()
   at StreamJsonRpc.MessageHandlerBase.<WriteAsync>d__23.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at StreamJsonRpc.JsonRpc.<TransmitAsync>d__144.MoveNext()
```

A workaround has been implemented in Roslyn disabling the ability to cancel RPC invocations:
https://github.com/dotnet/roslyn/pull/45982

The issue has not occurred in a build with the workaround.

AArnott:
When a message is written to the output stream but not flushed, and then cancellation is requested, although the flush was canceled, the message was eventually going to be sent even though the client was no longer tracking it. This led to an unexpected response from the server to eventually be received and in v2.5, `JsonRpc` terminates the connection when the server sends an unexpected response to a request it doesn't remember sending.

The fix is to stop honoring the `CancellationToken` after the message has been written. We still honor the *disposal* token during flush so as to not hold up shutdown, but after a message is written, we always wait for flush to complete if we're still running.

Fixes #489

This was already merged to master via #492

### DIFF
--- a/src/StreamJsonRpc/MessageHandlerBase.cs
+++ b/src/StreamJsonRpc/MessageHandlerBase.cs
@@ -170,6 +170,7 @@ namespace StreamJsonRpc
         /// <returns>A task that represents the asynchronous operation.</returns>
         /// <exception cref="InvalidOperationException">Thrown when <see cref="CanWrite"/> returns <c>false</c>.</exception>
         /// <exception cref="OperationCanceledException">Thrown if <paramref name="cancellationToken"/> is canceled before message transmission begins.</exception>
+        /// <exception cref="ObjectDisposedException">Thrown if this instance is disposed before or during transmission.</exception>
         /// <remarks>
         /// Implementations should expect this method to be invoked concurrently
         /// and use a queue to preserve message order as they are transmitted one at a time.
@@ -189,7 +190,18 @@ namespace StreamJsonRpc
                     {
                         cancellationToken.ThrowIfCancellationRequested();
                         await this.WriteCoreAsync(content, cancellationToken).ConfigureAwait(false);
-                        await this.FlushAsync(cancellationToken).ConfigureAwait(false);
+
+                        // When flushing, do NOT honor the caller's CancellationToken since the writing is done
+                        // and we must not throw OperationCanceledException back at them as if we hadn't transmitted it.
+                        // But *do* cancel flushing if we're being disposed.
+                        try
+                        {
+                            await this.FlushAsync(this.DisposalToken).ConfigureAwait(false);
+                        }
+                        catch (OperationCanceledException ex) when (ex.CancellationToken == this.DisposalToken)
+                        {
+                            throw new ObjectDisposedException(this.GetType().FullName, ex);
+                        }
                     }
                     finally
                     {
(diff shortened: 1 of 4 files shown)


src/StreamJsonRpc/MessageHandlerBase.cs
```diff
@@ -189,7 +190,18 @@ public async ValueTask WriteAsync(JsonRpcMessage content, CancellationToken canc
...
-                        await this.FlushAsync(cancellationToken).ConfigureAwait(false);
+
+                        // When flushing, do NOT honor the caller's CancellationToken since the writing is done
+                        // and we must not throw OperationCanceledException back at them as if we hadn't transmitted it.
+                        // But *do* cancel flushing if we're being disposed.
+                        try
+                        {
+                            await this.FlushAsync(this.DisposalToken).ConfigureAwait(false);
+                        }
+                        catch (OperationCanceledException ex) when (ex.CancellationToken == this.DisposalToken)
```
sharwell, line 201: 💭 Seems like the more general option here would be:

```suggestion
                        catch (OperationCanceledException ex) when (this.DisposalToken.IsCancellationRequested)
```

PR #501, opened 2020-07-17
Rename TransmissionCompletedSize event and add received event

AArnott:
Previously we only reported on the size of messages *transmitted*. Now we report on the size of messages received as well.
I also give the events a different `Task` name to distinguish it from the `MessageTransmission` that is already used by `JsonRpc`.


AArnott:
CC: @MokoSan

PR #519, opened 2020-08-11
Add JsonRpc.AddLocalRpcTarget<T> method

### LINKED ISSUE #517: Missing JsonRpc method to expose only members of a given interface from a given RPC target object
Please provide an overload of attach that allows me to:
1.  Specify the object being attached to
2.  Provide the type of the object being attached to. 

The reason I want to specify the type is as follows:
Currently I have an Object that I want to expose. I only want to expose certain members of the object and those members are defined in an interface. The interface is fully compatible with Json RPC.  The implementation of my interface is not because it contains incompatible events.

I believe that if I could attach and specify a type, I could constrain the generated proxy to the members of the interface.

AArnott:
This method restricts RPC surface area to only members defined on the type T instead of all (public) members on the target object.

Closes #517

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -684,26 +684,37 @@ namespace StreamJsonRpc
             return proxy;
         }
 
-        /// <summary>
-        /// Adds the specified target as possible object to invoke when incoming messages are received.  The target object
-        /// should not inherit from each other and are invoked in the order which they are added.
-        /// </summary>
-        /// <param name="target">Target to invoke when incoming messages are received.</param>
+        /// <inheritdoc cref="AddLocalRpcTarget(object, JsonRpcTargetOptions?)"/>
         public void AddLocalRpcTarget(object target) => this.AddLocalRpcTarget(target, null);
 
+        /// <inheritdoc cref="AddLocalRpcTarget(Type, object, JsonRpcTargetOptions?)"/>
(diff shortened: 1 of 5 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1583,27 +1594,29 @@ protected virtual void OnRequestTransmissionAborted(JsonRpcRequest request)
...
+            var mapping = new MethodNameMap(exposedMembersOnType);
 
-            for (TypeInfo? t = target.GetType().GetTypeInfo(); t != null && t != typeof(object).GetTypeInfo(); t = t.BaseType?.GetTypeInfo())
+            for (TypeInfo? t = exposedMembersOnType; t != null && t != typeof(object).GetTypeInfo(); t = t.BaseType?.GetTypeInfo())
             {
                 // As we enumerate methods, skip accessor methods
                 foreach (MethodInfo method in t.DeclaredMethods.Where(m => !m.IsSpecialName))
                 {
-                    if (!options.AllowNonPublicInvocation && !method.IsPublic)
+                    if (!options.AllowNonPublicInvocation && !method.IsPublic && !exposedMembersOnType.IsInterface)
```
milopezc, line 1619: Do we need a check here for static methods too?

AArnott: No. We (oddly enough) *do* support static RPC methods on target objects.

PR #525, opened 2020-08-13
Switch default to to retain order of incoming RPC invocations

### LINKED ISSUE #521: JsonRpc should preserve order of invocations on RPC server by default
RPC servers today have to [take steps](https://github.com/microsoft/vs-streamjsonrpc/blob/d4a87766ef28f48e2744f3480a6587b440f80b82/doc/resiliency.md#when-message-order-is-important) to ensure that RPC server methods are invoked in the order the client sent the requests, because otherwise server methods are scheduled on the threadpool which does not guarantee order of execution.

Most folks seem to assume order is preserved by default, so doing so makes sense. Also coming up is #346 which adds an `OnNext` scenario where ordering is assumed and significant.

AArnott:
Closes #521

### DIFF
--- a/src/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/src/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -586,8 +586,16 @@ public abstract class JsonRpcTests : TestBase
     }
 
     [Fact]
-    public void SynchronizationContext_DefaultIsNull()
+    public void SynchronizationContext_DefaultIsNonConcurrent()
     {
+        Assert.IsType<NonConcurrentSynchronizationContext>(this.serverRpc.SynchronizationContext);
+    }
+
+    [Fact]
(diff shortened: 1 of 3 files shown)


src/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1772,8 +1780,11 @@ public async Task DispatchCompletion_DispatchesLingerAfterDisconnect()
     }
 
     [Fact]
-    public async Task MultipleSyncMethodsExecuteConcurrentlyOnServer()
+    public async Task MultipleSyncMethodsExecuteConcurrentlyOnServer_WithClearedSyncContext()
```
milopezc, line 1783: nit: Test for the default non concurrent behavior.

AArnott: Good idea. I'll add one.

PR #536, opened 2020-08-25
Add support for marshaling IObserver<T> over RPC

### LINKED ISSUE #346: Add support for passing IObserver<T> objects over RPC
We might define a protocol by which an `IObservable<T>` property's Subscribe method could be turned into an RPC call to start the remote subscription. We'd then need to define how callbacks from the server regarding that property would be sent.
Then the local JsonRpc instance can track the `IObserver<T>` object until the observable's OnError or OnComplete method is called.
Also consider supporting `IObserver<T>` being passed as an argument to an RPC method so that the server can take other args that configure exactly what data should be passed to that observer.

Prerequisites:

1. [x] #505 Add support for serializing exceptions (to support `IObserver<T>.OnError(Exception)`).
1. [x] #519 `AddLocalRpcTarget<T>` to expose an object, limited to one interface it implements.
1. [x] #521 JsonRpc orders messages by default (to support `IObserver<T>`.OnNext(T)`).
1. [x] #523 JsonRpc.NotifyAsync needs an overload with declared parameter types
1. [x] #522 Dynamic proxies allow `void` returning methods, implemented as notifications.
1. [x] #532 Add support for marshaling `IDisposable`

Tracked internally with https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1146152

AArnott:
Also:
- Fix proxies to accept multiple Dispose calls
- Centralize list of implicitly marshaled types across formatters

Closes #346


src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -17,6 +17,36 @@ namespace StreamJsonRpc.Reflection
...
+                            // When OnError or OnCompleted is called, per IObserver<T> patterns that's implicitly a termination of the connection.
+                            if (methodName == nameof(IObserver<int>.OnError) || methodName == nameof(IObserver<int>.OnCompleted))
+                            {
+                                ((IDisposable)sender).Dispose();
+                            }
+                        };
+                        proxy.CallingMethod += (sender, methodName) =>
+                        {
+                            // Any RPC method call on IObserver<T> shouldn't happen if it has already been completed.
+                            Verify.NotDisposed((IDisposableObservable)sender);
```
BertanAygun, line 43: Is the cast always valid here?

src/StreamJsonRpc.Tests/ObserverMarshalingTests.cs
```diff
@@ -0,0 +1,383 @@
...
+        return new WeakReference(observer);
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private async Task<WeakReference> PushCompleteAndReturn_Helper()
+    {
+        var observer = new MockObserver<int>();
+        await Task.Run(() => this.client.PushCompleteAndReturn(observer)).WithCancellation(this.TimeoutToken);
+        await observer.Completion.WithCancellation(this.TimeoutToken);
+        var result = await observer.Completion;
```
BertanAygun, line 249: Nit: Can we not do "var result = await observer.Completion.WithCancellation(..);"

src/StreamJsonRpc/IJsonRpcClientProxyInternal.cs
```diff
@@ -0,0 +1,32 @@
...
+        /// Exceptions thrown from the handler propagate directly to the caller of the proxy and prevent the RPC request from being transmitted.
+        /// </remarks>
+        event EventHandler<string> CallingMethod;
+
+        /// <summary>
+        /// Occurs after a method is invoke and transmission over RPC has been queued.
+        /// </summary>
+        /// <remarks>
+        /// The event arg is the CLR name of the method being invoked.
+        /// Exceptions thrown from the handler propagate directly to the caller of the proxy and prevent the RPC request from being transmitted.
```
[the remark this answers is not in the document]
AArnott: Right. Too much copy-paste. I'll fix.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
AArnott: Yes. The `sender` is a dynamically generated proxy and is *always* generated to implement this interface.

src/StreamJsonRpc.Tests/ObserverMarshalingTests.cs
AArnott: I assume it's the `var` in particular that you dislike. Yes, I dislike it too. I've fixed it.

PR #541, opened 2020-09-03
Add ICancellationStrategy extensibility point

### LINKED ISSUE #436: support new CancellationStrategy capability added to vscode-jsonrpc in StreamjsonRpc
recently new ability is added to vscode-jsonrpc to support custom cancellation mechanism to jsonrpc
(https://github.com/microsoft/vscode-languageserver-node/commit/b1f6a443efad3d61cf83344589099dab5318ca66)

it was added since the current built-in cancellation mechanism assumes consumers can either run concurrently or asynchronously. any synchronous code can't use the existing cancellation mechanism to cancel ongoing request since, at the time, it checks cancellation request, work is already all done.

some of lsp implementation on the node has synchronous lsp implementation (such as typescript, python) so they need different cancellation mechanisms. for example, typescript uses this trick where it creates a cancellation file/pipe and uses the existence of the file/pipe to indicate whether the current request needed to be canceled. python needs yet another way to implement it since they don't want to customize vscode client for cancellation support as typescript did.

to support these various needs and constraints of each LSP implementation, a way to provide its own cancellation mechanism is added to vscode-jsonrpc.

now, we would like to share the same lsp node server with VS but without this custom cancellation mechanism support, we lose a way to cancel any request in VS.

wonder whether it is possible to support this new ability in StreamJsonRpc as well.

AArnott:
This allows specific JsonRpc instances to be configured to implement `CancellationToken` support in a totally different way. They are *not* given the freedom to mutate the the request object itself, but they may trigger and respond to any other JSON-RPC message or do something entirely out of band (e.g. creating a file and/or testing for existing of the file).

As part of this change, we had to move `$/cancelRequest` handling out of its very special place in `JsonRpc.HandleRpcAsync` into a standard method handler. But that would have changed the method handler to be invoked on the user-configurable `SynchronizationContext` (which is now by default order-preserving). But this special method *shouldn't* be subject to that order-preserving `SynchronizationContext` because in fact we want these cancellation requests handled as soon as they come in so they can cancel an ongoing RPC server method that has not yet yielded. In order to support this, I modified how we record `TargetMethod` to include an optional override `SynchronizationContext`.

Closes #436

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -1866,10 +1866,20 @@ namespace StreamJsonRpc
                     }
 
                     // Arrange for sending a cancellation message if canceled while we're waiting for a response.
-                    using (cancellationToken.Register(this.cancelPendingOutboundRequestAction, request.RequestId, useSynchronizationContext: false))
+                    try
                     {
-                        // This task will be completed when the Response object comes back from the other end of the pipe
-                        return await tcs.Task.ConfigureAwait(false);
+                        using (cancellationToken.Register(this.cancelPendingOutboundRequestAction, request.RequestId, useSynchronizationContext: false))
+                        {
+                            // This task will be completed when the Response object comes back from the other end of the pipe
+                            return await tcs.Task.ConfigureAwait(false);
+                        }
+                    }
+                    finally
+                    {
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            this.CancellationStrategy?.OutboundRequestEnded(request.RequestId);
+                        }
                     }
                 }
             }
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -27,23 +27,22 @@ namespace StreamJsonRpc
     /// </summary>
     public class JsonRpc : IDisposableObservable, IJsonRpcFormatterCallbacks, IJsonRpcTracingCallbacks
     {
+        /// <summary>
+        /// The <see cref="System.Threading.SynchronizationContext"/> to use to schedule work on the threadpool.
+        /// </summary>
+        internal static readonly SynchronizationContext DefaultSynchronizationContext = new SynchronizationContext();
+
         private const string ImpliedMethodNameAsyncSuffix = "Async";
-        private const string CancelRequestSpecialMethod = "$/cancelRequest";
         private static readonly ReadOnlyDictionary<string, string> EmptyDictionary = new ReadOnlyDictionary<string, string>(new Dictionary<string, string>(StringComparer.Ordinal));
         private static readonly object[] EmptyObjectArray = new object[0];
         private static readonly JsonSerializer DefaultJsonSerializer = JsonSerializer.CreateDefault();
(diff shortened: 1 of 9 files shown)


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1620,7 +1637,7 @@ protected virtual void OnRequestTransmissionAborted(JsonRpcRequest request)
                     }
 
                     // Skip this method if its signature matches one from a derived type we have already scanned.
-                    MethodSignatureAndTarget methodTarget = new MethodSignatureAndTarget(method, target, attribute);
+                    MethodSignatureAndTarget methodTarget = new MethodSignatureAndTarget(method, target, attribute, null);
```
heejaechang, line 1640: nit: useSynchronizationContext: ?

AArnott: Done.

src/StreamJsonRpc/StandardCancellationStrategy.cs
```diff
@@ -0,0 +1,144 @@
...
+        public virtual void IncomingRequestEnded(RequestId requestId)
+        {
+            lock (this.inboundCancellationSources)
+            {
+                this.inboundCancellationSources.Remove(requestId);
+            }
+        }
+
+        /// <inheritdoc />
+        public virtual void CancelOutboundRequest(RequestId requestId)
```
heejaechang, line 77: so IncommingRequestStarted/Ended and AddLocalRpcMethod for CancelInboundRequestMethodInfo is all to handle incoming cancellation message and that seems all good.

for outgoing requests, it seems it only has CancelOutboundRequest but doesn't have a notification for when the outbound request it is canceling has returned. we need that hook to find safe point to remove some state (a file)

heejaechang: hmmm.. it looks like (from test and code), IncommingRequestStarted/Ended actually used for out going request as well. then, all good for us!

heejaechang (approved):
thank you!

milopezc (approved):

src/StreamJsonRpc/StandardCancellationStrategy.cs
AArnott: > it looks like (from test and code), IncommingRequestStarted/Ended actually used for out going request as well. then, all good for us!

No, it doesn't. IncomingRequestStarted/Ended only get invoked on the RPC server side. 

>  it seems it only has CancelOutboundRequest but doesn't have a notification for when the outbound request it is canceling has returned. we need that hook to find safe point to remove some state (a file)

This is the kind of review I was hoping to get from you, to make sure you had the invocations you need. I'll add another method so that the RPC client gets a chance to recognize when the request is fullfilled/canceled.

AArnott: I've pushed a change to add the method. But it occurs to me that a race condition may exist where if cancellation is raised at about the same time as the response is received, we may raise the two methods out of order, which would leave a file behind on disk. I'm thinking about how to mitigate that.

AArnott: OK, I added a test and carefully timed it with the debugger to ensure we're good here. The two outbound request-related methods cannot be executed out of order or concurrently, although the `OutboundRequestEnded` method *may* be invoked without an associated `CancelOutboundRequest`. I documented this on the interface and trust folks can deal with it in their implementations. We might be able to avoid that in `JsonRpc` but it would add complexity that I suspect isn't necessary, and it won't happen except in narrow race conditions.

heejaechang:
@huguesv for FYI

AArnott:
@heejaechang StreamJsonRpc insertions happen nightly.

Issue #576, opened 2020-10-07
Replace usage of Newtonsoft.Json's CamelCaseNamingStrategy.ResolvePropertyName with more performant implementation


tmat:
We see some allocations of `char[]` in Roslyn perf tests that would be good to remove:

https://github.com/JamesNK/Newtonsoft.Json/blob/666d9760719e5ec5b2a50046f7dbd6a1267c01c6/Src/Newtonsoft.Json/Utilities/StringUtils.cs#L162       

Name | Inc % | Inc
-- | -- | --
streamjsonrpc.ni!StreamJsonRpc.CommonMethodNameTransforms+<>c.<get_CamelCase>b__2_0(System.String) | 1.9 | 1,707,768
   newtonsoft.json.ni!CamelCaseNamingStrategy.ResolvePropertyName | 1.9 | 1,707,768
 newtonsoft.json.ni!StringUtils.ToCamelCase | 1.9 | 1,707,768
 mscorlib.ni!System.String.ToCharArray() | 1.1 | 964,360
   mscorlib.ni!System.String.CtorCharArray(Char[]) | 0.8 | 743,408

AArnott:
Maybe this is the time to copy newtonsoft's camelcase transform into streamjsonrpc. Besides allowing us to optimize this allocation, it may remove the only reason to load newtonsoft.json.dll in messagepack.dll scenarios.

PR #583, opened 2020-10-19
Avoid loading Newtonsoft.Json in MessagePack scenarios

AArnott:
The only dependency was our camelCase conversion.

This avoids an assembly load, which is independently valuable. But it is also a prereq to fixing #576.


AArnott:
Since PR builds don't trigger ([AzDO bug](https://developercommunity.visualstudio.com/content/problem/1213995/github-pr-triggers-does-not-work.html)), [here is the manually triggered PR build](https://dev.azure.com/azure-public/vside/_build/results?buildId=9882&view=results).

milopezc (approved):

PR #584, opened 2020-10-19
Recycle `char[]` in ToCamelCase

### LINKED ISSUE #576: Replace usage of Newtonsoft.Json's CamelCaseNamingStrategy.ResolvePropertyName with more performant implementation

AArnott:
Instead of calling `string.ToCharArray()` which allocates a new `char[]` each time it is invoked, we rent a `char[]` that is at _least_ as long as the string, and copy the string's contents into it. After creating a new string with the modified characters, we return the rented array for recycling.

This cuts down on GC pressure that showed up in Roslyn's perf tests.

Fixes #576

### DIFF
--- a/src/StreamJsonRpc/Utilities.cs
+++ b/src/StreamJsonRpc/Utilities.cs
@@ -92,38 +92,45 @@ namespace StreamJsonRpc
                 return identifier;
             }
 
-            char[] chars = identifier.ToCharArray();
-
-            for (int i = 0; i < chars.Length; i++)
+            char[] chars = ArrayPool<char>.Shared.Rent(identifier.Length);
+            identifier.CopyTo(0, chars, 0, identifier.Length);
+            try
             {
-                if (i == 1 && !char.IsUpper(chars[i]))
+                for (int i = 0; i < identifier.Length; i++)
                 {
-                    break;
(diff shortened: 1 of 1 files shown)


javierdlg (approved):

tmat (approved):
:shipit:

PR #594, opened 2020-10-28
Reduce string encoding/decoding in MessagePackFormatter

AArnott:
Several strings are *very* common in JSON-RPC. Rather than encode/decode them repeatedly, allocating new strings each time we decode, we instead use pre-encoded bytes and recognize the encoded bytes without decoding them.

The techniques used here come from MessagePack's own mpc-tool code output. So even though we're using MessagePack's "internal" namespace, it needs to be stable within a major version to keep mpc generated code working.
- [Sample deserialization decoder optimization](https://github.com/neuecc/MessagePack-CSharp/blob/896ed3c6bb3264bbda5c9a410a27d61b4cf1d93e/sandbox/Sandbox/Generated.cs#L1553-L1597)
- [Sample serialization encoder optimization](https://github.com/neuecc/MessagePack-CSharp/blob/896ed3c6bb3264bbda5c9a410a27d61b4cf1d93e/sandbox/Sandbox/Generated.cs#L1519-L1531)

I recognize that this compromises readability, but perf traces indicate that the decoding time and allocation costs are significant, so I believe it is justified. I define constants and add code comments to help explain it.

Also note (because I only learned this recently) that in C# the syntax below produces *very* efficient code. It never allocates an array -- ever. Not even once. [Read the IL yourself](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyDMABGgQMIEDeOB11y6AbAQEoCmAhjAPIB2EAngGUADm24AeAEZ9gLAHwEAaiwBOAZzAB7bgAVlGoSuB8AcmwC2LAELSWqggF553FgHcCUmQG0AuhQIMAdih/AAZ6YPR0AFYIyNiQ2IAWWNRggE40ggBfAG4cLKA=). This syntax leads the C# compiler to generate code that initializes the `ReadOnlySpan<byte>` to point directly into the binary of the loaded assembly itself to get at the bytes in the compiled binary.

```cs
static ReadOnlySpan<byte> VersionPropertyNameBytes => new byte[] { 167, 106, 115, 111, 110, 114, 112, 99 };
```

I did _not_ optimize every single formatter in the `MessagePackFormatter.cs` file. Just the hot path ones. We can optimize the error formatter if that becomes important.

### DIFF
--- a/src/StreamJsonRpc/MessagePackFormatter.cs
+++ b/src/StreamJsonRpc/MessagePackFormatter.cs
@@ -34,29 +34,39 @@ namespace StreamJsonRpc
     public class MessagePackFormatter : IJsonRpcMessageFormatter, IJsonRpcInstanceContainer, IJsonRpcFormatterState, IJsonRpcFormatterTracingCallbacks, IDisposable
     {
         /// <summary>
-        /// The string "jsonrpc" as it appears in UTF-8 encoding where the bytes are interpreted as a <see langword="ulong"/>.
+        /// The constant "jsonrpc", in its various forms.
         /// </summary>
-        private const ulong VersionPropertyNameULong = 27989659475735402UL;
+        private static readonly CommonString VersionPropertyName = new CommonString("jsonrpc");
 
         /// <summary>
-        /// The string "id" as it appears in UTF-8 encoding where the bytes are interpreted as a <see langword="ulong"/>.
+        /// The constant "id", in its various forms.
         /// </summary>
-        private const ulong IdPropertyNameULong = 25705UL;
+        private static readonly CommonString IdPropertyName = new CommonString("id");
 
         /// <summary>
-        /// The string "method" as it appears in UTF-8 encoding where the bytes are interpreted as a <see langword="ulong"/>.
+        /// The constant "method", in its various forms.
         /// </summary>
-        private const ulong MethodPropertyNameULong = 110429656606061UL;
+        private static readonly CommonString MethodPropertyName = new CommonString("method");
 
         /// <summary>
-        /// The string "result" as it appears in UTF-8 encoding where the bytes are interpreted as a <see langword="ulong"/>.
+        /// The constant "result", in its various forms.
         /// </summary>
-        private const ulong ResultPropertyNameULong = 128009175786866UL;
+        private static readonly CommonString ResultPropertyName = new CommonString("result");
 
         /// <summary>
-        /// The string "error" as it appears in UTF-8 encoding where the bytes are interpreted as a <see langword="ulong"/>.
+        /// The constant "error", in its various forms.
         /// </summary>
-        private const ulong ErrorPropertyNameULong = 491496043109UL;
+        private static readonly CommonString ErrorPropertyName = new CommonString("error");
+
+        /// <summary>
+        /// The constant "params", in its various forms.
+        /// </summary>
+        private static readonly CommonString ParamsPropertyName = new CommonString("params");
+
+        /// <summary>
+        /// The constant "2.0", in its various forms.
+        /// </summary>
+        private static readonly CommonString Version2 = new CommonString("2.0");
 
         /// <summary>
         /// A cache of property names to declared property types, indexed by their containing parameter object type.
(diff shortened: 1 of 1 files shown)


src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -1324,22 +1407,34 @@ public JsonRpcMessage Deserialize(ref MessagePackReader reader, MessagePackSeria
...
+                    ReadOnlySpan<byte> stringKey = MessagePack.Internal.CodeGenHelpers.ReadStringSpan(ref readAhead);
+                    switch (stringKey.Length)
                     {
-                        return options.Resolver.GetFormatterWithVerify<Protocol.JsonRpcResult>().Deserialize(ref reader, options);
-                    }
-                    else if (propertyName == ErrorPropertyName)
-                    {
-                        return options.Resolver.GetFormatterWithVerify<Protocol.JsonRpcError>().Deserialize(ref reader, options);
-                    }
+                        case 6:
```
BertanAygun, line 1414: Since we are not using length to get the ulong from GetKey method do we need to separate out cases by length? Is there any harm in calling GetKey if the length is >6?

src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -1398,53 +1501,75 @@ public Protocol.JsonRpcRequest Deserialize(ref MessagePackReader reader, Message
                 int propertyCount = reader.ReadMapHeader();
                 for (int propertyIndex = 0; propertyIndex < propertyCount; propertyIndex++)
                 {
-                    switch (reader.ReadString())
+                    // We read the property name in this fancy way in order to avoid paying to decode and allocate a string when we already know what we're looking for.
```
BertanAygun, line 1504: Same question regarding switch statement by length here as well

src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -1387,7 +1485,12 @@ internal JsonRpcRequestFormatter(MessagePackFormatter formatter)
                 this.formatter = formatter;
             }
 
-            public Protocol.JsonRpcRequest Deserialize(ref MessagePackReader reader, MessagePackSerializerOptions options)
+            /// <summary>
+            /// Gets the raw msgpack encoding of the string "params".
+            /// </summary>
+            private static ReadOnlySpan<byte> ParamsPropertyNameBytes => new byte[] { 166, 112, 97, 114, 97, 109, 115 };
```
BertanAygun, line 1491: I wonder if there would be value in moving all these fields to a static singleton that calculates these values via UTF8 encoder as a one time init cost. It would also make adding new strings in future easier.

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: `GetKey` supports any non-zero length, but it can only return a fully representative `ulong` for lengths of 8 or less.
We check length first because there's no reason to call `GetKey` at all if the length doesn't match one that we expect.

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: It's an interesting question. I certainly could initialize these at runtime. I'll leave this comment active to mull it over.

There is a method in MessagePack that can return these arrays (`CodeGenHelpers.GetEncodedStringBytes`). But I used https://kawanet.github.io/msgpack-lite/ which was even easier.
The harder values to get was the expected `ulong` result of `GetKey`. For that I had to write a little code to actually call the function with various strings to see what it would return. Alternatively folks can run mpc.exe on their project to get all this written out for them automatically, but StreamJsonRpc is a complex enough case I didn't want to try it. 😛

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: Same answer. :)

src/StreamJsonRpc/MessagePackFormatter.cs
BertanAygun: At the very least I think we should check in that method to get ulong as a private utility method for future use.

BertanAygun (approved):
Approving based on comments

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: I pushed what amounts to a significant rewrite for readability and maintainability based on your feedback, @bertanaygun. Would you care to re-review?

BertanAygun: I had already approved before but these changes also look good to me. I agree that it is much easier to understand now and the details about string lenghts etc don't interfere with the serialization logic.

PR #595, opened 2020-10-29
Short-lived JsonRpc connection perf improvements

AArnott:
These are a few simple improvements to reduce GC pressure and improve perf when setting up a JsonRpc connection, as guided by ETW traces.


BertanAygun (approved):
All looks good to me.

Issue #575, opened 2020-10-07
Optimize allocations in AddLocalRpcMethod


tmat:
We see some allocations of `int[]` in Roslyn perf tests that would be good to remove:

Name | Inc % | Inc
-- | -- | --
`streamjsonrpc.ni!JsonRpc.AddLocalRpcMethod` | 3.0 | 2,680,744
` clr!JIT_New` | 1.2 | 1,068,888
`mscorlib.ni!System.Collections.Generic.Dictionary2[System.__Canon,System.__Canon].Insert(System.__Canon,   System.__Canon, Boolean)` | 1.1 | 965,736
`streamjsonrpc.ni!JsonRpc.TraceLocalMethodAdded` | 0.7 | 646,120

Link to internal RPS results:
http://aka.ms/BRAT?Form=BRAT&System=DTL&Locale=English&Runs=%25PerfDDRITs%25&Branch=main&Parent=main&FIBuild=30605.164&RIBuild=30605.164.vslsnap.201006.015741.279356&Private=False&GetResults=True&Since=10-01-2020

AArnott:
Fixed by several recent PRs. Please test  2.7.37 or later when verifying this.

PR #596, opened 2020-10-29
Optimize RPC target reflection

AArnott:
Cache the results of reflection based on the type and options so subsequent connections can be made faster and with less GC pressure.
Also move all the RPC target handling from JsonRpc to its own class.

This makes a short-lived "use once" JsonRpc connection 18% faster and cuts out some large allocations.

Related to #575

### DIFF
--- a/src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
+++ b/src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
@@ -1,0 +1,684 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace StreamJsonRpc.Reflection
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
+    using System.Reflection;
+    using System.Runtime.ExceptionServices;
+    using System.Text;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft;
+    using Microsoft.VisualStudio.Threading;
+    using StreamJsonRpc.Protocol;
+
+    internal class RpcTargetInfo : System.IAsyncDisposable
+    {
+        private const string ImpliedMethodNameAsyncSuffix = "Async";
+        private static readonly Dictionary<TypeInfo, MethodNameMap> MethodNameMaps = new Dictionary<TypeInfo, MethodNameMap>();
+        private static readonly Dictionary<(TypeInfo Type, bool AllowNonPublicInvocation, bool UseSingleObjectParameterDeserialization), Dictionary<string, List<MethodSignature>>> RequestMethodToClrMethodMap = new Dictionary<(TypeInfo Type, bool AllowNonPublicInvocation, bool UseSingleObjectParameterDeserialization), Dictionary<string, List<MethodSignature>>>();
+        private readonly JsonRpc jsonRpc;
+
+        /// <summary>
+        /// A collection of target objects and their map of clr method to <see cref="JsonRpcMethodAttribute"/> values.
+        /// </summary>
+        private readonly Dictionary<string, List<MethodSignatureAndTarget>> targetRequestMethodToClrMethodMap = new Dictionary<string, List<MethodSignatureAndTarget>>(StringComparer.Ordinal);
+
+        /// <summary>
+        /// A list of event handlers we've registered on target objects that define events. May be <c>null</c> if there are no handlers.
+        /// </summary>
+        private List<EventReceiver>? eventReceivers;
+
+        /// <summary>
+        /// A lazily-initialized list of objects to dispose of when the JSON-RPC connection drops.
+        /// </summary>
+        private List<object>? localTargetObjectsToDispose;
+
+        internal RpcTargetInfo(JsonRpc jsonRpc)
+        {
+            this.jsonRpc = jsonRpc;
+        }
+
+        private TraceSource TraceSource => this.jsonRpc.TraceSource;
+
+        private object SyncObject => this.targetRequestMethodToClrMethodMap;
+
+        public async ValueTask DisposeAsync()
+        {
+            if (this.localTargetObjectsToDispose is object)
+            {
+                List<Exception>? exceptions = null;
+                foreach (object target in this.localTargetObjectsToDispose)
+                {
+                    // We're calling Dispose on the target objects, so switch to the user-supplied SyncContext for those target objects.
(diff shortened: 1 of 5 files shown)


src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -0,0 +1,685 @@
...
+
+                    throw new AggregateException(exceptions);
+                }
+            }
+        }
+
+        internal static MethodNameMap GetMethodNameMap(TypeInfo type)
+        {
+            MethodNameMap? map;
+            lock (MethodNameMaps)
```
BertanAygun, line 106: I assume this means gains by reduced GC pressure outweighs the overhead of the lock?

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -0,0 +1,685 @@
...
+        /// Gets the <see cref="JsonRpcMethodAttribute"/> for a previously discovered RPC method, if there is one.
+        /// </summary>
+        /// <param name="methodName">The name of the method for which the attribute is sought.</param>
+        /// <param name="parameters">
+        /// The list of parameters found on the method, as they may be given to <see cref="JsonRpcRequest.TryGetTypedArguments(ReadOnlySpan{ParameterInfo}, Span{object})"/>.
+        /// Note this list may omit some special parameters such as a trailing <see cref="CancellationToken"/>.
+        /// </param>
+        internal JsonRpcMethodAttribute? GetJsonRpcMethodAttribute(string methodName, ReadOnlySpan<ParameterInfo> parameters)
+        {
+            if (this.targetRequestMethodToClrMethodMap.TryGetValue(methodName, out List<MethodSignatureAndTarget>? existingList))
```
BertanAygun, line 139: Do we need locks here or can a local rpc target object be added only before calling these?

BertanAygun (approved):
I assumed most of the new code in RpcTargetInfo.cs is actually moved from JsonRpc.cs and didn't review it thoroughly. Let me know if there is a section that's new code.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -0,0 +1,685 @@
...
+    using System.Threading.Tasks;
+    using Microsoft;
+    using Microsoft.VisualStudio.Threading;
+    using StreamJsonRpc.Protocol;
+
+    internal class RpcTargetInfo : System.IAsyncDisposable
+    {
+        private const string ImpliedMethodNameAsyncSuffix = "Async";
+        private static readonly Dictionary<TypeInfo, MethodNameMap> MethodNameMaps = new Dictionary<TypeInfo, MethodNameMap>();
+        private static readonly Dictionary<(TypeInfo Type, bool AllowNonPublicInvocation, bool UseSingleObjectParameterDeserialization), Dictionary<string, List<MethodSignature>>> RequestMethodToClrMethodMap = new Dictionary<(TypeInfo Type, bool AllowNonPublicInvocation, bool UseSingleObjectParameterDeserialization), Dictionary<string, List<MethodSignature>>>();
```
tmat, line 25: >Dictionary<(TypeInfo Type, bool AllowNonPublicInvocation, bool UseSingleObjectParameterDeserialization), Dictionary<string, List<MethodSignature>>> [](start = 214, length = 147)

`new()`?

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -0,0 +1,685 @@
...
+        private readonly JsonRpc jsonRpc;
+
+        /// <summary>
+        /// A collection of target objects and their map of clr method to <see cref="JsonRpcMethodAttribute"/> values.
+        /// </summary>
+        private readonly Dictionary<string, List<MethodSignatureAndTarget>> targetRequestMethodToClrMethodMap = new Dictionary<string, List<MethodSignatureAndTarget>>(StringComparer.Ordinal);
```
tmat, line 31: >Dictionary<string, List<MethodSignatureAndTarget>> [](start = 116, length = 50)

`new(StringComparer.Ordinal)`

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -0,0 +1,685 @@
...
+                            disposableTarget.Dispose();
+                        }
+                    }
+#pragma warning disable CA1031 // Do not catch general exception types
+                    catch (Exception ex)
+#pragma warning restore CA1031 // Do not catch general exception types
+                    {
+                        if (exceptions is null)
+                        {
+                            exceptions = new List<Exception>();
```
tmat, line 84: >exceptions [](start = 28, length = 10)

Can be simplified to `exceptions ??= new List<Exception>();`

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
AArnott: Oh yes, very much so. 18% faster, remember. 😏
And ya, the reflection that we skip here is expensive in terms of GC pressure. Almost every API we call creates _arrays_, even if we just wanted one object. Even `GetCustomAttribute` calls another API that returns an array before returning a single element from it.

AArnott: Also note, these are all optimistic locking. We take it just long enough to peek, then we release the lock to do the work. As a result, this lock is not held for very long at all => minimal lock contention.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
AArnott: Oooh. good call. Accessing this dictionary without a lock would be dangerous. I'm adding the lock now.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
AArnott: Love it. Just as soon as C# 9 hits GA. 😉

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
AArnott: When C# 9 is GA, certainly. :)

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
AArnott: And it works in C# 8. Nice. :)

drewnoakes: Same in https://github.com/microsoft/vs-streamjsonrpc/pull/598/files#diff-917aff0ba4f129621f1035d2159950ffe99d4ad6f160676ca183a72efb554122R784-R789

Issue #468, opened 2020-05-21
Serialize exceptions with ISerializable pattern for error messages


AArnott:
We should consider serializing exceptions with `Exception.GetObjectData` and deserializing using the associated constructor. We would pass in a `SerializationInfo` object that we control rather than use .NET's binary formatter.
If we did this, we'd likely still throw `RemoteInvocationException` as we do now, but could supply the deserialized exception as the `InnerException`.

If we do this, our deserializer should follow all the rules described [here](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.iserializable?view=netcore-3.1).

Tracked internally by https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1128351

kzu:
Might also be worth exploring whether it would be useful to also serialize [Exception.Data](https://docs.microsoft.com/en-us/dotnet/api/system.exception.data?view=netcore-3.1) if present. Might not be used a lot, but in these remove scenarios, it could provide a mechanism to propagate custom data without having to create custom exception types and enabling the corresponding serialization for each...

AArnott:
We recently used the Data dictionary ourselves in order to avoid a dependency on a new exception type.

kzu:
ok, hacked-up spike looks promising and shows that this could indeed be quite valuable for the .NET clients: https://github.com/kzu/spikes/tree/master/jsonrpc-exceptions

AArnott:
FWIW this should even be doable without changing StreamJsonRpc at all, since the methods involved are virtual.

kzu:
Absolutely. But that would mean that services would *need* to inherit from some `JsonRpc`-derived base class that could give this to them for free. But maybe that's fine?

AArnott:
Having it built-in is certainly an option. I wanted to call it out though that our extensibility model allows for such functionality without actually changing the library.

AArnott:
Let's work #502 into this.

AArnott:
After #505 this is pretty trivial. I have a prototype that I'm working into a shippable state.

PR #602, opened 2020-11-03
Improve handling of various non-serializable Exception situations

### LINKED ISSUE #588: Exceptions without [Serializable] fail to serialize/deserialize properly
In general we have several issues around serialization/deserialization of `Exception`-derived types that do not conform to the `ISerializable`/`[Serializable]` pattern when using `ExceptionProcessing.ISerializable`.

AArnott:
There were a few issues:
1. We didn't gracefully handle exception types that were missing the `[Serializable]` attribute. Now we log a warning and just write out `CommonErrorData`.
1. We didn't gracefully handle exception types that were missing the deserializing constructor. Now we log a warning and deserialize the nearest base type that _does_ conform to the serializable pattern.
1. We weren't even serializing *thrown* exceptions to JSON ourselves (newtonsoft.json was doing it) which means it wasn't treated as 'user data'. Now we serialize thrown exceptions with our own converter as user data.
1. Exceptions weren't serialized to include their assembly name, so unless the assembly was already loaded, we wouldn't be able to activate it. Now we artificially add `AssemblyName` as a property of the exception.
1. Even if exception's assembly were already loaded, unless it was loaded in the `From` load context it still wouldn't be found. That's still the case with this change, but if the assembly name is specified we'll load it into the `From` context automatically, assuming the CLR can find it.
1. Even if the assembly was loaded in the `From` context, if the assembly version of the transmitting side exceeded the assembly version on the receiving side, it wouldn't load properly. Now we fallback to allow an assembly _version_ mismatch (provided the name and public key token still match). We further fallback to any (loaded) assembly that defines a type with a matching full name.

Fixes #588 
See also DevDiv-1240018, #468

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -575,36 +575,41 @@ namespace StreamJsonRpc
             try
             {
                 this.serializingMessageWithId = jsonRpcMessage is IJsonRpcMessageWithId msgWithId ? msgWithId.RequestId : default;
-                if (jsonRpcMessage is Protocol.JsonRpcRequest request)
+                switch (jsonRpcMessage)
                 {
-                    this.serializingRequest = true;
+                    case Protocol.JsonRpcRequest request:
+                        this.serializingRequest = true;
 
-                    if (request.ArgumentsList != null)
-                    {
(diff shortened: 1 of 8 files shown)


AArnott:
See also DevDiv-1240018

ankitvarmait (approved):

PR #603, opened 2020-11-03
Add support for activity tracing

### LINKED ISSUE #470: Adding support for end-to-end activity tracing
When StreamJsonRpc is used to power microservices-based architectures, the need for rich (end-to-end) tracing and activity propagation, to enable better troubleshooting and monitoring becomes quite important. This has been an explicit focus in [.NET/ASP.NET Core 3.0](https://devblogs.microsoft.com/aspnet/improvements-in-net-core-3-0-for-troubleshooting-and-monitoring-distributed-apps/) and with the brand-new [W3C Trace Context recommendation](https://www.w3.org/TR/trace-context/) there's an opportunity to introduce this in StreamJsonRpc in a standards-compliant way. 

The goal of a built-in activity tracing support would be to enable rich tracing that can correlate calls across multiple remote invocations. This could then be used, for example, in System.Diagnostics.TraceSource for TraceTransfer/Start/Stop which model activities, or [EventSource.SetCurrentThreadActivityId](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventsource.setcurrentthreadactivityid?view=netcore-3.1) to do [activity-based logging](http://www.shujaat.net/2013/11/transfer-events-activity-based-logging.html) in ETW traces, or the newer [Activity](https://jimmybogard.com/building-end-to-end-diagnostics-and-tracing-a-primer-trace-context/) in .NET Core that has built-in support for the W3C standard.

Currently, there doesn't seem to be any specific support for propagating activity identifiers across RPC calls (just a request identifier). One option would be to leverage the fact that the request id does not need to have a specific format and therefore could accomodate a proper W3C trace context.

Attempting to add this support from outside of StreamJsonRpc (in a spike) turned out to be very tricky/hacky and not very satisfactory, which is why I think it's best if this was supported built-in.

This support would allow tools that can properly correlate and visualize traces to provide improved diagnostics like the Service Trace Viewer in this sample screenshot from a [spike](https://dcazzu.visualstudio.com/vsx/_git/vsx?path=%2Ftracing%2FReadme.md&_a=preview): 

image

AArnott:
This adds `traceparent` and `tracestate` properties to JSON-RPC request messages to carry an activity across the wire for end-to-end activity tracing.

This allows a set of structured log files to be aggregated with a viewer like Microsoft's Service Trace Viewer to produce a cross process/machine view like this (where the green boxes are from the RPC client and the yellow boxes are from the RPC server):

image

There are a variety of other views this tool offers. The `ActivityTracing_IntegrationTest` test produces the log files that led to the above screenshot.

Closes #470


src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -1706,6 +1766,73 @@ public void Serialize(ref MessagePackWriter writer, Protocol.JsonRpcRequest valu
...
+            private static unsafe string ReadTraceState(ref MessagePackReader reader)
+            {
+                int elements = reader.ReadArrayHeader();
+                if (elements % 2 != 0)
+                {
+                    throw new NotSupportedException("Odd number of elements not expected.");
+                }
+
+                // With care, we could probably assemble this string with just two allocations (the string + a char[]).
+                var resultBuilder = new StringBuilder();
```
kelltrick, line 1822: I wonder if there is a cheap way to give the StringBuilder the know final size of the string so it never needs to resize it's internal buffer.

test/StreamJsonRpc.Tests/ActivityTracingStrategyTests.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+        finally
+        {
+            Activity.Current.Stop();
+        }
+    }
+
+    [Theory]
+    [InlineData(null)]
+    [InlineData("k=v")]
```
kelltrick, line 47: nit: It probably makes sense to add a `[InlineData(k=v;k2=v2)]` to all of the tracestate tests here.

kelltrick (approved):
It all looks good, I left a few nits. The only thing I feel strongly about is that we should allow uppercase inbound hex simply because A. the standard says implementers must accept upper case hex and B. the reverse hex decoding array you have defined already to accommodates uppercase letters.

doc/tracecontext.md
```diff
@@ -0,0 +1,125 @@
...
+  "traceparent": "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01",
+  "tracestate": "rojo=00f067aa0ba902b7,congo=t61rcWkgMzE"
+}
+```
+
+### Binary encoding
+
+When using a binary encoding (e.g. MessagePack) the trace-context values are encoded as arrays of binary elements as follows:
+
+- `traceparent` as `[uint8, [byte[], byte[], byte]]`
```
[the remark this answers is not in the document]
AArnott: I've changed `byte` to `uint8` above.
The inner array costs 1 byte. So ya, it's just a convenient wrapper so that depending on the version, the serialization process can engage another formatter/deformatter that can expect to deserialize/serialize an entire 'object'.

src/StreamJsonRpc/IActivityTracingStrategy.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// Applies an activity described in an incoming RPC request to the current context so the dispatched method can inherit it.
+        /// </summary>
+        /// <param name="request">The inbound RPC request.</param>
+        /// <returns>An optional disposable object that can revert the effects taken by this method at the conclusion of the dispatched RPC server method.</returns>
+        /// <remarks>
+        /// This method may be invoked regardless of whether an activity is identified in the incoming request.
+        /// </remarks>
+        IDisposable? ApplyInboundActivity(JsonRpcRequest request);
```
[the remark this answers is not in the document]
AArnott: We don't reuse `JsonRpcRequest` objects, and various extensions already retain these objects for the duration of a request. This is no different. In fact one of our two implementations do just that. I'm adding an xml doc comment indicating that doing so is safe.

src/StreamJsonRpc/ActivityTracingStrategy.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        private class State : IDisposable
+        {
+            internal State(Activity newActivity)
+            {
+                this.PriorActivity = Activity.Current;
+                this.NewActivity = newActivity;
+            }
+
+            internal Activity PriorActivity { get; }
```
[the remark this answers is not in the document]
AArnott: I use `readonly` fields everywhere. But we do have the analyzer rules enabled that want to see fields be `private`, and these can't be.
But FWIW, I've seen the .NET team folks many times deny that auto-properties are any slower than fields for any application. The JIT will inline cheap methods, including this property accessor, so it will actually result in the same assembly code either way. Check [this out](https://sharplab.io/#v2:C4LghgzgtgNAJiA1AHwAICYCMBYAUKgZgAIMiBhIgbzyNpIBYiAxAexYAoBKKmuvgNzAAnIhGBhgAUyIBeIgDtJAdyIBlcVPaYYmTgG5efWgEt5wIgDNZojZIB0F45IA2cA7iN1T5gA7WxEvYACkIsPpJCwACe7nwAvoZEiT5CxoJSJOhqtjwenkTeEfJgztmB7N6WTq4wBWZEKWER0dzUefl0wAAWxhAO1XDWji5uiR3dvXYhTZFR1o3hs7H5Ce1GY15mRSV15sOuy56FQsWlldOL0VQA5pLAeqvxeHFAA=). Try commenting out the property retrieval then try commenting out the field retrieval. Both remove exactly 1 assembly instruction. So they're exactly equivalent. That's in Debug build. In release builds, it all disappears because we don't use the values.

src/StreamJsonRpc/CorrelationManagerTracingStrategy.cs
```diff
@@ -0,0 +1,131 @@
...
+            Requires.NotNull(request, nameof(request));
+
+            if (Trace.CorrelationManager.ActivityId != Guid.Empty)
+            {
+                var traceparent = default(TraceParent);
+
+                FillRandomBytes(new Span<byte>(traceparent.ParentId, TraceParent.ParentIdByteCount));
+                CopyGuidToBuffer(Trace.CorrelationManager.ActivityId, new Span<byte>(traceparent.TraceId, TraceParent.TraceIdByteCount));
+
+                if (this.TraceSource is object && (this.TraceSource.Switch.Level & SourceLevels.ActivityTracing) == SourceLevels.ActivityTracing && this.TraceSource.Listeners.Count > 0)
```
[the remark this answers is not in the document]
AArnott: `x is object` is exactly equivalent in generated code to `x != null` except that `x is object` won't compile when `x` is a `struct` whereas `x != null` will compile and allow the misperception that the check is useful to continue. So the C# team's recommendation is to use `is object` instead of `!= null` in all instances. By adopting this pattern myself, I've caught several bugs in my own code and others'.
Similarly `x is null` is preferred over `x == null` for all the same reasons.

`HasFlag` is easier on the eyes, there's no doubt. But it *boxes* the enum. While that might seem small, in code that's hit for every message (like this will be) that cost adds up quickly. So quickly, in fact, that it was the no. 1 allocated object in a perf trace I just ran and had to [fix it](https://github.com/microsoft/vs-streamjsonrpc/commit/db7f844ad32ca5cc236252d715187b0c2ad5c356).

src/StreamJsonRpc/CorrelationManagerTracingStrategy.cs
```diff
@@ -0,0 +1,131 @@
...
+            internal ActivityState(JsonRpcRequest request, TraceSource? traceSource, string? activityName, Guid parentTraceId, Guid childTraceId)
+            {
+                this.originalActivityId = Trace.CorrelationManager.ActivityId;
+                this.originalTraceState = TraceState;
+                this.activityName = activityName;
+                this.parentTraceId = parentTraceId;
+
+                if (traceSource is object)
+                {
+                    Trace.CorrelationManager.ActivityId = parentTraceId;
```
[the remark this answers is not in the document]
AArnott: Typically, yes. But not this time, because we immediately set `ActivityId` to something *else* on the very next line. So the only reason we set it here is *for* the `Transfer` call.
I'll add a code comment to indicate this.

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: There's a way, but it ain't cheap in terms of coding. per my code comment above, we could fully predict the length sufficient to allocate just `string` and one `char[]`. But the code is so complex that it seemed to add unnecessary risk. If/when we ever actually use `tracestate` (and I doubt we ever will) then we can certainly revisit this.

test/StreamJsonRpc.Tests/ActivityTracingStrategyTests.cs
AArnott: Good idea. I added it here and to `CorrelationManagerTracingStrategyTests` as well. I used `,` as the delimiter since the spec uses that rather than `;`

But neither of these test *serialization* of the value. And since each formatter is responsible for serialization, we should test both of those. I'll add tests to `JsonRpcTests` to add `tracestate` and verify it's intact on the other side.

AArnott: ...and it turns out I had a bug in serializing tracestate in messagepack that the new tests found. So thank you very much for your comment.

src/StreamJsonRpc/Hex.cs
```diff
@@ -0,0 +1,64 @@
...
+            var bytes = new byte['f' - '0' + 1];
+
+            for (int i = 0; i < 10; i++)
+            {
+                bytes[i] = (byte)i;
+            }
+
+            for (int i = 10; i < 16; i++)
+            {
+                bytes[i + 'a' - '0' - 0x0a] = (byte)i;
```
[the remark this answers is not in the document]
AArnott: Thanks for the code snippet. It would have taken me much longer to figure that out as I never knew how this code worked. :)

src/StreamJsonRpc/ActivityTracingStrategy.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (request.TraceParent is object)
+            {
+                var state = new State(new Activity(request.Method));
+                state.NewActivity.TraceStateString = request.TraceState;
+                state.NewActivity.SetParentId(request.TraceParent);
+                state.NewActivity.Start();
+                return state;
+            }
+
+            return null;
```
ankitvarmait, line 43: In the case inbound activity's trace parent is not an object, should we consider to start a new activity and set the parent to null/ emty? so for all JsonRpcRequest calls from here will have the parentId set to this new activity?

ankitvarmait (approved):

src/StreamJsonRpc/ActivityTracingStrategy.cs
AArnott: Good idea. That improves the diagnosability on the server-side when the client hasn't gone to any trouble to set a traceparent. I've made the change.

PR #619, opened 2020-11-18
Add extensibility point to customize deserialized exception types

AArnott:


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -1173,6 +1173,54 @@ internal void AddLocalRpcMethod(MethodInfo handler, object? target, JsonRpcMetho
...
+                        }
+                        catch (FileLoadException)
+                        {
+                            // If we fail again, we'll just try to load the exception type from the AppDomain without an assembly's context.
+                        }
+                    }
+                }
+            }
+
+            Type? runtimeType = typeDeclaringAssembly is object ? typeDeclaringAssembly.GetType(typeFullName) : Type.GetType(typeFullName);
```
ankitvarmait, line 1220: Should we recheck arguments :typeFullName for ArgumentNullException? Or we are sure here that, this would be checked by callers of this method or it would contain a valid value always.

ankitvarmait (approved):

src/StreamJsonRpc/JsonRpc.cs
AArnott: Good point, as a publicly exposed method (even though we don't expect external callers) we should validate our inputs.

PR #622, opened 2020-12-09
Simplified nuget feeds and removed unused package reference

javierdlg:


AArnott (commented):
We can't do this. All our OSS repos must retain multiple nuget feeds until Azure Artifacts supports upstreams from *unauthenticated* feeds.

Issue #630, opened 2021-01-12
CancellationToken no longer works on v2.6.121 when setting TypeNameHandling.Objects


rruizGit:
### Use Case:

- Using dynamic proxies.
- Using Cancellation Tokens.

A simple method implemented in the server as part of a unit test:

```C#
        public Task CancellableAsync(int secondsToDelay, CancellationToken cancellationToken)
        {
            return Task.Run(() =>
            {
                for (int i = 0; i < secondsToDelay; i++)
                {
                    Thread.Sleep(1000);
                    cancellationToken.ThrowIfCancellationRequested();
                }
            }, cancellationToken);
        }
```
Previous to v2.6.121, if we called this method through a Dynamic Proxy and provided a CancellationToken, cancelling the operation would cancel the token in this server method and the exception would be thrown.  With v2.6.121 the token is never cancelled in the server. I can see the cancellation message going throught the RPC. Not sure if this is a bug or if there is something new we have to do with the addition of ICancellationStrategy.

AArnott:
This should work. No breaking change was intended. Can you send a full repro and/or include verbose logs from both parties in the minimal repro so I can investigate?

rruizGit:
Andrew, 

Sorry it took me so long. Turns out it was not so easy to reproduce. I did eventually figure out what in our setup was causing the failure. In short, the failure only occurs in v2.6.121 **if** you are using the JsonMessageFormatter to format the JsonSerializer. We format ours this way:

```C#
    var serverFormatter = new JsonMessageFormatter(new UTF8Encoding(encoderShouldEmitUTF8Identifier: false));
    serverFormatter.JsonSerializer.TypeNameHandling = TypeNameHandling.Objects;
    serverFormatter.JsonSerializer.TypeNameAssemblyFormatHandling = TypeNameAssemblyFormatHandling.Simple;
    var serverRpc = new JsonRpc(new HeaderDelimitedMessageHandler(serverStream, serverFormatter));
```

If _TypeNameHandling_ is set to _Objects_ or _All_ the CancellationToken workflow fails, but only on v2.6.121. If I revert to the previous release it works fine. When it fails, we see this in the log:

_Invocation of "$/cancelRequest" cannot occur because arguments do not match any registered target methods._

I have created a small VS project which demonstrates the problem. It can be found [here](https://1drv.ms/u/s!AoUSEZvDd2WAiJ1metS0r2pe0A6qFg?e=LJ01M5).

[...]

PR #651, opened 2021-03-29
Ignore `$type` property in named arguments in incoming JSON requests

rruizGit:
Fix for issue #630: _CancellationToken no longer works on v2.6.121 when setting TypeNameHandling.Objects_. 

When enabling TypeNameHandling.Objects for the JsonSerializer, invocations with simple object dictionaries would end with the dictionary's type in the "params". 

```json
{
  "jsonrpc": "2.0",
  "method": "$/cancelRequest",
  "params": {
    "$type": "System.Collections.Generic.Dictionary`2[[System.String, mscorlib],[System.Object, mscorlib]], mscorlib",
    "id": 2
  }
}
```
With this type of message, the _JsonMessageFormatter_ would convert the properties into a _IReadOnlyDictionary<string, object>_. 
```c#
return args.Properties().ToDictionary(p => p.Name, p => (object)p.Value);
```
Unfortunately, this would treat the "type" JToken as a separate parameter which would then cause later failures because the number of parameters would be incorrect. The proposed solution is to first allow the JToken to convert itself to the right type of dictionary which uses and filters out the "type" and then convert it to the _IReadOnlyDictionary<string, object>_.
```C#
return args.ToObject<Dictionary<string, JToken>>().ToDictionary(k => k.Key, k => (object)k.Value);
```
This solves all the issues discussed in #630.


AArnott (changes requested):
A surprisingly simple solution. Thanks for sharing.

However we lack two parts to the fix:
1. We are sending invalid JSON-RPC requests still by transmitting `$type` as a property. We need to suppress that.
2. We need a regression test for this.

Are you interested in pursuing these items?

rruizGit:
Andrew, sorry for delay, I was on vacation. I can look into your request. I did think about "filtering out" the type handling for the parameters dictionary while I was working on this solution. But I came across a few problems with that. I will look into it again and get back to you.

First questions related to this topic, 
- Suppressing type handling when serializing the "parameters dictionary" also means that no custom type handling is allowed when invoking methods that way. Is that Ok? Or do you mean suppress the $type only for the dictionary but allow it for the parameters in the dictionary? When using dynamic interface proxies this is not an issue because it employs a totally different serialization/deserialization code path. 
- In term of unit tests, I don't think the codebase actually has any test with custom type handling. Is that your assessment as well? If so, are you asking me to create unit tests around these scenarios?

AArnott:
> Or do you mean suppress the $type only for the dictionary but allow it for the parameters in the dictionary

Generally the goal is to control what the protocol dictates, and allow the app to do anything within that boundary. So in this case that means the named parameters object has one property for every parameter, and nothing more. But within each argument value, the app can do whatever it wants including having `$type` properties.

> In term of unit tests, I don't think the codebase actually has any test with custom type handling. Is that your assessment as well? If so, are you asking me to create unit tests around these scenarios?

You are probably correct about the test gap. And yes, I'm suggesting we need to fill that gap and then verify that the client never emits a `$type` property as if there were a `$type` parameter on the method. This particular test should probably go in JsonRpcClient20InteropTests where we have full insight as an arbitrary server to inspect the message.

rruizGit:
@AArnott, here is another attempt at the fix. It preserves compatibility with the Json RPC specification. I also added unit tests mostly by inheriting from some of the existing unit tests and re-executing them with TypeHandling set to "Objects". If you take the fix out you will notice that a lot of those unit tests fail. I ran the fix through our own unit tests which are heavily dependent on TypeHandling and it seems to work well. 

Please take a look. Also, I'm thinking it might be worth my time to add some true TypeHandling dependent unit tests were interfaces and sub-types are being serialized/deserialized. Thoughts?

AArnott (approved):
Thank you for the fix to the product and tests.

I revised the product fix so we were less vulnerable to issues like this in the future, and to add a missing feature (declared types) that was lacking anyway.

PR #652, opened 2021-03-29
AddLocalRpcTarget<T> should honor base interfaces when T is an interface

### LINKED ISSUE #637: Usage mismatch between AddTarget<T> and Attach<T> when using interface inheritance.
Note that this is not necessarily a bug but pointing out a disconnect between AddTarget and Attach when using types (e.g. dynamic proxies). The ability to limit publishing to a specified type was added in v2.6.121.

If I define two interfaces, where one inherits the other:

```c#
    public interface IParentInterface : IChildInterface
    {
        Task<bool> MethodInParent();
    }

    public interface IChildInterface
    {
        Task<int> MethodInChild();
    }
```
I then implement the interfaces with a test class.
```C#
    public class TestClass : IParentInterface
    {
        public Task<bool> MethodInParent() => Task.Run(() => true);

        public Task<int> MethodInChild() => Task.Run(() => 42);
    }
```
I then publish the the test class object using the interface as the "exposingMembersOn" type:
```C#
       serverRpc.AddLocalRpcTarget<IParentInterface>(testClass, null);
```
And extract a dynamic proxy on the client side like this:
```C#
       IParentInterface proxy = clientRpc.Attach<IParentInterface>();
```
The dynamic proxy will include the method _MethodInChild()_ but will throw a _RemoteMethodNotFoundException_. This is because the publishing did not automatically publish the child interface, while the dynamic proxy implemented the parent and child interfaces. 

There is a workaround where you also publish the child:
```C#
      serverRpc.AddLocalRpcTarget<IParentInterface>(testClass, null);
      serverRpc.AddLocalRpcTarget<IChildInterface>(testClass, null);
```
and it works, but it seems like a disconnect in behavior. Personally, I think if I publish an interface, it should also publish any interfaces it inherits from, since that is what you get with the dynamic proxy. Or, possibly add a new option to _JsonRpcTargetOptions_ to trigger this behavior.

rruizGit:
The fix is to basically "collect" types to map differently if the `exposedMembersOnType` is an interface or a class. 

Fixes #637: Usage mismatch between AddTarget<T> and Attach<T> when using interface inheritance.

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -2119,11 +2119,9 @@ namespace StreamJsonRpc
                 // If t is just a Task, there is no Result property on it.
                 // We can't really write direct code to deal with Task<T>, since we have no idea of T in this context, so we simply use reflection to
                 // read the result at runtime.
-#pragma warning disable VSTHRD002 // misfiring analyzer https://github.com/Microsoft/vs-threading/issues/60
-#pragma warning disable VSTHRD102 // misfiring analyzer https://github.com/Microsoft/vs-threading/issues/60
+#pragma warning disable VSTHRD103 // misfiring analyzer https://github.com/Microsoft/vs-threading/issues/60
                 const string ResultPropertyName = nameof(Task<int>.Result);
-#pragma warning restore VSTHRD002
-#pragma warning restore VSTHRD102
+#pragma warning restore VSTHRD103
 
                 PropertyInfo? resultProperty = taskOfTTypeInfo.GetDeclaredProperty(ResultPropertyName);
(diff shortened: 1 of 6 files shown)


AArnott (approved):
Thanks for this. I have a couple of touch-ups to make on this as well as adding a regression test. I'll merge this together and complete the PR soon.

AArnott (approved):

AArnott:
@rruizGit  can you try again to sign the CLA?

AArnott:
This didn't end up making the cut for v2.8. I've reverted the merge in v2.8 (8cd2c4a5b790) and reapplied for v2.9 (in main, ccb5c03b64531).

PR #659, opened 2021-04-19
Adding automated compliance tools to the build pipeline.

matteo-prosperi:
Adding automated compliance tools to the build pipeline.


azure-pipelines/secure-development-tools.yml
```diff
@@ -1,5 +1,14 @@
 steps:
 
+ ### Scan binaries with Windows Defender
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
+  displayName: 'Scan with Windows Defender'
+  inputs:
+    TreatSignatureUpdateFailureAs: Warning
+    SignatureFreshness: OneDay
+    FileDirPath: '$(Build.SourcesDirectory)/bin/'
+  continueOnError: True
```
AArnott, line 10: Why scan and only emit a warning on failure? Those will typically go unnoticed. Is it because of a high false positive rate?

azure-pipelines/secure-development-tools.yml
```diff
@@ -1,5 +1,14 @@
 steps:
 
+ ### Scan binaries with Windows Defender
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
+  displayName: 'Scan with Windows Defender'
+  inputs:
+    TreatSignatureUpdateFailureAs: Warning
```
AArnott, line 7: If we're going to get this all the time, can we not treat it as a warning either? I really want our pipelines to usually "pass" without showing a warning indicator all the time.

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
   inputs:
     targetType: F
+
+ ### Analyze managed and unmanaged binaries (exe, dll) for security vulnerabilities
+- task: BinSkim@3
+  displayName: Run BinSkim
+  inputs:
+    InputType: 'Basic'
+    Function: 'analyze'
+    AnalyzeTarget: '$(Build.SourcesDirectory)/bin/StreamJsonRpc/$(BuildConfiguration)/netstandard2.0/StreamJsonRpc.dll;$(Build.SourcesDirectory)/bin/StreamJsonRpc/$(BuildConfiguration)/netcoreapp2.1/StreamJsonRpc.dll'
```
AArnott, line 28: I want to avoid hard-coded paths if possible, and especially hidden hard-coded lists of our TFMs. For instance when we add a target for `net5.0`, we'll never think to add it to this list. Instead, we can find an existing script or add a new one to the `azure-pipelines/variables` directory that searches for and returns all the matching binaries, and then just use `$(ScriptName)` here.

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
+  condition: ne(variables['DisableAPIScan'], 'true')
+
+- task: CopyFiles@2
+  displayName: 'Copy Files for APIScan'
+  inputs:
+    SourceFolder: '$(Build.SourcesDirectory)/bin/StreamJsonRpc/$(BuildConfiguration)'
+    Contents: |
+      netstandard2.0\StreamJsonRpc.*
+    TargetFolder: $(Agent.TempDirectory)\APIScanFiles\netstandard2.0
+  condition: ne(variables['DisableAPIScan'], 'true')
```
AArnott, line 54: Again, this is another discrete list of TFMs that is likely to fall out of date without our realizing it. Is there any reason we can't use `**/StreamJsonRpc.*` as the `Contents` parameter on just one `CopyFiles` task instead?

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
+
+ ### Run APIScan
+- task: APIScan@1
+  displayName: Run APIScan
+  inputs:
+    softwareFolder: $(Agent.TempDirectory)\APIScanFiles
+    softwareName: 'Microsoft.VisualStudio.LanguageServer'
```
AArnott, line 61: copy/paste error?

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
+    softwareVersionNum: '$(Build.BuildId)'
```
AArnott, line 62: I wonder if `$(Build.BuildNumber)` would produce a better experience wherever this report ends up being displayed. But if it *must*  be unique, keep what you have because `BuildNumber` is usually unique, but not guaranteed.

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
+    isLargeApp: false
+    toolVersion: LatestPreRelease
```
AArnott, line 64: This `toolVersion` seems most likely to be unstable. Can we use a latest stable version or even better just *pin* the version so that it doesn't suddenly start failing when they push an update?

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
+    InputType: 'Basic'
+    Function: 'analyze'
+    AnalyzeTarget: '$(Build.SourcesDirectory)/bin/StreamJsonRpc/$(BuildConfiguration)/netstandard2.0/StreamJsonRpc.dll;$(Build.SourcesDirectory)/bin/StreamJsonRpc/$(BuildConfiguration)/netcoreapp2.1/StreamJsonRpc.dll'
+
+ ### Copy existing Roslyn Analyzer logs to be analyzed
+- task: RoslynAnalyzers@3
+  displayName: Run Roslyn Analyzers.
+  inputs:
+    copyLogsOnly: 'true'
+  continueOnError: True
```
AArnott, line 35: We already run the analyzers as part of compilation. Why run them here again? That just seems like it would slow us down and double up on the warnings.

azure-pipelines/secure-development-tools.yml
```diff
@@ -9,3 +18,87 @@ steps:
...
+    ModernCop: false
+    PoliCheck: true
+    PoliCheckBreakOn: 'Severity4Above'
+    RoslynAnalyzers: true
+    RoslynAnalyzersBreakOn: 'WarningAbove'
+    SDLNativeRules: false
+    Semmle: false
+    TSLint: false
+    ToolLogsNotFoundAction: 'Standard'
+  condition: eq(variables['BreakOnError'], 'true')
```
AArnott, line 104: This looks like the default will be that failures will not break the pipeline. Can we instead make sure we pass cleanly, and then shut the door to silent regressions by ensuring any block the pipeline immediately?

src/StreamJsonRpc/StreamJsonRpc.csproj
```diff
@@ -3,8 +3,10 @@
     <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
AArnott, line 6: We do *not* want this here, as that significantly slows down the dev's inner loop, which has no reason to block on compiler warnings. We set this in the pipeline itself so that warnings cannot be introduced to the shared repo, but are allowed on local dev boxes.

src/StreamJsonRpc/StreamJsonRpc.csproj
```diff
@@ -3,8 +3,10 @@
...
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-
+    <ErrorLog>$(MSBuildProjectFullPath).$([System.Guid]::NewGuid().ToString().Substring(0,8)).sarif</ErrorLog>
```
AArnott, line 8: What reads this in order to produce the files?
Can we place it in the `$(IntermediateOutputPath)` instead of in the source directory to preserve our read-only source tree?
If this should be applied to all non-test projects, can we promote it to the `src/Directory.Build.props` file?

src/StreamJsonRpc/StreamJsonRpc.csproj
```diff
@@ -33,6 +35,10 @@
     <PackageReference Include="System.Threading.Tasks.Dataflow" Version="5.0.0" PrivateAssets="compile" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
     <PackageReference Include="XliffTasks" Version="1.0.0-beta.19505.1" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
```
AArnott, line 41: This package is already brought in by our `Directory.Build.props` file.

AArnott (changes requested):

azure-pipelines/secure-development-tools.yml
matteo-prosperi: The available options are: Error, Standard, Warning. Standard is the same as Warning.

azure-pipelines/secure-development-tools.yml
matteo-prosperi: This task when using `copyLogsOnly: 'true'` simply copies the Roslyn Analyzer results so that they can be used by `PostAnalysis@1`. This does not run the Roslyn Analyzer again (that would require a full new build which would risk overwriting the assemblies).

azure-pipelines/secure-development-tools.yml
matteo-prosperi: yes, I will fixt it

azure-pipelines/secure-development-tools.yml
AArnott: Are we required to run this task? If it's going to cause every build result to be a warning, I'm not happy. I'd want to file a bug against them at the _very_ least.

azure-pipelines/secure-development-tools.yml
AArnott: Why do we need the logs copied? If there were any warnings our build would have broken earlier in the pipeline.

src/StreamJsonRpc/StreamJsonRpc.csproj
matteo-prosperi: Cool. I will remove it.

src/StreamJsonRpc/StreamJsonRpc.csproj
matteo-prosperi:
```
 ### Copy existing Roslyn Analyzer logs to be analyzed
- task: RoslynAnalyzers@3
  displayName: Run Roslyn Analyzers.
  inputs:
    copyLogsOnly: 'true'
  continueOnError: True
```
at line 31 will read this file.

I will try moving the file and see if it works but the docs for the `RoslynAnalyzers` task specify that this is what should be used.

src/StreamJsonRpc/StreamJsonRpc.csproj
matteo-prosperi: Ok, I will remove it.

azure-pipelines/secure-development-tools.yml
matteo-prosperi: In my experience with reporting to the team who owns this tools, it would be better if this were a unique build number.

azure-pipelines/secure-development-tools.yml
matteo-prosperi: I will remove this.

azure-pipelines/secure-development-tools.yml
matteo-prosperi: I meant to configure BreakOnError = true as the default configuration of the pipeline and leave setting it to false as an option, but I can remove this.

azure-pipelines/secure-development-tools.yml
```diff
@@ -46,15 +39,14 @@ steps:
...
+- task: APIScan@2
   displayName: Run APIScan
   inputs:
     softwareFolder: $(Agent.TempDirectory)\APIScanFiles
     softwareName: 'StreamJsonRpc'
     softwareVersionNum: '$(Build.BuildId)'
     isLargeApp: false
-    toolVersion: LatestPreRelease
-  continueOnError: true
+    toolVersion: '3.1.29231'
```
AArnott, line 49: Where are the versions published? Adding a comment here with a link may help us keep it current over time.

azure-pipelines/secure-development-tools.yml
matteo-prosperi: Removing in favor of the existing Microbuild Signing task that covers the requirement

azure-pipelines/secure-development-tools.yml
matteo-prosperi: Agreed. Removed

src/StreamJsonRpc/StreamJsonRpc.csproj
matteo-prosperi: Removed as we rely on the build breaking to avoid Roslyn Analyzers issues

.gitignore
```diff
@@ -250,3 +250,6 @@ paket-files/
 msbuild.err
 msbuild.wrn
 *.binlog
+
+# Roslyn analyzer output
+*.sarif
```
AArnott, line 255: Do we need this change any more?

src/StreamJsonRpc/StreamJsonRpc.csproj
```diff
@@ -4,7 +4,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-
+
```
AArnott, line 7: Revert the whitespace change

AArnott (changes requested):

src/StreamJsonRpc/StreamJsonRpc.csproj
matteo-prosperi: Done already in the latest iteration

.gitignore
matteo-prosperi: I am reverting this.

azure-pipelines/secure-development-tools.yml
matteo-prosperi: Done

azure-pipelines/secure-development-tools.yml
matteo-prosperi: Done, I added a script in `azure-pipelines/variables`

AArnott (approved):

azure-pipelines/secure-development-tools.yml
matteo-prosperi: I added it

AArnott (approved):
I pushed one change to skip compliance tasks for optprof runs. 

Please squash, or rebase to clean up the commits before merging.

PR #666, opened 2021-05-21
Omit line info handling

AArnott:
@davkean reports this costs us 20MB of memory.


davkean (approved):

javierdlg (approved):

davkean:
This is from a trace during solution open of 468 projects:

image

168 projects looks about the same, so looks like a fixed overhead.

PR #686, opened 2021-08-06
Render named arguments in ETW events, including when using the MessagePack formatter

amcasey:
Presently, `JsonRpcEventSource.GetArgumentsString` prints dictionaries as ``System.Collections.Generic.Dictionary`2[System.String,System.Object]``, making `StreamJsonRpc/InboundCall` events unhelpful. Since `Dictionary<string, object>` seems like a case that would be generally useful when stringifying JSON objects, add special handling to `GetArgumentsString`.

### DIFF
--- a/src/StreamJsonRpc/JsonRpcEventSource.cs
+++ b/src/StreamJsonRpc/JsonRpcEventSource.cs
@@ -282,7 +282,7 @@ namespace StreamJsonRpc
             {
                 return string.Empty;
             }
-            else if (arguments is object[] args)
+            else if (arguments is object?[] args)
             {
                 var stringBuilder = new StringBuilder();
                 for (int i = 0; i < args.Length; ++i)
@@ -296,10 +296,10 @@ namespace StreamJsonRpc
 
                 return stringBuilder.ToString();
             }
-            else if (arguments is Dictionary<string, object> dict)
+            else if (arguments is IReadOnlyDictionary<string, object?> dict)
             {
                 var stringBuilder = new StringBuilder();
-                foreach (KeyValuePair<string, object> entry in dict)
+                foreach (KeyValuePair<string, object?> entry in dict)
                 {
                     stringBuilder.Append($"{entry.Key}: {entry.Value}, ");
                     if (stringBuilder.Length > maxLength)
--- a/src/StreamJsonRpc/JsonRpcEventSource.cs
+++ b/src/StreamJsonRpc/JsonRpcEventSource.cs
@@ -278,18 +278,33 @@ namespace StreamJsonRpc
 
             const int maxLength = 128;
 
-            if (arguments is object?[] args)
+            if (arguments is null)
+            {
+                return string.Empty;
+            }
+            else if (arguments is object[] args)
             {
                 var stringBuilder = new StringBuilder();
-                if (args != null && args.Length > 0)
+                for (int i = 0; i < args.Length; ++i)
                 {
-                    for (int i = 0; i < args.Length; ++i)
+                    stringBuilder.Append($"arg{i}: {args[i]}, ");
+                    if (stringBuilder.Length > maxLength)
                     {
-                        stringBuilder.Append($"arg{i}: {args[i]}, ");
-                        if (stringBuilder.Length > maxLength)
-                        {
-                            return $"{stringBuilder.ToString(0, maxLength)}...(truncated)";
-                        }
+                        return $"{stringBuilder.ToString(0, maxLength)}...(truncated)";
+                    }
+                }
+
+                return stringBuilder.ToString();
+            }
+            else if (arguments is Dictionary<string, object> dict)
+            {
+                var stringBuilder = new StringBuilder();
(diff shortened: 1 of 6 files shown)


amcasey:
I couldn't figure out a way to handle other dictionary key types (other than reflection), but I'm open to suggestions.

amcasey:
This is what I'm seeing when I record a trace using the VS feedback tool:

image

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -278,18 +278,33 @@ internal static string GetArgumentsString(object? arguments)
 
             const int maxLength = 128;
 
-            if (arguments is object?[] args)
+            if (arguments is null)
+            {
+                return string.Empty;
+            }
+            else if (arguments is object[] args)
```
AArnott, line 285: It looks like you inadvertently dropped the `?` in `object?[]` here.

src/StreamJsonRpc/JsonRpcEventSource.cs
```diff
@@ -278,18 +278,33 @@ internal static string GetArgumentsString(object? arguments)
...
+                    stringBuilder.Append($"arg{i}: {args[i]}, ");
+                    if (stringBuilder.Length > maxLength)
+                    {
+                        return $"{stringBuilder.ToString(0, maxLength)}...(truncated)";
+                    }
+                }
+
+                return stringBuilder.ToString();
+            }
+            else if (arguments is Dictionary<string, object> dict)
```
AArnott, line 299: By using the type `IReadOnlyDictionary<string, object?>` I think you'll match more, and recognize that the values may be null.

AArnott (changes requested):
Thanks for this submission.

From your screenshot, was that before or after the changes in this PR?

src/StreamJsonRpc/JsonRpcEventSource.cs
amcasey: Oops

src/StreamJsonRpc/JsonRpcEventSource.cs
amcasey: Done

amcasey:
My screenshot is from before my change.  TBH, I wasn't sure how to get an after screenshot - clobber the the DLL installed by VS?  The code works in isolation, so I'd just been testing it in a console app on the side.

src/StreamJsonRpc/JsonRpcEventSource.cs
amcasey: I did belatedly realize I could use `IDictionary.Keys` to do the iteration.  It's not as tidy, but it would catch even more cases.  Thoughts?

amcasey: Actually, it looks like `Dictionary<K, V>` doesn't implement `IReadOnlyDictionary<K, V>`?

amcasey:
Looking better :smile:

image

However, it looks like there's still a case where arguments is a singleton array containing a dictionary:

image

I'm pretty reluctant to make this method recursive, but it might be worthwhile to special-case singleton arrays.  Thoughts?

amcasey:
The CI failure appears to be a test timeout.  The new code is definitely slower than `ToString`, but I'd be surprised to learn the tests run with ETW attached, so I wouldn't expect this method to be called anyway.  Also, it looks like the previous commit passed, so I'm assuming this is just a flaky test.  @AArnott is there something you'd like me to investigate or do we just need to trigger a re-run?

amcasey:
Force push is a rebase.

src/StreamJsonRpc/JsonRpcEventSource.cs
AArnott: @amcasey: [it most decidedly does implement `IReadOnlyDictionary<K, V>`](https://apisof.net/catalog/589059d3-d7f8-e09b-705d-91a461971cc2). And I'm 99% sure they implemented that interface in the same release of .NET that introduced the interface.

AArnott:
> but it might be worthwhile to special-case singleton arrays

How do you determine whether an array is a singleton? Or maybe rather than singleton (a `static` array that has unique meaning) you meant an array with just 1 element?
I'm looking into that now to see if that's a legit special case.

AArnott:
I found that `SetAdditionalContextPropertiesAsync` actually passes an args array that just happens to pass one arg with a dictionary, so this isn't a strong pattern where we'd want to special case it. The dictionary there could have been any argument in an array (or dictionary) of any size.

AArnott (approved):

amcasey:
> you meant an array with just 1 element

Yes, that's what I meant.  Sorry for not being clearer.

src/StreamJsonRpc/JsonRpcEventSource.cs
amcasey: Odd, I'm sure that comment was based on a failure I saw at runtime, but I can't reproduce it now.  I must have done something else wrong and misinterpreted the resulting exception.

PR #693, opened 2021-08-17
Drop .NET Core 2.1 support

AArnott:
.NET Core 2.1 falls out of LTS support on August 21, 2021. No reason to continue targeting it with new versions of this library.

As part of the library targeting netcoreapp3.1, we get a bunch more nullable ref annotations to update the code for.


src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -2493,9 +2497,9 @@ private void FaultPendingRequests()
         /// Cancels an individual outbound pending request.
         /// </summary>
         /// <param name="state">The <see cref="RequestId"/> associated with the request to be canceled.</param>
-        private void CancelPendingOutboundRequest(object state)
+        private void CancelPendingOutboundRequest(object? state)
```
RyanToth3, line 2500: Why allow this to be nullable if we're just going to immediately throw in that case?

RyanToth3 (approved):

src/StreamJsonRpc/JsonRpc.cs
AArnott: Because it must fit into a `CancellationToken.Register` delegate, which is `Action<object?>`. So I have to suppress the warning about a mismatch somewhere because I absolutely expect a non-null value. I opted to put it here, but it could have been that I type the delegate as `Action<object>` and then used `!` to shoe-horn it into the `Register` method call. I suppose that may have been the better option. I'll change it.

PR #697, opened 2021-08-26
Add support for top level properties and delegated dispatch

BertanAygun:
Adds support for:
* Optional top level properties on RPC messages, and implementation of it for JSON and MessagePack formatters
* An option for JsonRpc implementations to handle dispatching of calls to target object to support scenarios like ordering calls, serializing calls without blocking JsonRpc stream.

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -33,7 +33,7 @@ namespace StreamJsonRpc
     /// <remarks>
     /// Each instance of this class may only be used with a single <see cref="JsonRpc" /> instance.
     /// </remarks>
-    public class JsonMessageFormatter : IJsonRpcAsyncMessageTextFormatter, IJsonRpcFormatterState, IJsonRpcInstanceContainer, IDisposable
+    public class JsonMessageFormatter : IJsonRpcAsyncMessageTextFormatter, IJsonRpcFormatterState, IJsonRpcInstanceContainer, IJsonRpcMessageFactory, IDisposable
     {
         /// <summary>
         /// The key into an <see cref="Exception.Data"/> dictionary whose value may be a <see cref="JToken"/> that failed deserialization.
@@ -472,6 +472,24 @@ namespace StreamJsonRpc
         public object GetJsonText(JsonRpcMessage message) => throw new NotSupportedException();
 
         /// <inheritdoc/>
+        public Protocol.JsonRpcRequest CreateRequestMessage()
+        {
+            return new JsonRpcRequest(this);
+        }
+
+        /// <inheritdoc/>
+        public JsonRpcError CreateErrorMessage()
+        {
+            return new JsonRpcError();
+        }
+
+        /// <inheritdoc/>
+        public Protocol.JsonRpcResult CreateResultMessage()
+        {
+            return new JsonRpcResult(this.JsonSerializer);
+        }
+
+        /// <inheritdoc/>
         public void Dispose()
         {
             this.Dispose(true);
@@ -717,6 +735,19 @@ namespace StreamJsonRpc
                 }
             }
 
+            Dictionary<string, JToken?>? topLevelProperties = null;
+            JObject? topLevelPropertyObject = json["toplevelproperties"] as JObject;
+
+            if (topLevelPropertyObject is object)
+            {
+                topLevelProperties = new Dictionary<string, JToken?>();
+
+                foreach (JProperty property in topLevelPropertyObject.Properties())
+                {
+                    topLevelProperties[property.Name] = property.Value;
+                }
+            }
+
             return new JsonRpcRequest(this)
             {
                 RequestId = id,
@@ -724,6 +755,7 @@ namespace StreamJsonRpc
                 Arguments = arguments,
                 TraceParent = json.Value<string>("traceparent"),
                 TraceState = json.Value<string>("tracestate"),
+                TopLevelProperties = topLevelProperties,
             };
         }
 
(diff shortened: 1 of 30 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -717,13 +735,27 @@ private JsonRpcRequest ReadRequest(JToken json)
                 }
             }
 
+            Dictionary<string, JToken?>? topLevelProperties = null;
+            JObject? topLevelPropertyObject = json["toplevelproperties"] as JObject;
```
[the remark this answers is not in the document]
AArnott: You could either just store the entire message in the `JsonRpcRequest` for later, or you might store just those top-level properties that you don't recognize as included in the JSON-RPC spec. In fact such a list of recognized properties would be good to share across formatters to block reading and writing of them through these new APIs.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -866,6 +909,31 @@ public override bool TryGetArgumentByNameOrIndex(string? name, int position, Typ
 
                 return false;
             }
+
+            public override bool TryGetTopLevelProperty<T>(string name, [MaybeNull] out T value)
+            {
+                value = default;
+
+                JToken? serializedValue = null;
+                if (this.topLevelProperties?.TryGetValue(name, out serializedValue) is true)
```
AArnott, line 918: Aren't we going to add this capability to all message types? I thought we talked about adding version information from the RPC server to error messages, for example.

src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -2239,6 +2299,44 @@ public override bool TryGetArgumentByNameOrIndex(string? name, int position, Typ
                     this.formatter.deserializingMessageWithId = default;
                 }
```
AArnott, line 2300: Go ahead and throw. The caller can make the choice to trace, or retry with a different type to deserialize as, etc.

src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -2140,6 +2192,7 @@ public unsafe void Serialize(ref MessagePackWriter writer, TraceParent value, Me
         private class JsonRpcRequest : Protocol.JsonRpcRequest, IJsonRpcMessageBufferManager, IJsonRpcMessagePackRetention
         {
             private readonly MessagePackFormatter formatter;
+            private Dictionary<string, ReadOnlySequence<byte>>? topLevelProperties;
```
AArnott, line 2196: If you type this as `Sequence<byte>`, then after you transmit the message you could clear the dictionary and `Dispose` of all the values in order to recycle the buffers used for these.

test/StreamJsonRpc.Tests/JsonRpcDelegatedDispatchTests.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+        }
+
+        public bool EnableBuffering { get; set; }
+
+        public JsonRpcRequest? LastRequestDispatched { get; private set; }
+
+        public async Task FlushRequestQueueAsync(int expectedCount)
+        {
+            while (this.requestSignalQueue.Count != expectedCount)
```
AArnott, line 111: This loop looks like you're expecting another thread to enqueue messages. Yet you have no thread-safety protections in place.
What if you use `AsyncQueue<T>` instead, which is thread-safe and lets you simply `await DequeueAsync` `expectedCount` times in a loop?

AArnott (changes requested):

src/StreamJsonRpc/JsonMessageFormatter.cs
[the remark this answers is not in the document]
BertanAygun: I will update the code to that, in that case I agree that we should have a common method (maybe in JsonRpcMessage) that determines if a property name is a specced property or not

test/StreamJsonRpc.Tests/JsonRpcDelegatedDispatchTests.cs
BertanAygun: I will replace with AsyncQueue, The delay was enough 99.9% cases for queue to be all filled anyway in the test case but I agree there was still chance of failure.

src/StreamJsonRpc/MessagePackFormatter.cs
BertanAygun: I don't see Sequence being as disposable, do you mean just clear the dictionary contents after serializing the message? If so I think the new changes might achieve that.

src/StreamJsonRpc/JsonMessageFormatter.cs
BertanAygun: New changes add this to all message types, unfortunately the fact that we can't create a base class for a custom message type caused some code repetition but I've tried to have no new object allocations as much as possible in the code paths where these methods are not used.

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: > `Sequence<T>` [is definitely disposable](http://fuget.org/packages/Nerdbank.Streams/2.8.46/lib/netstandard2.0/Nerdbank.Streams.dll/Nerdbank.Streams/Sequence%601).

We might have two dictionaries: one for outbound that stores `Sequence<T>` so it can be disposed of, and one for inbound that stores `ReadOnlySequence<T>` whose buffers are already recycled at the conclusion of deserialization.

AArnott:
I'm making some changes to your PR while reviewing. Will push shortly.

AArnott (approved):

src/StreamJsonRpc/MessagePackFormatter.cs
BertanAygun: The sequences are released now with the latest changes from both of us :)

PR #707, opened 2021-09-02
Fix tracing regression

AArnott:
In #697 we started using formatter-specific `JsonRpcRequest` objects on the transmitting side. This causes RPC requests with arguments to fail when ETW tracing is enabled because they try to read arguments using code that was only designed to do so with _deserialized_ messages rather than the one created to be transmitted.

The fix is to avoid trying to deserialize outbound arguments. We could have done an `if` check within the `JsonRpcRequest`-derived classes, but it seemed more complete and safe to just stop reusing classes that were written only with deserialization in mind by creating a far simpler outbound-only request class.

We had a test hole in the ETW case which this PR (and #706) fills.


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -864,6 +864,32 @@ internal void SetTopLevelProperty<T>(string name, [MaybeNull] T value)
...
+            internal OutboundJsonRpcRequest(JsonMessageFormatter formatter)
+            {
+                this.formatter = formatter ?? throw new ArgumentNullException(nameof(formatter));
+            }
+
+            [JsonIgnore]
+            [IgnoreDataMember]
+            public TopLevelPropertyBag? TopLevelPropertyBag { get; set; }
+
+            public override bool TrySetTopLevelProperty<T>(string name, [MaybeNull] T value)
```
BertanAygun, line 882: Now that inbound/outbound is separate, do we need to implement TrySetTopLevelProperty on the inbound object?

AArnott: Maybe not. But inbound messages become outbound messages in the remote target relay case. That isn't adequately tested, I'm learning, so I didn't want to risk regressing it.

src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -2285,7 +2290,31 @@ internal void SetTopLevelProperty<T>(string name, [MaybeNull] T value)
...
+            private readonly MessagePackFormatter formatter;
+
+            internal OutboundJsonRpcRequest(MessagePackFormatter formatter)
+            {
+                this.formatter = formatter ?? throw new ArgumentNullException(nameof(formatter));
+            }
+
+            public TopLevelPropertyBag? TopLevelPropertyBag { get; set; }
+
+            public override bool TrySetTopLevelProperty<T>(string name, [MaybeNull] T value)
```
BertanAygun, line 2304: Same comment as the Json requests messages here regarding TrySetTopLevelProperty implementation in the inbound one

BertanAygun (approved):

src/StreamJsonRpc/MessagePackFormatter.cs
AArnott: Same response, but especially true here, since remote RPC target tests don't run on messagepack at all.

src/StreamJsonRpc/JsonMessageFormatter.cs
BertanAygun: Right, I remember that case now from unit tests. Does the relay case only work with json messages or do we support with message pack as well? I am now curious if it would work with messagepack and top level properties since we had different inbound/outbound dictionaries in message pack case.

PR #711, opened 2021-09-07
Adding third party notice.

matteo-prosperi:
Adding a third party notice file to the repo.


AArnott:
As this repo does not contain any of these 3rd party binaries, nor does the package we ship, I don't _think_ we really have to add this, do we?
I would expect it's up to the *app* that actually redistributes other people's software to have such a file.

matteo-prosperi:
@AArnott, you are correct. Closing

PR #734, opened 2021-11-11
Honor ClientRequiresNamedArguments when UseSingleObjectParameterDeserialization is set

dibarbet:
Discovered while attempting to use the new attributes added in https://github.com/microsoft/vs-streamjsonrpc/pull/733

Essentially when the client sends a request with a progress token using named parameters *and* the server is using `UseSingleObjectParameterDeserialization`, vs-streamjsonrpc fails to pass along the value of `clientRequiresNamedArguments` to the `JsonProgress` object.

This is exactly the configuration I hit as vscode uses named parameters and we make heavy use of `UseSingleObjectParameterDeserialization` in our lsp server.

The problem was that we essentially have a separate code path for handling RPC server methods that accept all args as a single argument (a special case built for LSP support), and we hadn't updated that code path to propagate the `JsonRpcMethodAttribute` in the special case.

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -969,9 +969,20 @@ namespace StreamJsonRpc
                                     obj.Add(new JProperty(property.Key, property.Value));
                                 }
 
+                                // Deserialization of messages should never occur concurrently for a single instance of a formatter.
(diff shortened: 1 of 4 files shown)


AArnott (approved):

milopezc (approved):

dibarbet:
thanks for the fix!

PR #748, opened 2021-12-06
Apply latest Library.Template

AArnott:


AArnott:
The SDT tools are shown to run and pass in [this pipeline run](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=5517395&view=results).

milopezc (approved):

Issue #772, opened 2022-03-02
Unexpected exception when using ExceptionProcessing.ISerializable.


rruizGit:
Andrew,

As part of the ISerializable exception support, JsonRpc.LoadType() is called to load the type for the serialized exception. We found an issue in that method when, for example, the serialized exception's type came from v1.0.24 of a library but the client has v1.0.22 of the library. 

The code is written to handle this scenario by catching a FileLoadException and retrying without the version, but, in our case it actually received a FileNotFoundException. For example.

> System.IO.FileNotFoundException: Could not load file or assembly 'Dell.TechHub.Exceptions, Version=1.0.24.4072, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified

This is the client trying to find v1.0.24 when it has v1.0.22 of the library. The result is that instead of receiving the actual JsonRpc exception with the custom inner exception, the client gets the FileNotFoundException. 

If we replace the:
`catch (FileLoadException)`
with 
`catch (Exception ex) when (ex is FileNotFoundException or FileLoadException)`
the solution works. It removes the version from the assembly info and retries, thus finding the correct assembly.

This occurred with .Net 5 processes, we did not try with .Net framework processes. 

I'll have a PR up soon for this change.

PR #773, opened 2022-03-02
Handle FileNotFoundException thrown while deserializing exceptions

rruizGit:
See explanation in issue #772.


AArnott (approved):
Thank you. I rebased your PR to target our stable 2.10 release so you can get the fix faster.

AArnott:
Yes. It should arrive at nuget.org tomorrow.

Issue #774, opened 2022-03-03
Make support for marshaled objects available for custom types


matteo-prosperi:
Could you please consider making [marshaled object](https://github.com/microsoft/vs-streamjsonrpc/blob/main/doc/general_marshaled_objects.md) support available for custom interfaces?

The goal would be to allow an RPC interface to expose APIs like

```
public interface IMyServer
{
  Task<IFoo> GetSomethingAsync();
  Task GiveSomethingAsync(IBar bar);
}
```

The caller would call `await GetSomethingAsync()` and receive a disposable proxy of an `IFoo` object.
The caller would call `await GiveSomethingAsync(bar)` and the server would receive a disposable proxy for the `bar` object.

Calls to methods exposed by the `IFoo` and `IBar` proxies would result in RPC calls and the corresponding method being executed on the corresponding object.

The receiver of a marshaled object will be responsible for disposing of the proxy. I am not sure if it should be mandatory for interfaces that are marshaled as objects to be `IDisposable` (i.e., `interface IFoo : IDisposable` and `interface IBar : IDisposable`).

This feature request doesn't cover any other members other than methods (e.g. properties), I don't have a use case for that.

Thanks

AArnott:
The place to start investigation of this feature is probably here:

https://github.com/microsoft/vs-streamjsonrpc/blob/114ee4fe8ba1cac33b6594765012db3f33c29fbb/src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs#L23

This line (and the surrounding array) is what controls which types will get marshaled instead of serialized.

At the moment this is a static list. I believe the static list should remain, but any use of that list may be augmented with another list that is maintained on a per-JsonRpc instance basis, or perhaps by checking for an attribute on the interface being tested.

I think we should start with requiring any additional interfaces to derive from `IDisposable`, since that makes it more obvious to the receiver that they should dispose of these proxies. If we need to we can always remove that requirement, but adding it later would be a breaking change, so I prefer to start conservatively.

PR #777, opened 2022-03-07
Add RPC marshaling support for custom interfaces

matteo-prosperi:
This PR addresses https://github.com/microsoft/vs-streamjsonrpc/issues/774.

I added the `RpcMarshalable` attribute that can be applied to an interface that
- extends `IDisposable`
- doesn't have properties
- doesn't have events

I added unit tests and docs.

### DIFF
--- a/doc/rpc_marshalable_objects.md
+++ b/doc/rpc_marshalable_objects.md
@@ -236,4 +236,3 @@ These resources are released and the `IDisposable.Dispose()` method is invoked o
 The protocol for proxying a disposable object is based on [general marshaled objects](general_marshaled_objects.md).
 
 The responsibility to release resources is on the receiver of the proxy.
-That is, when the proxy holder calls either of these methods, it should also send the `$/releaseMarshaledObject` message back to the target object owner. When the proxy holder is using StreamJsonRpc, this is handled automatically when `IDisposable.Dispose()` is invoked on the proxy.
--- a/doc/rpc_marshalable_objects.md
+++ b/doc/rpc_marshalable_objects.md
@@ -1,6 +1,7 @@
 # `RpcMarshalableAttribute` support
 
-StreamJsonRpc allows the definition of marshalable interfaces: interfaces marked with `RpcMarshalableAttribute`.
+StreamJsonRpc typically *serializes* values that are passed in arguments or return values of RPC methods, which effectively transmits the data of an object or struct to the remote party.
+By applying the `RpcMarshalableAttribute` to an interface, it a proxy can be sent to effectively marshal *behavior* to the remote party instead of data, similar to other [exotic types](exotic_types.md).
 
 StreamJsonRpc allows transmitting marshalable objects (i.e., objects implementing a marshalable interface) in arguments and return values.
 
@@ -8,13 +9,13 @@ Marshalable interfaces must:
 
 1. Extend `IDisposable`.
 1. Not include any properties.
-1. Not include any event.
+1. Not include any events.
 
-The marshalable object can include properties and events as well as other additional members but only the methods defined by the marshalable interface will be available on the proxy.
+The object that implements a marshalable interface may include properties and events as well as other additional members but only the methods defined by the marshalable interface will be available on the proxy, and the data will not be serialized.
 
 ## Use cases
 
-In all cases, the special handling of a marshalable object only occurs if the container of that value is typed as the corrsponding marshalable interface.
+In all cases, the special handling of a marshalable object only occurs if the container of that value is typed as the corresponding marshalable interface.
 This means that an object that implements a marshalable interface will not necessarily be marshaled instead of serialized.
 
 For each use case, assume `Counter` is a class defined for demonstration purposes like this:
@@ -34,11 +35,11 @@ class Counter : ICounter
     {
         if (IsDisposed)
         {
-            throw new ObjectDisposedException();
+            throw new ObjectDisposedException(null);
         }
 
         count++;
-        InterfaceEvent?.Raise(this, EventArgs.Empty);
+        IncrementedEvent?.Invoke(this, EventArgs.Empty);
         return Task.CompletedTask;
     }
 
@@ -51,14 +52,14 @@ class Counter : ICounter
     {
         if (IsDisposed is false)
         {
-            DisposedEvent?.Raise(this, EventArgs.Empty);
+            DisposedEvent?.Invoke(this, EventArgs.Empty);
             IsDisposed = true;
         }
     }
 }
 
 [RpcMarshalable]
-class ICounter : IDisposable
+interface ICounter : IDisposable
 {
     Task IncrementAsync(CancellationToken ct);
 
@@ -71,7 +72,12 @@ class ICounter : IDisposable
 In the simplest case, the RPC server returns a marshalable interface object.
 
 ```cs
-class RpcServer
+interface IRpcServer
+{
+    ICounter GetCounter();
+}
+
+class RpcServer : IRpcServer
 {
     private int activeCounters;
 
@@ -137,7 +143,7 @@ var arg = new SomeClass
 await client.ProvideClassAsync(arg);
 ```
 
-While this use case is supported, be very wary of this pattern because it becomes less obvious to the receiver that an `IDisposable` value is tucked into the object tree of an argument somewhere that *must* be disposed to avoid a resource leak.
+⚠️ While this use case is supported, be very wary of this pattern because it becomes less obvious to the receiver that an `IDisposable` value is tucked into the object tree of an argument somewhere that *must* be disposed to avoid a resource leak.
 
 ### As an argument without a proxy for an RPC interface
 
@@ -171,7 +177,7 @@ class ByValueCounter : ICounter
     {
         if (IsDisposed)
         {
-            throw new ObjectDisposedException();
+            throw new ObjectDisposedException(null);
         }
 
         Count++;
@@ -223,13 +229,13 @@ In each of these cases, the receiving part will get a `ByValueCounter` object, b
 
 ## Resource leaks concerns
 
-When an marshalable object instance is sent over RPC, resources are held by both parties to marshal interactions
+When a marshalable object instance is sent over RPC, resources are held by both parties to marshal interactions
 with that object.
 
 These resources are released and the `IDisposable.Dispose()` method is invoked on the sender's marshalable object when any of these occur:
 
-1. The receiver calls `IDisposable.Dispose()` on the proxy.
-1. The JSON-RPC connection is closed.
+* The receiver calls `IDisposable.Dispose()` on the proxy.
+* The JSON-RPC connection is closed.
 
 ## Protocol
 
--- a/doc/rpc_marshalable_objects.md
+++ b/doc/rpc_marshalable_objects.md
@@ -1,0 +1,239 @@
+# `RpcMarshalableAttribute` support
+
+StreamJsonRpc allows the definition of marshalable interfaces: interfaces marked with `RpcMarshalableAttribute`.
+
+StreamJsonRpc allows transmitting marshalable objects (i.e., objects implementing a marshalable interface) in arguments and return values.
+
+Marshalable interfaces must:
+
+1. Extend `IDisposable`.
+1. Not include any properties.
+1. Not include any event.
+
+The marshalable object can include properties and events as well as other additional members but only the methods defined by the marshalable interface will be available on the proxy.
+
+## Use cases
+
+In all cases, the special handling of a marshalable object only occurs if the container of that value is typed as the corrsponding marshalable interface.
+This means that an object that implements a marshalable interface will not necessarily be marshaled instead of serialized.
+
+For each use case, assume `Counter` is a class defined for demonstration purposes like this:
+
+```cs
+class Counter : ICounter
+{
+    private int count;
+
+    public bool IsDisposed { get; private set; }
+
+    public event EventHandler? IncrementedEvent;
+
+    public event EventHandler? DisposedEvent;
+
+    public Task IncrementAsync(CancellationToken ct)
+    {
+        if (IsDisposed)
+        {
+            throw new ObjectDisposedException();
+        }
+
+        count++;
+        InterfaceEvent?.Raise(this, EventArgs.Empty);
+        return Task.CompletedTask;
+    }
+
+    public Task<int> GetCountAsync(CancellationToken ct)
+    {
+        return Task.FromResult(count);
+    }
+
+    void IDisposable.Dispose()
+    {
+        if (IsDisposed is false)
+        {
+            DisposedEvent?.Raise(this, EventArgs.Empty);
+            IsDisposed = true;
+        }
+    }
+}
+
+[RpcMarshalable]
+class ICounter : IDisposable
+{
+    Task IncrementAsync(CancellationToken ct);
+
+    Task<int> GetCountAsync(CancellationToken ct);
+}
+```
+
+### Method return value
+
+In the simplest case, the RPC server returns a marshalable interface object.
+
+```cs
+class RpcServer
+{
+    private int activeCounters;
+
+    public ICounter GetCounter()
+    {
+        var counter = new Counter();
+        counter.IncrementedEvent += OnIncrement;
+        counter.DisposedEvent += OnCounterDisposed;
+        activeCounters++;
+        return counter;
+    }
+
+    private OnIncrement(object sender, EventArgs e)
+    {
+        // Do something
+    }
+
+    private OnCounterDisposed(object sender, EventArgs e)
+    {
+        counter.IncrementedEvent -= OnIncrement;
+        counter.DisposedEvent -= OnCounterDisposed;
+        activeCounters--;
+    }
+}
+```
+
+### Method argument
+
+In this use case the RPC *client* provides the marshalable object to the server:
+
+```cs
+interface IRpcContract
+{
+    Task ProvideCounterAsync(ICounter counter);
+}
+
+IRpcContract client = jsonRpc.Attach<IRpcContract>();
+var counter = new Counter();
+await client.ProvideCounterAsync(counter);
+```
+
+### Value within a single argument's object graph
+
+In this use case the RPC client again provides the marshalable object to the server,
+but this time it passes it as a property of an object used as the argument.
+
+```cs
+class SomeClass
+{
+    public ICounter Counter { get; set; }
+}
+
+interface IRpcContract
+{
+    Task ProvideClassAsync(SomeClass value);
+}
+
+IRpcContract client = jsonRpc.Attach<IRpcContract>();
+var arg = new SomeClass
+{
+    Counter = new Counter(),
+};
+await client.ProvideClassAsync(arg);
+```
+
+While this use case is supported, be very wary of this pattern because it becomes less obvious to the receiver that an `IDisposable` value is tucked into the object tree of an argument somewhere that *must* be disposed to avoid a resource leak.
+
+### As an argument without a proxy for an RPC interface
+
+When you are not using an RPC interface and dynamically generated proxy that implements it, you can still pass a marshalable object as an argument by explicitly passing in the declared parameter types to the `InvokeWithCancellationAsync` call:
+
+```cs
+ICounter arg = new Counter();
+await jsonRpc.InvokeWithCancellationAsync(
+    "methodName",
+    new object?[] { arg },
+    new Type[] { typeof(ICounter) },
+    cancellationToken);
+```
+
+### Invalid cases
+
+Here are some examples of where an object that implements a marshalable interface is serialized (i.e. by value) instead of being marshaled (i.e. by reference).
+
+In this example, although `ByValueCounter` implements `ICounter`, its declared parameter type is `ByValueCounter`:
+
+```cs
+[DataContract]
+class ByValueCounter : ICounter
+{
+    [DataMember]
+    public int Count { get; set; }
+
+    public bool IsDisposed { get; private set; }
+
+    public Task IncrementAsync(CancellationToken ct)
+    {
+        if (IsDisposed)
+        {
+            throw new ObjectDisposedException();
+        }
+
+        Count++;
+        return Task.CompletedTask;
+    }
+
+    public Task<int> GetCountAsync(CancellationToken ct)
+    {
+        return Task.FromResult(Count);
+    }
+
+    void IDisposable.Dispose()
+    {
+        IsDisposed = true;
+    }
+}
+
+ICounter arg = new ByValueCounter();
+await jsonRpc.InvokeWithCancellationAsync(
+    "methodName",
+    new object?[] { arg },
+    new Type[] { typeof(ByValueCounter) },
+    cancellationToken);
+```
+
+Here is the same situation with an RPC interface:
+
+```cs
+interface IRpcContract
+{
+    Task ProvideCounterbyValueAsync(ByValueCounter value);
+}
+
+IRpcContract client = jsonRpc.Attach<IRpcContract>();
+ByValueCounter arg = new ByValueCounter();
+await client.ProvideCounterbyValueAsync(arg);
+```
+
+Or similar for an object returned from an RPC server method:
+
+```cs
+class RpcServer
+{
+    public ByValueCounter GetCounterByValue() => new ByValueCounter();
+}
+```
+
+In each of these cases, the receiving part will get a `ByValueCounter` object, but calling any method on that object will be a local call to that object rather than being remoted back to the original object. Also, disposing the received object or closing the JSON-RPC connection will not dispose the original object.
+
+## Resource leaks concerns
+
+When an marshalable object instance is sent over RPC, resources are held by both parties to marshal interactions
+with that object.
+
+These resources are released and the `IDisposable.Dispose()` method is invoked on the sender's marshalable object when any of these occur:
+
+1. The receiver calls `IDisposable.Dispose()` on the proxy.
+1. The JSON-RPC connection is closed.
+
+## Protocol
+
+The protocol for proxying a disposable object is based on [general marshaled objects](general_marshaled_objects.md).
+
+The responsibility to release resources is on the receiver of the proxy.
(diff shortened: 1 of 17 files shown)


src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -1479,34 +1457,29 @@ public override void WriteJson(JsonWriter writer, Stream? value, JsonSerializer
...
+            private string DebuggerDisplay => $"Converter for marhalable interfaces";
 
             public override bool CanConvert(Type objectType) =>
-                objectType == this.implicitlyConvertedType ||
-                (this.implicitlyConvertedType.IsGenericTypeDefinition && objectType.IsGenericType && objectType.GetGenericTypeDefinition() == this.implicitlyConvertedType);
+                MessageFormatterRpcMarshaledContextTracker.GetMarshalOptionsForType(objectType) is not null;
 
             public override object? ReadJson(JsonReader reader, Type objectType, object? existingValue, JsonSerializer serializer)
             {
+                (JsonRpcProxyOptions ProxyOptions, JsonRpcTargetOptions TargetOptions)? marshalOptions
```
matteo-prosperi, line 1474: The current implementation is simpler than the original one which had multiple instances of `RpcMarshalableImplicitConverter`, one for each marshalable type. On the other hand, we are not caching the options anymore and we have to evaluate the `objectType` twice (one during `CanConvert` and then again during `ReadJson`). Should I go back to caching the options in different instances of `RpcMarshalableImplicitConverter` (one per type)?

src/StreamJsonRpc/RpcMarshalableAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace StreamJsonRpc
+{
+    using System;
+
+    /// <summary>
+    /// Marks an interface as marshalable.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Interface, AllowMultiple = false, Inherited = false)]
```
matteo-prosperi, line 11: Note: this can only be applied to interfaces. And is not automatically inherited..

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -87,6 +87,44 @@ private enum MarshalMode
...
+                {
+                    throw new NotSupportedException(Resources.MarshalableInterfaceHasEvents);
+                }
+
+                if (type.GetProperties().Any())
+                {
+                    throw new NotSupportedException(Resources.MarshalableInterfaceHasProperties);
+                }
+
+                return (new JsonRpcProxyOptions(), new JsonRpcTargetOptions { NotifyClientOfEvents = false, DisposeOnDisconnect = true });
```
matteo-prosperi, line 122: @AArnott, what should we do for `DisposeOnDisconnect` for custom interfaces? I opted for `true` considering that we are forcing the interface to extend `IDisposable`, but I don't have a strong opinion on this.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,520 @@
...
+using Nerdbank.Streams;
+using Newtonsoft.Json;
+using StreamJsonRpc;
+using Xunit;
+using Xunit.Abstractions;
+
+/// <summary>
+/// Tests the proxying of interfaces marked with <see cref="RpcMarshalableAttribute"/>.
+/// </summary>
+public abstract class MarshalableProxyTests : TestBase
```
matteo-prosperi, line 21: I replicated all the tests that are available for `IDisposable` and added some making sure that
- `DisposeOnDisconnect` is honored
- the proxies can be used
- exceptions are thrown if the interfaces are invalid for marshaling (non disposable, have events, or have properties)
- generic interfaces work

src/StreamJsonRpc/RpcMarshalableAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
AArnott, line 1: nit: all the other attribute classes are declared in files under the `Reflection` folder. But the namespace you have is correct.

src/StreamJsonRpc/RpcMarshalableAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace StreamJsonRpc
```
AArnott, line 4: nit: consider using a namespace statement instead of a block:

```cs
namespace StreamJsonRpc;
```

src/StreamJsonRpc/Resources.resx
```diff
@@ -208,6 +208,15 @@
   <data name="JsonRpcMethodAndIgnoreAttributesFound" xml:space="preserve">
     <value>Conflicting JsonRpcMethodAttribute and JsonRpcIgnoreAttribute found on the same method: {0}.</value>
   </data>
+  <data name="MarshalableInterfaceHasEvents" xml:space="preserve">
+    <value>Interfaces with events cannot be marshaled</value>
```
AArnott, line 212: Please end each of your error messages with periods.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -1479,34 +1457,29 @@ public override void WriteJson(JsonWriter writer, Stream? value, JsonSerializer
...
-                this.targetOptions = targetOptions;
-            }
-
-            private RpcMarshalableImplicitConverter(RpcMarshalableImplicitConverter copyFrom, Type implicitlyConvertedType)
-                : this(implicitlyConvertedType, copyFrom.jsonMessageFormatter, copyFrom.proxyOptions, copyFrom.targetOptions)
-            {
             }
 
-            private string DebuggerDisplay => $"Implicit converter for: {this.implicitlyConvertedType.Name}";
+            private string DebuggerDisplay => $"Converter for marhalable interfaces";
```
AArnott, line 1467: typo: marhalable

src/StreamJsonRpc/JsonMessageFormatter.cs
AArnott: I think what you have here is fine, so long as if it's too slow to look up the options for a given type, we can optimize the `GetMarshalOptionsForType` method itself to cache the result.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -87,6 +87,44 @@ private enum MarshalMode
             MarshallingRealObject = 1,
         }
 
+        internal static (JsonRpcProxyOptions ProxyOptions, JsonRpcTargetOptions TargetOptions)? GetMarshalOptionsForType(Type type)
+        {
+            Requires.NotNull(type, nameof(type));
```
AArnott, line 92: No need to test for null inputs for internal methods since we're using nullable-type reference annotations which would produce a build warning (which would break PR/CI builds).

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -87,6 +87,44 @@ private enum MarshalMode
...
+            }
+
+            if (type.IsInterface && type.GetCustomAttribute<RpcMarshalableAttribute>() is not null)
+            {
+                if (typeof(IDisposable).IsAssignableFrom(type) is false)
+                {
+                    throw new NotSupportedException(Resources.MarshalableInterfaceNotDisposable);
+                }
+
+                if (type.GetEvents().Any())
```
AArnott, line 112: using the `.Any()` extension method is slower than using the `Length` or `Count` properties, and sometimes allocates memory when the property alternatives would not. For each of these, as they return arrays can you use the `Length` property instead?

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -87,6 +87,44 @@ private enum MarshalMode
...
+                if (implicitlyMarshaledType == type ||
+                    (implicitlyMarshaledType.IsGenericTypeDefinition &&
+                     type.IsConstructedGenericType &&
+                     implicitlyMarshaledType == type.GetGenericTypeDefinition()))
+                {
+                    return (proxyOptions, targetOptions);
+                }
+            }
+
+            if (type.IsInterface && type.GetCustomAttribute<RpcMarshalableAttribute>() is not null)
```
AArnott, line 105: The happy path through this block allocates 2 arrays and potentially many objects. It seems worth caching any Type that passes through the happy path so that each compliant one only pays the penalty once. 
We could also reuse the resulting tuple to avoid a couple more allocations (effectively making this method alloc-free after the first call for a given Type) since although they are mutable, this is an internal method and its callers are not expected to mutate the returned objects.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,522 @@
...
+    {
+    }
+
+    [RpcMarshalable]
+    public interface IMarshalableWithProperties : IDisposable
+    {
+        int Foo { get; }
+    }
+
+    [RpcMarshalable]
```
AArnott, line 71: What about a negative test for an interface that *isn't* attributed?

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,522 @@
...
+        await disposed.WaitAsync(this.TimeoutToken);
+        Assumes.NotNull(this.server.ReceivedProxy);
+
+        WeakReference weakProxy = new WeakReference(this.server.ReceivedProxy);
+        this.server.ReceivedProxy = null;
+        return (weakProxy, weakTarget);
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private async Task<WeakReference> NoLeakWhenTransmissionFailsAfterTokenGenerated_Helper()
```
AArnott, line 349: Such helper methods *do* need to be separate methods, but now that C# allows us to attribute local methods, the most maintainable and reader-friendly way to lay this out is to move all the helper methods into their test methods as local methods (typically named something shorter as well, like `Helper`).

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,522 @@
...
+    }
+
+    [RpcMarshalable]
+    public interface IMarshalable : IDisposable
+    {
+        Task DoSomethingAsync();
+    }
+
+    [RpcMarshalable]
+    public interface IGenericMarshalable<T> : IMarshalable
```
AArnott, line 55: How about a negative test for a non-attributed interface that derives from an attributed interface?

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,522 @@
...
+
+        public Task AcceptNonDisposableMarshalableAsync(INonDisposableMarshalable nonDisposable) => Task.CompletedTask;
+
+        public Task AcceptMarshalableWithPropertiesAsync(IMarshalableWithProperties marshalableWithProperties) => Task.CompletedTask;
+
+        public Task AcceptMarshalableWithEventsAsync(IMarshalableWithEvents marshalableWithEvents) => Task.CompletedTask;
+    }
+
+    [DataContract]
+    public class ProxyContainer<T>
```
AArnott, line 445: How about a class which implements a marshalable interface and includes `[DataContract]` but is passed by its concrete type rather than as the interface? In such a case, it should be passed by-value instead of by-reference.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,522 @@
...
+        public Task<int> AcceptDataContainerAsync(DataContainer dataContainer) => Task.FromResult(dataContainer.Data?.Value ?? 0);
+
+        public Task AcceptNonDisposableMarshalableAsync(INonDisposableMarshalable nonDisposable) => Task.CompletedTask;
+
+        public Task AcceptMarshalableWithPropertiesAsync(IMarshalableWithProperties marshalableWithProperties) => Task.CompletedTask;
+
+        public Task AcceptMarshalableWithEventsAsync(IMarshalableWithEvents marshalableWithEvents) => Task.CompletedTask;
+    }
+
+    [DataContract]
```
AArnott, line 444: How about a class which implements a marshalable interface that *isn't* attributed with `[DataContract]` (which I expect to work)?

AArnott (changes requested):
Looks good overall. It's exciting to see this come together so quickly.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
matteo-prosperi: `IMarshalableDataAsArg_ShouldSerialize` and `IMarshalableDataAsObjectWithinArg_ShouldSerialize` should already test that. We were already testing this case for `IDisposable` and I duplicated all the `IDisposable` tests here.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
AArnott: I think that's a good default.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: Ah yes, I see. Thanks.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -87,6 +90,50 @@ private enum MarshalMode
...
+
+                    if (type.GetProperties().Length > 0)
+                    {
+                        throw new NotSupportedException(Resources.MarshalableInterfaceHasProperties);
+                    }
+
+                    return RpcMarshalableInterfaceDefaultOptions;
+                }
+
+                return null;
```
matteo-prosperi, line 133: @AArnott I added caching but should I cache `null`?
The only case in which we could have `null` here if I get an interface here but it is not marked as marshalable and it's not in the implicit list. Should I instead throw an exception if I ever encounter such interface? Or is there a valid use case (or a future use case) for an interface to be passed here that is not marshalable? (the only case I can think of is if there is a lower priority formatter that will also accept interfaces)

matteo-prosperi: Not caching `null` anymore as discussed in person.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -1477,31 +1461,58 @@ public override void WriteJson(JsonWriter writer, Stream? value, JsonSerializer
         }
 
         [DebuggerDisplay("{" + nameof(DebuggerDisplay) + "}")]
-        private class RpcMarshalableImplicitConverter : JsonConverter
+        private class RpcMarshalableReadConverter : JsonConverter
```
matteo-prosperi, line 1464: @AArnott, I split `RpcMarshalableImplicitConverter` into two.
`RpcMarshalableReadConverter` resolves the type options during deserialization, but cannot write. This is added to the formatter's converter list.
`RpcMarshalableConverter` has properties for the interface type and the corresponding options (like the old `RpcMarshalableImplicitConverter` did). This is necessary because `WriteJson` doesn't take the definition type and trying to find the type at runtime would misbehave in case an object implements multiple marshalable interfaces (this is what was breaking the tests on Mono).

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -87,6 +90,62 @@ private enum MarshalMode
...
+                    proxyOptions = typeProxyOptions;
+                    targetOptions = typeTargetOptions;
+                    MarshaledTypes.TryAdd(type, (proxyOptions, targetOptions));
+                    return true;
+                }
+            }
+
+            if (type.GetCustomAttribute<RpcMarshalableAttribute>() is not null)
+            {
+                if (typeof(IDisposable).IsAssignableFrom(type) is false)
```
BertanAygun, line 125: Do we need to check for default interface methods, can we even check for it? I was thinking it could be trouble if client/server has slightly different versions where the default implementation is different. Although it wouldn't be a problem specific to marshaled interface case.

matteo-prosperi, line 125: @AArnott, should I allow `IAsyncDisposable` or `Microsoft.VisualStudio.Threading.IAsyncDisposable` as an alternative to `IDisposable`?

src/StreamJsonRpc/Reflection/RpcMarshalableAttribute.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace StreamJsonRpc;
+
+using System;
+
+/// <summary>
+/// Marks an interface as marshalable.
+/// </summary>
+[AttributeUsage(AttributeTargets.Interface, AllowMultiple = false, Inherited = false)]
+public class RpcMarshalableAttribute : Attribute
```
matteo-prosperi, line 12: @AArnott, should we have `DisposeTargetOnDisconnect` as a property of this attribute? If we do, should the default be `true` or `false`?

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
matteo-prosperi: Discussed in person. Any consideration regarding default methods in interfaces should be addressed across both marshaling interfaces and "normal" server/client proxies as they should behave in the same way.

src/StreamJsonRpc/Reflection/RpcMarshalableAttribute.cs
AArnott: I don't think we should. The only behavior should be `true`. My thinking is that the underlying protocol already describes that disposing the proxy _will_ dispose the marshaled object. This means that passing a marshalable object over RPC signifies an ownership transfer. A dropped connection should not leave any resources unfreed, so we should guarantee that Dispose is called on behalf of the proxy if the proxy cannot do it do to a dropped connection.

src/StreamJsonRpc/Reflection/RpcMarshalableAttribute.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace StreamJsonRpc;
+
+using System;
+
+/// <summary>
+/// Marks an interface as marshalable.
```
AArnott, line 9: I feel like this could be beefed up a bit. How about:

> Designates an interface that is used in an RPC contract to marshal the object so the receiver can invoke remote methods on it instead of serializing the object to send its data to the remote end.

Then a `<remarks>` section that has something like this would probably be helpful:

```cs
/// <remarks>
/// <see href="https://github.com/microsoft/vs-streamjsonrpc/blob/main/doc/rpc_marshalable_objects.md">Learn more about marshable interfaces</see>.
/// </remarks>
```

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -0,0 +1,621 @@
...
+    [Fact]
+    public async Task InterfacesMustBeMarkedAsRpcMarshalable()
+    {
+        await Assert.ThrowsAnyAsync<Exception>(() => this.client.AcceptNonMarshalableAsync(new NonDataContractMarshalable()));
+    }
+
+    [Fact]
+    public async Task RpcMarshalableAttributeDoesntAffectDerivedInterfaces()
+    {
+        await Assert.ThrowsAnyAsync<Exception>(() => this.client.AcceptNonMarshalableDerivedFromMarshalablesAsync(new NonDataContractMarshalable()));
```
AArnott, line 195: Why does this throw? 
What if the interface included attributes to instruct the formatter as to how to format it as data, the way interfaces can typically do? I would expect that to work, so it would be interesting to see a test to show it.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
AArnott: I don't think so. Proxies never implement these async dispose interfaces because Dispose is always transmitted to the server as a notification, so there is nothing for the proxy to wait for. I think the marshaling behavior is that the server object can implement the async disposable interfaces and if it does, StreamJsonRpc will invoke it, but this doesn't manifest on the proxy. As this interface becomes what the proxy implements, it shouldn't derive from the async interface.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
matteo-prosperi: You are saying that an interface could be marked as `[DataContract]`? If that is supported, I will add tests for that.

But what do we do in case an interface is both `[DataContract]` and `RpcMarshalable`? Which one wins?

About why this throws: `AcceptNonMarshalableDerivedFromMarshalablesAsync` takes a parameter of type `INonMarshalableDerivedFromMarshalable` where
```cs
    public interface INonMarshalableDerivedFromMarshalable : IMarshalable
    {
    }

    [RpcMarshalable]
    public interface IGenericMarshalable<T> : IMarshalable
    {
        Task<T> DoSomethingWithParameterAsync(T paremeter);
    }
```

We don't allow to use a derived interface type as if it were also `RpcMarshalable`.

This is reinforced by `RpcMarshalableAttribute` being marked as `Inherited = false`.

matteo-prosperi: We clarified in person that an interface could be both `[RpcMarshalable]` and otherwise serializable (e.g., through `[MessagePackFormatter]` or `[JsonConverter]`) and that the desired behavior would be either throwing an exception or marshaling.

I changed the implementation to marshal these interfaces (I fear that the check necessary to throw an exception would lead to negative performance impact).

For MessagePack, the solution is simply to move `RpcMarshalableImplicitResolver` from the bottom to the top of the resolver list.

For Newtonsoft.Json, the solution is to leverage the contract resolver.

AArnott (approved):
Looks really good. Just a few touch-up suggestions.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -1515,26 +1491,17 @@ public override void WriteJson(JsonWriter writer, object? value, JsonSerializer
                 {
                     writer.WriteNull();
                 }
+                else if (this.interfaceType.IsAssignableFrom(value.GetType()) is false)
```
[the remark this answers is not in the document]
matteo-prosperi: Will change to `!`.
Myself, I never liked the `!` syntax and even preferred `== false` to it because the `!` at the beginning of the statement is sometimes hard to see. Also `is false` resembles natural language more.
But will adapt to the local codestyle and use `!` here :-)

src/StreamJsonRpc/MessagePackFormatter.cs
```diff
@@ -587,6 +587,9 @@ private MessagePackSerializerOptions MassageUserDataOptions(MessagePackSerialize
             // Add our own resolvers to fill in specialized behavior if the user doesn't provide/override it by their own resolver.
             var resolvers = new IFormatterResolver[]
             {
+                // Support for marshalled objects.
+                new RpcMarshalableImplicitResolver(this),
```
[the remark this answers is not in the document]
matteo-prosperi: If I switch the order of the resolvers, an interface that is both marshalable and serializable (e.g. through `MessagePackFormatterAttribute`) would be serialized:
1. it would be a different behavior than when using Newtonsoft.Json
2. we discussed in [this comment](https://github.com/microsoft/vs-streamjsonrpc/pull/777#discussion_r825001683) that marshalability should have precedence. For reference:
> `RpcMarshable` should win because it's universal (same attribute regardless of formatter), and switching formatters shouldn't change a type from serializing to marshalling

doc/rpc_marshalable_objects.md
```diff
@@ -0,0 +1,244 @@
...
+
+1. The receiver calls `IDisposable.Dispose()` on the proxy.
+1. The JSON-RPC connection is closed.
+
+## Protocol
+
+The protocol for proxying a disposable object is based on [general marshaled objects](general_marshaled_objects.md).
+
+The responsibility to release resources is on the receiver of the proxy.
+That is, when the proxy holder calls either of these methods, it should also send the `$/releaseMarshaledObject` message back to the target object owner. When the proxy holder is using StreamJsonRpc, this is handled automatically when `IDisposable.Dispose()` is invoked on the proxy.
```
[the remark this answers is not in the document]
matteo-prosperi: It was copied from [observer.md](https://github.com/microsoft/vs-streamjsonrpc/blob/main/doc/observer.md#protocol), but I don't think we need this statement,
```
The responsibility to release resources is on the receiver of the proxy.
```
should be sufficient.

src/StreamJsonRpc/MessagePackFormatter.cs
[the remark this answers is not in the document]
AArnott: I get that if you move it too far down it would be serialized, but the user-types resolver is typically empty until they populate it IIRC. So I thought one position down wouldn't have broken this.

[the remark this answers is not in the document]
matteo-prosperi: Moving it one position down breaks the case I described above because `userSuppliedOptions.Resolver` by default includes `AttributeFormatterResolver`.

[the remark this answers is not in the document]
AArnott: Ok. We'll leave it then.

Issue #781, opened 2022-03-22
`AddLocalRpcTarget<IServerDerived>` does not publish events in parent interface


rruizGit:
In the case where:

_public interface IServerDerived : IServer_

where IServer declares an event and 

_JsonRpc.AddLocalRpcTarget<IServerDerived>_ 

is employed to publish the object using an interface, the methods of the parent interface (_IServer_) will be available but not the events. 

(I will be adding a PR shortly with an extra unit test that exercises this situation along with a fix).

PR #782, opened 2022-03-22
Hook events in base interfaces when calling `AddLocalRpcTarget<T>`

rruizGit:
Fix for issue #781. It includes a unit test that would have failed before the fix.


AArnott (approved):
First rate. Thank you!

PR #783, opened 2022-03-22
Honor user-provided Json IContractResolver while maintaining marshaling capabilities

matteo-prosperi:
Addressing failures similar to
```
StreamJsonRpc.RemoteMethodNotFoundException : Unable to find method 'xxx' on {no object} for the following reasons: Deserializing JSON-RPC argument with name "observer" and position 1 to type "xxx" failed: Could not create an instance of type xxx. Type is an interface or abstract class and cannot be instantiated. Path 'params[1].__jsonrpc_marshaled'.
```
when calling APIs under the following conditions:
- using Newtonsoft.Json
- the API has marshalable (`IObserver<>`, `IDisposable` or interfaces with `RpcMarshalableAttribute`) objects as parameters or return values
- the user replaces `JsonMessageFormatter`'s `ContractResolver`.

This issue has worsened since https://github.com/microsoft/vs-streamjsonrpc/pull/777 due to more behavior being delegated to the `ContractResolver`

This PR integrates the `ContractResolver` provided by the user with `MarshalContractResolver` before the first serialization or deserialization.

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -13,6 +13,7 @@ namespace StreamJsonRpc
     using System.IO.Pipelines;
     using System.Linq;
     using System.Reflection;
+    using System.Runtime.CompilerServices;
     using System.Runtime.ExceptionServices;
     using System.Runtime.Serialization;
     using System.Text;
@@ -161,6 +162,16 @@ namespace StreamJsonRpc
         /// The message whose arguments are being deserialized.
         /// </summary>
         private JsonRpcMessage? deserializingMessage;
+
+        /// <summary>
+        /// Whether <see cref="EnforceFormatterIsInitialized"/> has been executed.
+        /// </summary>
+        private bool formatterInitializationChecked;
+
+        /// <summary>
+        /// Object used to lock when running <see cref="EnforceFormatterIsInitialized"/>.
+        /// </summary>
+        private object formatterInitializationLock = new();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="JsonMessageFormatter"/> class
@@ -390,6 +401,8 @@ namespace StreamJsonRpc
         {
             Requires.NotNull(json, nameof(json));
 
+            this.EnforceFormatterIsInitialized();
+
             try
             {
                 switch (this.ProtocolVersion.Major)
@@ -432,6 +445,8 @@ namespace StreamJsonRpc
         /// <returns>The JSON of the message.</returns>
         public JToken Serialize(JsonRpcMessage message)
         {
+            this.EnforceFormatterIsInitialized();
+
             try
             {
                 this.observedTransmittedRequestWithStringId |= message is JsonRpcRequest request && request.RequestId.String is not null;
@@ -526,6 +541,25 @@ namespace StreamJsonRpc
             }
 
             return jtokenArray;
+        }
+
+        private void EnforceFormatterIsInitialized()
+        {
+            if (!this.formatterInitializationChecked)
+            {
+                lock (this.formatterInitializationLock)
+                {
+                    if (!this.formatterInitializationChecked)
+                    {
+                        this.formatterInitializationChecked = true;
+                        IContractResolver? originalContractResolver = this.JsonSerializer.ContractResolver;
+                        if (originalContractResolver is not MarshalContractResolver)
+                        {
+                            this.JsonSerializer.ContractResolver = new MarshalContractResolver(this, originalContractResolver);
+                        }
+                    }
+                }
+            }
         }
 
         private void VerifyProtocolCompliance(bool condition, JToken message, string? explanation = null)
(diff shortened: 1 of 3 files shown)


test/StreamJsonRpc.Tests/JsonContractResolverTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    protected class MockObserver<T> : IObserver<T>
+    {
+        private readonly TaskCompletionSource<ImmutableList<T>> completed = new TaskCompletionSource<ImmutableList<T>>();
+
+        internal event EventHandler<T>? Next;
+
+        [System.Runtime.Serialization.IgnoreDataMember]
+        internal ImmutableList<T> ReceivedValues { get; private set; } = ImmutableList<T>.Empty;
+
+        [System.Runtime.Serialization.IgnoreDataMember]
```
AArnott, line 91: I'm puzzled as to why these `IgnoreDataMember` attributes are necessary. Surely the stock implementations of this interface that we use already have these applied. Were they necessary before your original object marshaling PR?

test/StreamJsonRpc.Tests/JsonContractResolverTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        var observer = new MockObserver<int>();
+        await Task.Run(() => this.client.PushCompleteAndReturn(observer)).WithCancellation(this.TimeoutToken);
+        await observer.Completion.WithCancellation(this.TimeoutToken);
+        ImmutableList<int> result = await observer.Completion;
+    }
+
+    protected IJsonRpcMessageFormatter CreateFormatter()
+    {
+        var formatter = new JsonMessageFormatter();
+        formatter.JsonSerializer.ContractResolver = new CamelCasePropertyNamesContractResolver();
```
AArnott, line 59: I feel like adding a test that combines your new marshaled interfaces feature with use of `IObserver<T>` and a custom contract resolver would go a long way to alleviating concern that these will all work together.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -162,6 +163,16 @@ public class JsonMessageFormatter : IJsonRpcAsyncMessageTextFormatter, IJsonRpcF
...
 
+        /// <summary>
+        /// Whether <see cref="EnforceFormatterIsInitialized"/> has been executed.
+        /// </summary>
+        private bool formatterInitializationChecked;
+
+        /// <summary>
+        /// Object used to lock when running <see cref="EnforceFormatterIsInitialized"/>.
+        /// </summary>
+        private object formatterInitializationLock = new();
```
AArnott, line 174: Studies have shown more than one lock field on a class greatly increases odds of threading bugs. So rather than dedicate this field to just locking one particular thing, I'd like to consider it _the_ object to lock for anything the formatter requires. To that end, can you please rename it to `syncObject` to match the naming convention used elsewhere?
Also please add the `readonly` modifier. This defends against thread-safety bugs when/if this field were ever reassigned (which I have seen as real bugs in the wild before). StyleCop will require that you move the field up at that point to be amongst the other `readonly` fields.

test/StreamJsonRpc.Tests/JsonContractResolverTests.cs
matteo-prosperi: These are copied from the original `IObserver` tests, I removed them now. I don't think they have any effect on the test.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -528,6 +543,25 @@ private static object[] PartiallyParsePositionalArguments(JArray args)
...
 
+        private void EnforceFormatterIsInitialized()
+        {
+            if (!this.formatterInitializationChecked)
+            {
+                lock (this.formatterInitializationLock)
+                {
+                    if (!this.formatterInitializationChecked)
+                    {
+                        this.formatterInitializationChecked = true;
```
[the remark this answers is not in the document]
AArnott: Here's another excellent article (having only skimmed it): https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/double-checked-locking/

test/StreamJsonRpc.Tests/JsonContractResolverTests.cs
AArnott: I'd like to see the same formatter work for both IObserver and custom interface marshaling, while configured with a custom contract resolver.

src/StreamJsonRpc/JsonMessageFormatter.cs
```diff
@@ -1667,12 +1701,19 @@ public override void WriteJson(JsonWriter writer, Exception? value, JsonSerializ
         private class MarshalContractResolver : DefaultContractResolver
```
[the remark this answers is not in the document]
matteo-prosperi: It is the only method on the interface, but I am still updating the code to follow your suggestion.

src/StreamJsonRpc/JsonMessageFormatter.cs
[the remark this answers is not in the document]
matteo-prosperi: Sorry, I was trying to fix this quickly and I totally spaced on setting the flag in the wrong place.

I have moved it to the bottom of the `lock` block and removed the outer `if`. I believe that, if we want to reduce locking, we could restore the outer `if` and make the flag `volatile`. But I will leave that for another day

AArnott (approved):

PR #793, opened 2022-05-03
Configure HighEntropyVA to true

matteo-prosperi:


AArnott:
This is already the default for msbuild 
https://github.com/dotnet/msbuild/blob/8872ed650643d12340fb22ece223c1aacf5a30a0/src/Tasks/Microsoft.NET.props#L37

binlog confirms this is the default for our pipeline.

matteo-prosperi:
Turns out this is not necessary

PR #804, opened 2022-06-10
Fix visibility checks when generating proxies based on internal interfaces

### LINKED ISSUE #789: ServiceJsonRpcDescriptor.JsonRpcConnection.ConstructRpcClient<T> fails to create an instance
When called with an interface that implements another interface like so `TypeLoadException` is thrown.

Assembly1:
```C#
internal class C
{
  internal interface ICallback : I
  {
  }
}
```

Assembly2:
```C#
[assembly: InternalsVisibleTo("Assembly1")]

internal interface I
{
   ValueTask<E> Get(D d, CancellationToken token);
}

internal class D
{
}

internal class E
{
}
```

```
Feature 'Convert tuple to struct refactoring' is currently unavailable due to an internal error. System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.TypeLoadException: Access is denied: 'D'.
       at System.Reflection.Emit.TypeBuilder.TermCreateClass(RuntimeModule module, Int32 tk, ObjectHandleOnStack type)
       at System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
       at System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
       at StreamJsonRpc.ProxyGeneration.CreateParameterObjectType(ModuleBuilder moduleBuilder, ParameterInfo[] parameters)
       at StreamJsonRpc.ProxyGeneration.Get(TypeInfo serviceInterface)
       at StreamJsonRpc.JsonRpc.Attach[T](JsonRpcProxyOptions options)
       at Microsoft.ServiceHub.Framework.ServiceJsonRpcDescriptor.JsonRpcConnection.ConstructRpcClient[T]()
--- End of inner exception stack trace ---
       at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
       at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
       at Microsoft.ServiceHub.Framework.ServiceRpcDescriptor.RpcConnection.ConstructRpcClient(Type interfaceType)
```

AArnott:
.NET Reflection doesn't report members as part of a derived interface. One must ask each interface up the type hierarchy for its members. In #789, an internal interface was defined in the local assembly and declared no members, so we thought no skip visibility attribute was necessary. Once we add the type hierarchy traversal, the additional assembly is discovered and the skip visibility attribute added, allowing the new test to pass.

Fixes #789

### DIFF
--- a/src/StreamJsonRpc/SkipClrVisibilityChecks.cs
+++ b/src/StreamJsonRpc/SkipClrVisibilityChecks.cs
@@ -87,12 +87,15 @@ namespace StreamJsonRpc
             CheckForNonPublicTypes(typeInfo, assembliesDeclaringInternalTypes, visitedTypes);
 
             // Enumerate members on the interface that we're going to need to implement.
-            foreach (MethodInfo methodInfo in typeInfo.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.FlattenHierarchy))
+            foreach (TypeInfo iteratedTypeInfo in ThisAndBaseTypes(typeInfo))
             {
-                CheckForNonPublicTypes(methodInfo.ReturnType.GetTypeInfo(), assembliesDeclaringInternalTypes, visitedTypes);
-                foreach (ParameterInfo parameter in methodInfo.GetParameters())
+                foreach (MethodInfo methodInfo in iteratedTypeInfo.GetMethods(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance))
                 {
-                    CheckForNonPublicTypes(parameter.ParameterType.GetTypeInfo(), assembliesDeclaringInternalTypes, visitedTypes);
+                    CheckForNonPublicTypes(methodInfo.ReturnType.GetTypeInfo(), assembliesDeclaringInternalTypes, visitedTypes);
+                    foreach (ParameterInfo parameter in methodInfo.GetParameters())
+                    {
+                        CheckForNonPublicTypes(parameter.ParameterType.GetTypeInfo(), assembliesDeclaringInternalTypes, visitedTypes);
+                    }
                 }
             }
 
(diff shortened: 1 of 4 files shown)


test/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
```diff
@@ -137,14 +134,27 @@ public interface IServerWithGenericMethod
...
+
+    internal interface IRemoteService
+    {
+        internal interface ICallback : ExAssembly.IInternalGenericInterface<ExAssembly.SomeOtherInternalType?>
+        {
+        }
+    }
+
+    [Fact]
+    public void Tomas_Internal()
```
BertanAygun, line 155: Is it possible to update the test name to describe the scenario better?

src/StreamJsonRpc/SkipClrVisibilityChecks.cs
```diff
@@ -131,6 +134,25 @@ internal void SkipVisibilityChecksFor(AssemblyName assemblyName)
...
+                foreach (Type iface in startingPoint.GetInterfaces())
+                {
+                    yield return iface.GetTypeInfo();
+                }
+            }
+            else
+            {
+                for (TypeInfo? t = startingPoint.GetTypeInfo(); t is not null && t != typeof(object).GetTypeInfo(); t = t.BaseType?.GetTypeInfo())
+                {
+                    yield return t;
```
BertanAygun, line 151: Do we have to recurse into interfaces implemented by these base types or do we only care about the interfaces declared directly by the original starting point?

BertanAygun (approved):

src/StreamJsonRpc/SkipClrVisibilityChecks.cs
AArnott: You know, I copied this code from another place in the library without thinking. This code will never be hit because proxies are always based on *interfaces*, not classes. I'll delete.

test/StreamJsonRpc.Tests/JsonRpcProxyGenerationTests.cs
tmat: Isn't my name self-explanatory? ;)

PR #843, opened 2022-09-13
Allow marshalable objects to implement additional optional interfaces

matteo-prosperi:


src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -295,6 +334,9 @@ public MarshalToken(int __jsonrpc_marshaled, long handle, string? lifetime)
 
         [DataMember(Name = "lifetime", EmitDefaultValue = false)]
         public string? Lifetime { get; }
+
+        [DataMember(Name = "subTypeCode", EmitDefaultValue = false)]
+        public int? SubTypeCode { get; }
```
matteo-prosperi, line 339: We could get rid of this and use the full type name instead.

src/StreamJsonRpc/Reflection/RpcMarshalableKnownSubtypeAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+/// In an object doesn't implement any of the <see cref="SubType"/> interfaces,
+/// <see cref="RpcMarshalableKnownSubtypeAttribute"/> attributes are simply ignored.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Interface, AllowMultiple = true, Inherited = false)]
+public class RpcMarshalableKnownSubtypeAttribute : Attribute
+{
+    public RpcMarshalableKnownSubtypeAttribute(Type subType, int subTypeCode)
+    {
+        this.SubType = subType;
+        this.SubTypeCode = subTypeCode;
```
matteo-prosperi, line 26: We could get rid of this and use the full type name instead.

src/StreamJsonRpc/Reflection/RpcMarshalableKnownSubtypeAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+/// <summary>
+/// Attribute to be used on an interface marked with <see cref="RpcMarshalableAttribute"/> to indicate that a known
+/// inherited interface exists.
+///
+/// When an object that implements <see cref="SubType"/> is marshaled as its base interface marked with
+/// <see cref="RpcMarshalableKnownSubtypeAttribute"/>, <see cref="SubTypeCode"/> is included in the StreamJsonRpc
+/// message allowing the creation of a proxy which implements <see cref="SubType"/>.
+/// </summary>
+/// <remarks>
+/// If an object implements more than one of the <see cref="SubType"/> insterfaces, the behavior is undefined.
```
matteo-prosperi, line 15: Do we want to be more precise? Use the first matching or last matching (by SubTypeCode ordering)

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -157,20 +160,34 @@ internal MarshalToken GetToken(IRpcMarshaledContext<object> marshaledContext)
 
         long handle = this.nextUniqueHandle++;
 
+        Type actualInterface = declaredType;
+        int? subTypeCode = null;
+        foreach (RpcMarshalableKnownSubtypeAttribute attribute in declaredType.GetCustomAttributes(typeof(RpcMarshalableKnownSubtypeAttribute), inherit: false))
```
matteo-prosperi, line 165: Is this amount of reflection of upon serialization and deserialization?

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -117,10 +117,34 @@ public interface IMarshalableWithEvents : IDisposable
...
 
+    [RpcMarshalable]
+    [RpcMarshalableKnownSubtype(typeof(IMarshalableSubType1), subTypeCode: 1)]
+    [RpcMarshalableKnownSubtype(typeof(IMarshalableSubType2), subTypeCode: 2)]
+    public interface IMarshalableWithSubTypes : IDisposable
+    {
+        Task<int> GetAsync(int value);
+    }
+
+    public interface IMarshalableSubType1 : IMarshalableWithSubTypes
```
matteo-prosperi, line 128: Should I be more concerned about explicit interface implementations?

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -210,16 +227,37 @@ internal MarshalToken GetToken(IRpcMarshaledContext<object> marshaledContext)
...
 
+        Type actualInterfaceType = interfaceType;
+        if (token.Value.SubTypeCode.HasValue)
+        {
+            Type? matchedSubType = null;
+            foreach (RpcMarshalableKnownSubtypeAttribute attribute in interfaceType.GetCustomAttributes(typeof(RpcMarshalableKnownSubtypeAttribute), inherit: false))
+            {
+                if (attribute.SubTypeCode == token.Value.SubTypeCode.Value)
+                {
+                    matchedSubType = attribute.SubType;
```
matteo-prosperi, line 238: I should probably add a `throw` here if `attribute.SubType` doesn't extend `interfaceType` (otherwise it will fail client-side when casting).

Are there other conditions that we want to check?
We could verify that there aren't duplicated `SubTypeCode` values but it would make this code more complicated and slower.

src/StreamJsonRpc/Reflection/RpcMarshalableKnownSubtypeAttribute.cs
AArnott: Ya, undefined is probably unacceptable, as it could mean things work and then break later. At the very least, we should explicitly check for this condition and force a failure. 
If we want to go beyond that and support multiple interfaces, that's fine. Maybe when we encode the subtype id's, we can do so as an array instead of a scalar so we can signify all the interfaces that are allowed.

src/StreamJsonRpc/Reflection/RpcMarshalableKnownSubtypeAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+///
+/// In an object doesn't implement any of the <see cref="SubType"/> interfaces,
+/// <see cref="RpcMarshalableKnownSubtypeAttribute"/> attributes are simply ignored.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Interface, AllowMultiple = true, Inherited = false)]
+public class RpcMarshalableKnownSubtypeAttribute : Attribute
```
AArnott, line 21: nit: you use `SubType` casing in some places an `Subtype` in others. We should be consistent. I'm leaning toward `SubType`, but if "subtype" is a word, then `Subtype` is also acceptable.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: There is no such thing as explicit interface implementations at the interface level. That's something determined by the implementing class. And yes, we should support those. I suspect we already do and have tests to show it. That seems like it should be orthogonal to this feature, so I don't think you're on the hook to find or add tests for it.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -499,6 +523,40 @@ public async Task DisposeOnDisconnect()
         await disposed.WaitAsync(this.TimeoutToken);
     }
 
+    [Fact]
+    public async Task SubTypeMarshalable_CanCallMethods()
+    {
+        this.server.ReturnedMarshalableWithSubTypes = new MarshalableWithSubTypes();
+        IMarshalableWithSubTypes? proxy = await this.client.GetMarshalableWithSubTypesAsync();
+        Assert.Equal(1, await proxy!.GetAsync(1));
```
AArnott, line 531: Asserting that the proxy does _not_ implement certain interfaces here seems prudent.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -499,6 +523,40 @@ public async Task DisposeOnDisconnect()
...
+    }
+
+    [Fact]
+    public async Task SubTypeMarshalable_SubType1_CanCallMethods()
+    {
+        this.server.ReturnedMarshalableWithSubTypes = new MarshalableSubType1();
+        IMarshalableSubType1? proxy = (IMarshalableSubType1?)await this.client.GetMarshalableWithSubTypesAsync();
+        Assert.Equal(1, await proxy!.GetAsync(1));
```
AArnott, line 539: Assert that `IMarshalableSubType2` is not implemented.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -499,6 +523,40 @@ public async Task DisposeOnDisconnect()
...
+        Assert.Equal(2, await proxy!.GetPlusOneAsync(1));
+    }
+
+    [Fact]
+    public async Task SubTypeMarshalable_SubType2_CanCallMethods()
+    {
+        this.server.ReturnedMarshalableWithSubTypes = new MarshalableSubType2();
+        IMarshalableSubType2? proxy = (IMarshalableSubType2?)await this.client.GetMarshalableWithSubTypesAsync();
+        Assert.Equal(1, await proxy!.GetAsync(1));
+        Assert.Equal(3, await proxy!.GetPlusTwoAsync(1));
```
AArnott, line 549: nit: no need to use the `!` null forgiveness operator on the second line. The compiler recognizes that the line above it has already sealed your fate (it'll throw, or prove that it's non-null.)

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -728,4 +793,44 @@ public void Dispose()
...
+        public Task<int> GetAsync(int value) => Task.FromResult(value);
+
+        public Task<int> GetPlusOneAsync(int value) => Task.FromResult(value + 1);
+
+        public void Dispose()
+        {
+        }
+    }
+
+    public class MarshalableSubType2 : IMarshalableSubType2
```
AArnott, line 817: _One_ of these test classes should derive from another _class_ that implements the interface, to ensure we're not looking only one-level deep for interface implementations.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -499,6 +523,40 @@ public async Task DisposeOnDisconnect()
...
+    }
+
+    [Fact]
+    public async Task SubTypeMarshalable_UnknownSubType_CanCallMethods()
+    {
+        this.server.ReturnedMarshalableWithSubTypes = new MarshalableUnknownSubType();
+        IMarshalableWithSubTypes? proxy = await this.client.GetMarshalableWithSubTypesAsync();
+        Assert.Equal(1, await proxy!.GetAsync(1));
+    }
+
```
AArnott, line 559: Where is your null object test? We don't want to throw NRE when you call `obj.GetType()`. 😬

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -295,6 +334,9 @@ public MarshalToken(int __jsonrpc_marshaled, long handle, string? lifetime)
 
         [DataMember(Name = "lifetime", EmitDefaultValue = false)]
         public string? Lifetime { get; }
+
+        [DataMember(Name = "subTypeCode", EmitDefaultValue = false)]
```
AArnott, line 338: This addendum to the protocol should be reflected in [the doc](https://github.com/microsoft/vs-streamjsonrpc/blob/main/doc/general_marshaled_objects.md) for it as well.

AArnott: and a nit: call it `subType` instead of `subTypeCode`. `Code` doesn't really add any value, and could be added as a suffix to almost anything. This is feedback on the wire protocol property name specifically. The C# property name of `SubTypeCode` is totally fine, as it makes it clear that the type is an integer rather than a `Type`.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
AArnott: And while we're caching the initial result, we should validate it:
1. each ordinal value is used at most once.
1. each subtype actually derives from the interface the attribute is applied to.
1. each subtype is used at most once.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
AArnott: I don't think we should throw. That doesn't lend well to duck typing and versioning being independent between the two parties. Rather, I think we should implement every interface that _does_ match, including the base interface itself (which doesn't appear in the list, but is assumed to be a given because we require subtypes to derive from it.)

src/StreamJsonRpc/Reflection/RpcMarshalableKnownSubtypeAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+/// When an object that implements <see cref="SubType"/> is marshaled as its base interface marked with
+/// <see cref="RpcMarshalableKnownSubtypeAttribute"/>, <see cref="SubTypeCode"/> is included in the StreamJsonRpc
+/// message allowing the creation of a proxy which implements <see cref="SubType"/>.
+/// </summary>
+/// <remarks>
+/// If an object implements more than one of the <see cref="SubType"/> insterfaces, the behavior is undefined.
+///
+/// In an object doesn't implement any of the <see cref="SubType"/> interfaces,
+/// <see cref="RpcMarshalableKnownSubtypeAttribute"/> attributes are simply ignored.
+/// </remarks>
```
AArnott, line 19: We should document the behavior people should expect if they have a multi-level hierarchy of interfaces, including when they apply all attributes to the ultimate base interface or if they apply some of them to mid-level derived types.

AArnott (changes requested):
This is remarkable how well this fits in and how quickly you wrote it up. I have lots of comments, but I'm very enthusiastic about what you're proposing here.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
matteo-prosperi: As discussed in person, I won't add multiple interfaces to the proxy because it may result in confusing situations if the different interfaces define the methods with the same name that are implemented separately with explicit implementations.
I have updated my proposed additions to the protocol to allow multiple subtype codes though, so that we can add this feature in the future,

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
matteo-prosperi: What does this comment refer to?

src/StreamJsonRpc/Reflection/RpcMarshalableKnownSubTypeAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+/// If a message is received containing no <see cref="SubTypeCode"/>, for a marshalable interface that has known
+/// sub-types, a proxy will be created using the base interface. Unkonow <see cref="SubTypeCode"/> values will be
+/// ignored when creating the proxy.
+/// </para><para>
+/// <see cref="RpcMarshalableKnownSubTypeAttribute"/> is honored only when an object is marshaled through an RPC method
+/// that used exactly the interface the attribute is assigned to: <see cref="RpcMarshalableKnownSubTypeAttribute"/>
+/// attributes applied to interfaces extending or extended from the one the attribute is assigned to are ignored.
+/// </para></remarks>
+[AttributeUsage(AttributeTargets.Interface, AllowMultiple = true, Inherited = false)]
+public class RpcMarshalableKnownSubTypeAttribute : Attribute
```
matteo-prosperi, line 24: @AArnott, do we want to change the name of this attribute now that we allow adding interfaces don't extend the declared one?
Ideas:
- `RpcMarshalableKnownAdditionalInterfaceAttribute`
- `RpcMarshalableKnownInterfaceAttribute`
Or we can keep this name considering that most implementations will use subtypes of the declared interface.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -499,6 +585,154 @@ public async Task DisposeOnDisconnect()
...
+        IMarshalableWithSubTypes? proxy = await this.client.GetMarshalableWithSubTypesAsync();
+        Assert.Equal(1, await proxy!.GetAsync(1));
+
+        Assert.Equal(1, await ((IMarshalableSubTypeIntermediateInterface)proxy).GetAsync(1));
+        Assert.Equal(2, await ((IMarshalableSubTypeIntermediateInterface)proxy).GetPlusOneAsync(1));
+
+        // This should return 3, since MarshalableSubTypeWithIntermediateInterface implements this method explicitly
+        // but StreamJsonRpc doesn't know about the IMarshalableSubTypeIntermediateInterface because it is not part
+        // of the RPC contract, so IMarshalableSubTypeWithIntermediateInterface.GetPlusTwoAsync is invoked instead.
+        Assert.Equal(-3, await ((IMarshalableSubTypeIntermediateInterface)proxy).GetPlusTwoAsync(1));
```
matteo-prosperi, line 672: @AArnott, I want to bring your attention to this.
I think I covered, in the next two unit tests, all other combinations of objects implementing multiple interfaces with implicit and explicit implementations and the use of the `new` keyword in interfaces.

This is the only case I could find that doesn't behave well. But the problem is already present in the current implementation (even the main service proxy is affected): if you implicitly implement a hidden method from a base interface that is not part of the Rpc contract (is not one of the known sub-types), we cannot dispatch the call to that method because it doesn't have a name. The call is dispatched to the wrong method instead.

I will add a detailed explanation of this issue in the docs and provide a general guidance to users to avoid, when possible, using the `new` keyword and method name conflicts when building Rpc interfaces.

I think it is good to have this weird behavior tested here, even if it is less than ideal, so that we at least keep this behavior constant.

doc/rpc_marshalable_objects.md
```diff
@@ -237,8 +237,141 @@ These resources are released and the `IDisposable.Dispose()` method is invoked o
...
+
+The `subTypeCode` values used in `RpcMarshalableKnownSubType` attributes are used as part of the wire protocol. While it is backward compatible to remove an `RpcMarshalableKnownSubType` attribute from an interface, its `subTypeCode` value should never be reused to add a different additional interface to the same interface declaration.
+
+#### Method name conflicts and non-marshalable interfaces
+
+When a method is invoked on a marshalable object proxy, it is traslated into an RPC call to the best-matching interface method.
+
+For example, given the interfaces defined below:
+- a call to `((IFoo)proxy).DoFooAsync()` would result in an RPC call to the `DoFooAsync` method as defined by the `IFoo` interface.
+- a call to `((IBaz)proxy).DoFooAsync()` would result in an RPC call to the `DoFooAsync` method as defined by the `IBaz` interface, if the marshalable object implements `IBaz`.
```
BertanAygun, line 302: What happens otherwise? I assume proxy wouldn't even implement IBaz if the underlying object didn't right?

matteo-prosperi: Exactly right. That's explained above on line 251:
> The proxy object created for the marshalable object will only implement the subset of interfaces described by `RpcMarshalableKnownSubType` attributes that are applied to the interface type used in the RPC method declaration. The receiver of the proxy can use the `is` operator to check if an optional interface is implemented.
And 286
> A client that uses an earlier RPC contract definition will simply ignore any unknow additional interface.

AArnott (changes requested):
Exciting stuff. I still need to review the ProxyGeneration and tests, but thought I'd offer my feedback so far.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: > in the next two unit tests, all other combinations of objects implementing multiple interfaces with implicit and explicit implementations and the use of the `new` keyword in interfaces.

Thank you! Exhaustive testing of this complex area is super important both for correctness and to maintain such subtle behaviors over time.

> I think it is good to have this weird behavior tested here, even if it is less than ideal, so that we at least keep this behavior constant.

💯 I absolutely agree with your reasoning on that point.

> if you implicitly implement a hidden method from a base interface that is not part of the Rpc contract (is not one of the known sub-types), we cannot dispatch the call to that method because it doesn't have a name.

This part I don't understand. How can you *implicitly* implement an interface method without your method having a name? I understand that `public void Hello() { }` may implicitly implement an interface method, but that method has a name, and it _should_ be available to dispatch RPC requests to.
An *explicit* interface implementation on the other hand (e.g. `void Iface.Hello()`) may be a different matter. I don't remember whether we add these as target methods without the caller specifically asking for that interface to be added. Is that what you're talking about?

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: I'm looking for a test where you pass `null` in as your marshalable object. Such that if we have any code that calls `GetType` on it, it would thrown NRE if we didn't have an appropriate null check.

AArnott (changes requested):
Finished reviewing.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
matteo-prosperi: When I wrote "it doesn't have a name", I explained myself really badly.
I should have written

> we cannot dispatch the call to that method because it is neither declared nor inherited by one of the Rpc interfaces: methods from base interfaces are usually inherited by the Rpc interface but, in this case, the method is hidden.

I think I have clearly understandable examples for this issue that I added to the docs (it's much easier to reason on this with an example). We can have a call and go through this if you want though.

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -143,12 +146,62 @@ internal static bool TryGetMarshalOptionsForType(Type type, [NotNullWhen(true)]
...
+            if (attributes.Select(a => a.SubType).Distinct().Count() != attributes.Count())
+            {
+                throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, Resources.RpcMarshalableDuplicatedSubType, declaredType.FullName));
+            }
+
+            if (attributes.Any(a => a.SubType.GetCustomAttribute<RpcMarshalableAttribute>() is null))
+            {
+                throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, Resources.RpcMarshalableSubTypesMustBeMarshalable, declaredType.FullName));
+            }
+
```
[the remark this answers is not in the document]
matteo-prosperi: Removed the unnecessary LINQ. I only left two `.Distinct().Count()`

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -157,20 +210,43 @@ internal MarshalToken GetToken(IRpcMarshaledContext<object> marshaledContext)
...
+                subTypeCodes ??= new();
+                subTypeCodes.Add(attribute.SubTypeCode);
+
+                this.jsonRpc.AddRpcInterfaceToTargetInternal(
+                    attribute.SubType,
+                    context.Proxy,
+                    new JsonRpcTargetOptions(context.JsonRpcTargetOptions)
+                    {
+                        NotifyClientOfEvents = false, // We don't support this yet.
+                        MethodNameTransform = mn => Invariant($"$/invokeProxy/{handle}/{attribute.SubTypeCode}.{context.JsonRpcTargetOptions.MethodNameTransform?.Invoke(mn) ?? mn}"),
```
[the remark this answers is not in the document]
matteo-prosperi: This is tested differently:
1. on the server side, method ordering is not important because we don't have name collisions: we use [TypeInfo.DeclaredMethods](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.typeinfo.declaredmethods) to retrieve the list of methods for each interface. `DeclaredMethods` doesn't return hidden methods (not clearly documented, but I tested it to be sure). Methods that have the same name across different known interfaces are registered as different RPC names with the `n.foo` pattern.
2. on the client side, the order is important. That is tested by making sure that various combinations of interface implementations result in a dispatch to the right method.

Do you want something else?

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -266,6 +239,27 @@ internal bool TryGetTargetMethod(JsonRpcRequest request, [NotNullWhen(true)] out
...
+    /// <param name="options">A set of customizations for how the target object is registered. If <c>null</c>, default options will be used.</param>
+    /// <param name="revertAddLocalRpcTarget">An optional object that may be disposed of to revert the addition of the target object. This object is returned by an earlier call to <see cref="AddRpcInterfaceToTarget(Type, object, JsonRpcTargetOptions?, IDisposable?)"/>.</param>
+    internal void AddRpcInterfaceToTarget(Type exposingMembersOn, object target, JsonRpcTargetOptions? options, IDisposable? revertAddLocalRpcTarget)
+    {
+        RevertAddLocalRpcTarget? revert = revertAddLocalRpcTarget as RevertAddLocalRpcTarget;
+        if (revertAddLocalRpcTarget is not null && revert is null)
+        {
+            throw new ArgumentException(nameof(revertAddLocalRpcTarget));
+        }
+
```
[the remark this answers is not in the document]
matteo-prosperi: `RevertAddLocalRpcTarget` is a private nested class, so I can't expose it from an internal method. I can make the `RevertAddLocalRpcTarget` `internal` though if we want to avoid this

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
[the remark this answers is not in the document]
AArnott: What I mean is, do we have tests where a custom method name transform is applied to a marshalable object's interface? I think there isn't. Yet we have functionality here that seems to support it, so I'm thinking we ought to test that.

doc/general_marshaled_objects.md
```diff
@@ -87,6 +96,26 @@ The receiver of the above request can invoke `DoSomething` on the marshaled `ISo
...
+    "id": 20,
+    "method": "$/invokeProxy/5/1.DoSomethingElse",
+    "params": []
+}
+```
+
+Note the `1.` prefix added to the leaf method name.
+This allows distinguishing groups of functionality on the object that may otherwise have colliding method names.
+Hosts of marshaled objects MUST support this prefix.
+Hosts MAY also support requests that omit the prefix, but should be prepared to handle any ambiguities that may arise.
```
matteo-prosperi, line 116: @AArnott, I think this is ok because you used "MAY", but I wanted to clarify that we are ONLY registering a `1.DoSomethingElse` RPC method on the server-side, not an unprefixed `DoSomethingElse`.

These are the registered RPC methods on the server-side for this example:
- `DoSomething`
- `1.DoSomething`
- `1.DoSomethingElse`

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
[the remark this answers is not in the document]
matteo-prosperi: You cannot apply a method name transform to a marshaled object's method.

I added a test to validate that a method name transform on the server doesn't trickle down to the marshaled object methods. And also to verify that applying a method name transform on the server object doesn't break the marshaled object functionality.

src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -837,6 +900,56 @@ private static IEnumerable<T> FindAllOnThisAndOtherInterfaces<T>(TypeInfo interf
...
+            int hashCode = this.baseInterfaceType.GetHashCode();
+            if (this.implementedSubTypes is not null)
+            {
+                foreach (int subType in this.implementedSubTypes)
+                {
+                    hashCode = (hashCode * 31) + subType.GetHashCode();
+                }
+            }
+
+            this.hashCode ??= hashCode;
```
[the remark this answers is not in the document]
matteo-prosperi: Good point. I removed the caching, this is internal, so we can make it a class later if we figure out there is a need.

src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -55,38 +55,54 @@ internal static class ProxyGeneration
     /// Gets a dynamically generated type that implements a given interface in terms of a <see cref="JsonRpc"/> instance.
     /// </summary>
     /// <param name="serviceInterface">The interface that describes the RPC contract, and that the client proxy should implement.</param>
+    /// <param name="additionalInterfaces">Additional proxy interfaces that the client proxy should implement. This parameter is used
+    /// only when creating a proxy for a marshalable interface.</param>
```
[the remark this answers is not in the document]
matteo-prosperi: I renamed the method parameters 
- `serverInterface` => `contractInterface`
- `additionalInterfaces` => `implementedOptionalInterfaces`

and cleaned up the parameter docs.

src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -231,169 +247,182 @@ internal static TypeInfo Get(TypeInfo serviceInterface)
...
-                ILGenerator il = methodBuilder.GetILGenerator();
+                    bool hasReturnValue = method.ReturnType.GetTypeInfo().IsGenericType;
+                    Type? invokeResultTypeArgument = hasReturnValue
+                        ? (returnTypeIsIAsyncEnumerable ? method.ReturnType : method.ReturnType.GetTypeInfo().GenericTypeArguments[0])
+                        : null;
 
-                EmitThrowIfDisposed(proxyTypeBuilder, il, disposedField);
+                    string methodName = method.Name;
+                    string rpcMethodName = methodNameMap.GetRpcMethodName(method);
+                    if (rpcInterfaceCode.HasValue)
```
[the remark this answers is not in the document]
matteo-prosperi: Good catch. I meant to do
```
methodName = rpcInterfaceCode + "." + method.Name;
rpcMethodName = rpcInterfaceCode + "." + rpcMethodName;
```

I am also adding a test for a method renamed using `JsonRpcMethodAttribute` (we were missing that test for "normal" marshalable interfaces)

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -117,10 +117,96 @@ public interface IMarshalableWithEvents : IDisposable
...
+    }
+
+    [RpcMarshalable]
+    public interface IMarshalableSubTypesCombined : IMarshalableSubType1Extended, IMarshalableSubType2, IMarshalableNonExtendingBase
+    {
+        Task<int> GetPlusFiveAsync(int value);
+    }
+
+    [RpcMarshalable]
+    [RpcMarshalableKnownSubType(typeof(IMarshalableSubType2Extended), subTypeCode: 99)]
```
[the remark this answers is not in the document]
matteo-prosperi: Done

doc/general_marshaled_objects.md
AArnott: Understood. This protocol doc may describe more (and less) than StreamJsonRpc itself directly supports. Another host (e.g. typescript) will very likely decide to support non-prefixed method names because they don't have interfaces at runtime and far more limited reflection than .NET has. I'm merely calling out this possibility.

AArnott (approved):

PR #845, opened 2022-09-22
Upgrade 2.13 to stable version

javierdlg:


version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "2.13-alpha",
+  "version": "2.13",
```
AArnott, line 3: This will cause a CI build failure because [unstable dependencies exist](https://github.com/microsoft/vs-streamjsonrpc/blob/bd27af10e104685be469c799b6536808afb24d50/Directory.Packages.props#L24).

AArnott (changes requested):
This will cause a CI build failure because [unstable dependencies exist](https://github.com/microsoft/vs-streamjsonrpc/blob/bd27af10e104685be469c799b6536808afb24d50/Directory.Packages.props#L24).

javierdlg:
Fixing unstable dependencies...
https://github.com/AArnott/Nerdbank.Streams/pull/537

AArnott:
The pipeline failure appears to be because nerdbank.streams cannot be ingested by the pipeline. This is something that #846 should be able to fix. Once that's merged, we can re-queue this PR validation.

PR #882, opened 2023-01-24
Allow users to pass in an ActivitySource to be used for generating a new Activity

CherechesC:
If there is no listener attached to the activity source, the activity will not be created, so fallback to create the Activity using the constructor

### DIFF
--- a/src/StreamJsonRpc/ActivityTracingStrategy.cs
+++ b/src/StreamJsonRpc/ActivityTracingStrategy.cs
@@ -12,7 +12,7 @@ namespace StreamJsonRpc;
 /// <seealso cref="CorrelationManagerTracingStrategy"/>
 public class ActivityTracingStrategy : IActivityTracingStrategy
 {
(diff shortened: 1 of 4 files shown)


src/StreamJsonRpc/ActivityTracingStrategy.cs
```diff
@@ -40,6 +59,22 @@ public void ApplyOutboundActivity(JsonRpcRequest request)
...
+            activity = this.activitySource.StartActivity(name);
+        }
+
+        if (activity is null)
+        {
+            activity = new Activity(name);
+        }
+
+        return activity;
+    }
```
AArnott, line 76: nit: this can all be one line.

```suggestion
    private Activity CreateNewActivity(string name) => this.activitySource?.StartActivity(name) ?? new Activity(name);
```

AArnott (changes requested):
Looks great. Just one syntax improvement please.

AArnott (approved):

PR #886, opened 2023-03-01
Add JoinableTaskFactory integration

AArnott:
This avoids deadlocks that RPC would otherwise introduce by carrying a `JoinableTask` token across RPC so that when/if the RPC call ever makes it back to the original AppDomain (or never leaves it) such that the callee needs the main thread, it will be able to reach it if the caller owns the main thread.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -5,6 +5,7 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
 
     <MicroBuildVersion>2.0.107</MicroBuildVersion>
+    <VisualStudioThreadingVersion>17.6.29-alpha</VisualStudioThreadingVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.13.2" />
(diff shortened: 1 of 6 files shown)


Directory.Packages.props
```diff
@@ -5,6 +5,7 @@
     <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
 
     <MicroBuildVersion>2.0.107</MicroBuildVersion>
+    <VisualStudioThreadingVersion>17.6.35-alpha-gd646d3a5e1</VisualStudioThreadingVersion>
```
AArnott, line 8: This will be replaced with an official build when https://github.com/microsoft/vs-threading/pull/1159 merges.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -2044,7 +2072,22 @@ private async ValueTask<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcRequ
...
 
-                return await this.DispatchRequestAsync(request, targetMethod, cancellationToken).ConfigureAwait(false);
+                if (this.JoinableTaskFactory is null || request.TryGetTopLevelProperty<string>(JoinableTaskTokenHeaderName, out string? parentToken) is false || parentToken is null)
+                {
+                    return await this.DispatchRequestAsync(request, targetMethod, cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    JsonRpcMessage? result = null;
+                    await this.JoinableTaskFactory.RunAsync(
```
AArnott, line 2082: This closure pattern suggests adding a `RunAsync<T>` overload would probably be a worthy improvement.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -2044,7 +2072,22 @@ private async ValueTask<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcRequ
                     }
                 }
 
-                return await this.DispatchRequestAsync(request, targetMethod, cancellationToken).ConfigureAwait(false);
+                if (this.JoinableTaskFactory is null || request.TryGetTopLevelProperty<string>(JoinableTaskTokenHeaderName, out string? parentToken) is false || parentToken is null)
```
AArnott, line 2075: We can use our own `AsyncLocal<string?>` to propagate the token even without a JTF instance.

src/StreamJsonRpc/JsonRpc.cs
```diff
@@ -2044,7 +2081,23 @@ private async ValueTask<JsonRpcMessage> DispatchIncomingRequestAsync(JsonRpcRequ
...
                 }
 
-                return await this.DispatchRequestAsync(request, targetMethod, cancellationToken).ConfigureAwait(false);
+                request.TryGetTopLevelProperty<string>(JoinableTaskTokenHeaderName, out string? parentToken);
+                if (this.JoinableTaskFactory is null)
+                {
+                    this.joinableTaskTokenWithoutJtf.Value = parentToken;
+                }
+
+                if (this.JoinableTaskFactory is null || parentToken is null)
```
BertanAygun, line 2090: If there is a JTF instance but no token was provided in the header, should we still use JTF to dispatch for the benefit of the current process? (although I am not sure what benefit there would be)

AArnott: Lifeng and I couldn't think of any benefit in doing so. But there _is_ a non-zero overhead from doing so. Thus, we only do it when we feel it's justified by skipping when there is no parent token.

BertanAygun (approved):

PR #893, opened 2023-03-06
Document JTF integration

AArnott:
#886 introduced JsonRpc-JoinableTaskFactory integration, but did not document it.

### DIFF
--- a/doc/joinableTaskFactory.md
+++ b/doc/joinableTaskFactory.md
@@ -1,0 +1,117 @@
+# Integration with the `JoinableTaskFactory`
+
+StreamJsonRpc can help mitigate deadlocks that may occur when a JSON-RPC client blocks the main thread of an application during an outbound RPC call and the RPC server requires the main thread in order to complete the operation through integrating with the application's `JoinableTaskFactory`.
+
+[Learn more about `JoinableTaskFactory` and how it can mitigate deadlocks for your application](https://aakaka.ms/vsthreading).
+_This_ topic will focus on the JSON-RPC aspects of deadlock mitigation.
+
+## Problem statement
+
+Let's start by examining the problem to be solved.
+
+### An RPC client and server in the same process.
+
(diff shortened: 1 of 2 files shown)


doc/joinableTaskFactory.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Processes without a main thread
+
+Considering the more complex scenario in the above problem statement, there may be an intermediary process in a multi-hop RPC chain that itself doesn't have a main thread and thus no need for a `JoinableTaskFactory`.
+
+StreamJsonRpc makes accommodation for this scenario by automatically propagating the `joinableTaskToken` top-level property it sees in inbound requests to all outbound requests that come from that request.
+In other words, StreamJsonRpc considers that all outbound RPC requests are causally related to the inbound RPC request that caused the outbound request to be made.
+
+This does _not_ mean that any and all requests that are running concurrently with an outbound RPC request will be seen as causally related.
+Only the outbound requests that derive from the .NET execution context that dispatched an incoming request will be assigned that request's `joinableTaskToken`.
```
matteo-prosperi, line 71: I would love a more prescriptive description (something like "don't use `await Task.Run`"), or a short list of the most common practices that would take you off the current execution context.
Also, if I need to do any of the operations that would take you off the current execution context, like `await Task.Run`, is there a way to restore the causal relation (would using `JTF.RunAsync` instead of `Task.Run` help, even if I don't have a main thread myself)?

AArnott: `await Task.Run` wouldn't do it. In fact it's pretty far off the track before you can escape the ExecutionContext. You have to do something like `ThreadPool.UnsafeQueueUserWorkItem`. But this is beyond the scope of JSON-RPC specific concerns. If we needed to document this, it would belong in the vs-threading repo's docs. And so far, it hasn't come up. What _does_ come up are folks who occasionally _want_ to escape the ExecutionContext, and that's usually in connection with a problem with `AsyncLazy<T>` which also doesn't apply here.

PR #907, opened 2023-04-13
Intern strings when using the MessagePackFormatter

AArnott:
This feature is available in MessagePack but not on by default because it has a (small) perf cost when deserializing. But we want it on for RPC by default.

String interning means that we reuse string objects that are equal by value rather than creating new ones. We *don't* use `string.Intern` here because that keeps them in memory forever, and because that requires allocating a string before switching it for the long-lived instance. The way MessagePack (and this PR) does it requires no allocations at all if the string has been interned before. And the interning collection (which comes from MSBuild) is a weak intern, meaning string will not be held in memory for the life of the process.

I also use my `CommonString` type in a few more places. This has the effect of never deserializing even the first string, nor taking the perf hit of an interned string lookup. It memorizes the encoding and decoding between chars and bytes and reuses them each time.

### DIFF
--- a/src/StreamJsonRpc/MessagePackFormatter.cs
+++ b/src/StreamJsonRpc/MessagePackFormatter.cs
@@ -90,6 +90,11 @@ public class MessagePackFormatter : IJsonRpcMessageFormatter, IJsonRpcInstanceCo
         EventArgsFormatter.Instance);
 
     /// <summary>
+    /// The resolver to include when we want to enable string interning.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


BertanAygun (approved):

AArnott:
No. This doesn't impact the wire protocol at all. It's strictly an implementation detail on how we read (and write) the strings.

BertanAygun (approved):

matteo-prosperi (approved):

PR #908, opened 2023-04-21
Add a System.Text.Json formatter

### LINKED ISSUE #305: Add a system.text.json formatter
The [new `System.Text.Json` APIs](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/) are much faster than Newtonsoft.Json, and thus offer a huge perf boost without sacrificing interop as a move to MessagePack has.

Be sure when doing so to implement full streaming support such that we don't buffer an entire message before beginning deserialization.

AArnott:
Closes #305

Remaining tasks:
* [x] Finish fixing tests.
* [x] Add/activate more tests so more exotic types are tested against the new formatter.
* [x] Implement the tracing functionality.
* [x] Add md documentation about this formatter.
* [x] Review TODO comments.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -28,6 +28,7 @@
     <PackageVersion Include="System.IO.Pipelines" Version="7.0.0" />
     <PackageVersion Include="System.IO.Pipes" Version="4.3.0" />
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
+    <PackageVersion Include="System.Text.Json" Version="7.0.2" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="7.0.0" />
     <PackageVersion Include="System.Threading.Tasks.Dataflow" Version="7.0.0" />
     <PackageVersion Include="System.ValueTuple" Version="4.5.0" />
--- a/doc/extensibility.md
+++ b/doc/extensibility.md
@@ -119,10 +119,16 @@ Check out the `BasicJsonRpc` method in our [MessagePackFormatterTests][MessagePa
 When the remote party does not support MessagePack but does support UTF-8 encoded JSON,
 `SystemTextJsonFormatter` offers the most performant choice available.
 
+This formatter is compatible with remote systems that use `JsonMessageFormatter`, provided they use the default UTF-8 encoding.
+The remote party must also use the same message handler, such as `HeaderDelimitedMessageHandler`.
+
 #### When to use `JsonMessageFormatter`
 
 This formatter is the default for legacy reasons, and offers compatibility with data types that can only be serialized with Newtonsoft.Json.
 It produces JSON text and allows configuring the text encoding, with UTF-8 being the default.
+
+This formatter is compatible with remote systems that use `SystemTextJsonFormatter` when using the default UTF-8 encoding.
+The remote party must also use the same message handler, such as `HeaderDelimitedMessageHandler`.
 
 [MessagePackLibrary]: https://github.com/neuecc/MessagePack-CSharp
 [MessagePackUsage]: ../src/StreamJsonRpc.Tests/MessagePackFormatterTests.cs
--- a/doc/extensibility.md
+++ b/doc/extensibility.md
@@ -88,14 +88,24 @@ StreamJsonRpc includes the following `IJsonRpcMessageFormatter` implementations:
     You can contribute your own via `MessagePackFormatter.SetOptions(MessagePackSerializationOptions)`.
     See alternative formatters below.
 
+1. `SystemTextJsonFormatter` - Uses the [`System.Text.Json` library][SystemTextJson] to serialize each
+    JSON-RPC message as UTF-8 encoded JSON. 
+    All RPC method parameters and return types must be serializable by System.Text.Json,
+    with the additional benefit of `DataContract` and `DataMember` attributes being supported by default
+    within StreamJsonRpc where System.Text.Json alone does not support them.
+    You can leverage `JsonConverter<T>` and add your custom converters via attributes or by
+    contributing them to the `SystemTextJsonFormatter.JsonSerializerOptions.Converters` collection.
+
 When authoring a custom `IJsonRpcMessageFormatter` implementation, consider supporting the [exotic types](exotic_types.md) that require formatter participation.
 We have helper classes to make this straightforward.
 Refer to the source code from our built-in formatters to see how to use these helper classes.
 
-### Alternative formatters
+### Choosing your formatter
 
-For performance reasons when both parties can agree, it may be appropriate to switch out the textual JSON
- representation for something that can be serialized faster and/or in a more compact format.
+#### When to use `MessagePackFormatter`
+
+The very best performance comes from using the `MessagePackFormatter` with the `LengthHeaderMessageHandler`.
+This combination is the fastest and produces the most compact serialized format.
 
 The [MessagePack format][MessagePackFormat] is a fast, binary serialization format that resembles the
 structure of JSON. It can be used as a substitute for JSON when both parties agree on the protocol for
@@ -104,7 +114,18 @@ significant wins in terms of performance and payload size.
 Utilizing `MessagePack` for exchanging JSON-RPC messages is incredibly easy.
 Check out the `BasicJsonRpc` method in our [MessagePackFormatterTests][MessagePackUsage] class.
 
+#### When to use `SystemTextJsonFormatter`
+
+When the remote party does not support MessagePack but does support UTF-8 encoded JSON,
+`SystemTextJsonFormatter` offers the most performant choice available.
+
+#### When to use `JsonMessageFormatter`
+
+This formatter is the default for legacy reasons, and offers compatibility with data types that can only be serialized with Newtonsoft.Json.
+It produces JSON text and allows configuring the text encoding, with UTF-8 being the default.
+
 [MessagePackLibrary]: https://github.com/neuecc/MessagePack-CSharp
 [MessagePackUsage]: ../src/StreamJsonRpc.Tests/MessagePackFormatterTests.cs
 [MessagePackFormat]: https://msgpack.org/
+[SystemTextJson]: https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/overview
 [spec]: https://www.jsonrpc.org/specification
--- a/src/StreamJsonRpc/SystemTextJsonFormatter.cs
+++ b/src/StreamJsonRpc/SystemTextJsonFormatter.cs
@@ -2,10 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Buffers;
+using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 using System.Runtime.ExceptionServices;
+using System.Runtime.Serialization;
 using System.Text;
 using System.Text.Json;
 using System.Text.Json.Serialization;
+using System.Text.Json.Serialization.Metadata;
 using StreamJsonRpc.Protocol;
 using StreamJsonRpc.Reflection;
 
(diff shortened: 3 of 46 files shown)


src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -0,0 +1,1047 @@
...
+using System.Text.Json.Serialization.Metadata;
+using StreamJsonRpc.Protocol;
+using StreamJsonRpc.Reflection;
+
+namespace StreamJsonRpc;
+
+/// <summary>
+/// A formatter that emits UTF-8 encoded JSON where user data should be serializable via the <see cref="JsonSerializer"/>.
+/// </summary>
+public partial class SystemTextJsonFormatter : FormatterBase, IJsonRpcMessageFormatter, IJsonRpcMessageTextFormatter, IJsonRpcInstanceContainer, IJsonRpcFormatterState, IJsonRpcMessageFactory
```
BertanAygun, line 24: Is this supposed to be partial? Rest of the class doesn't seem be to be in the PR.

Directory.Packages.props
```diff
@@ -28,6 +28,7 @@
     <PackageVersion Include="System.IO.Pipelines" Version="7.0.0" />
     <PackageVersion Include="System.IO.Pipes" Version="4.3.0" />
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
+    <PackageVersion Include="System.Text.Json" Version="7.0.2" />
```
BertanAygun, line 31: Does both .net fx and .net core compilations remove an assembly from output if it is not used in the main assembly? I remember seeing that behavior before but couldn't be sure.

AArnott: No, I've never heard of that feature. The build cannot be confident that a dll isn't used, even if the compiler doesn't need it. If you wanted to suppress deployment of an assembly, you can add `ExcludeAssets="runtime"` to a PackageReference.
But this doesn't add a package reference. It merely specifies the version to use when some project _does_ reference the package.

AArnott:
@matteo-prosperi and @bertanaygun, as I'm looking to you as reviewers of this rather large PR,

Some functionality that was duplicated across the two pre-existing formatters has been extracted out into a new base class. This should reduce overall code, but makes this PR bigger than it would otherwise be. 
BTW this technically *can* be a binary breaking change, but I believe (as you can see in the `PublicAPI.Shipped.txt` changes in this PR) that the only breaking change would hit folks who explicitly dispose of their own formatters, which I don't expect anyone does. If we do think folks may be doing this, I can mitigate be redeclaring a root-level `Dispose` method on each derived class.

Would you like me to break out the refactoring parts of this PR into its own so that the new formatter PR is smaller?

Directory.Packages.props
BertanAygun: I think I was remembering an issue I encountered for vstest/mstest scenario in VS repo when a referenced but unused assembly wasn't included in test out directory.

In any case, this does get referenced in StreamJsonRpc though so my question was more around for those components using StreamJsonRpc but are not interested in System.Text.Json support. Can they add a package reference to System.Text.Json on their own and exclude assets to save space on their final packaging as long as they dont use SystemTextJsonFormatter.

AArnott: Yes, they can do that.

doc/extensibility.md
```diff
@@ -88,14 +88,24 @@ StreamJsonRpc includes the following `IJsonRpcMessageFormatter` implementations:
     You can contribute your own via `MessagePackFormatter.SetOptions(MessagePackSerializationOptions)`.
     See alternative formatters below.
 
+1. `SystemTextJsonFormatter` - Uses the [`System.Text.Json` library][SystemTextJson] to serialize each
+    JSON-RPC message as UTF-8 encoded JSON. 
+    All RPC method parameters and return types must be serializable by System.Text.Json,
+    with the additional benefit of `DataContract` and `DataMember` attributes being supported by default
```
matteo-prosperi, line 94: This is an interesting choice considering that STJ serializes by default all public properties without requiring DataContract/Member attributes, this means that an object that is already STJ-serializable (STJ would simply ignore the attributes) would be serialized differently through plain STJ compared to StreamJsonRpc+STJ.

src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -0,0 +1,1414 @@
...
+                {
+                    jsonPropertyInfo = jsonTypeInfo.CreateJsonPropertyInfo(propertyType, dataMemberAttribute?.Name ?? memberInfo.Name);
+                    if (dataMemberAttribute is not null)
+                    {
+                        jsonPropertyInfo.Order = dataMemberAttribute.Order;
+                        jsonPropertyInfo.IsRequired = dataMemberAttribute.IsRequired;
+                        if (!dataMemberAttribute.EmitDefaultValue)
+                        {
+                            object? defaultValue = propertyType.IsValueType ? FormatterServices.GetUninitializedObject(propertyType) : null;
+                            jsonPropertyInfo.ShouldSerialize = (_, value) => !object.Equals(defaultValue, value);
```
matteo-prosperi, line 1369: There may not be much we can do as long as we rely on the IJsonTypeInfoResolver for this, but this will result in boxing upon serialization of all value types property with `EmitDefaultValue == false` which is very common for booleans.
We can consider a future optimization where we address this replacing the `IJsonTypeInfoResolver` with a `JsonConverterFactory`

src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -0,0 +1,1414 @@
...
+
+    /// <inheritdoc/>
+    public JsonRpcMessage Deserialize(ReadOnlySequence<byte> contentBuffer, Encoding encoding)
+    {
+        if (encoding is not UTF8Encoding)
+        {
+            throw new NotSupportedException("Only our default encoding is supported.");
+        }
+
+        JsonDocument document = this.deserializingDocument = JsonDocument.Parse(contentBuffer, DocumentOptions);
```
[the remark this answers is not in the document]
AArnott: I do dispose of it. Notice I'm assigning to `this.deserializingDocument`, which is disposed of in any of several other places, whenever deserialization has truly finished.

doc/extensibility.md
AArnott: The trade-off is a good point, and not one I had fully considered. 
This decision was based on getting existing tests to pass, and to ease migration of JsonMessageFormatter users to SystemTextJsonFormatter. I figure if they have DCS attributes applied to their type, they probably mean for them to be considered. I don't know why a type would have them when honoring them would lead to a bad serialization result.

src/StreamJsonRpc/SystemTextJsonFormatter.cs
AArnott: Ooh, that's a good point. I'll brainstorm on this a bit more.

matteo-prosperi: The same applies to `typeInfo.CreateObject = () => FormatterServices.GetUninitializedObject(type);` above BTW

doc/extensibility.md
```diff
@@ -104,7 +114,18 @@ significant wins in terms of performance and payload size.
...
 
+#### When to use `SystemTextJsonFormatter`
+
+When the remote party does not support MessagePack but does support UTF-8 encoded JSON,
+`SystemTextJsonFormatter` offers the most performant choice available.
+
+#### When to use `JsonMessageFormatter`
+
+This formatter is the default for legacy reasons, and offers compatibility with data types that can only be serialized with Newtonsoft.Json.
+It produces JSON text and allows configuring the text encoding, with UTF-8 being the default.
```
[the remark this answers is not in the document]
BertanAygun: It looks like if you are not using Newtonsoft attributes and Newtonsoft converters then STJ vs Newtonsoft should be compatible. It may be worth repeating compatibility with DataContract attributes on when to use  SystemTextJsonFormatter section again in case someone directly jumps through this section.

src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -0,0 +1,1414 @@
...
+            this.formatter.deserializingDocument = null;
+        }
+
+        private static int CountArguments(JsonElement arguments)
+        {
+            int count = 0;
+            switch (arguments.ValueKind)
+            {
+                case JsonValueKind.Array:
+                    foreach (JsonElement element in arguments.EnumerateArray())
```
BertanAygun, line 579: Could this be used here: https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getarraylength?view=net-8.0 to avoid enumerating?

BertanAygun: or JsonElement.Item.Length?

doc/extensibility.md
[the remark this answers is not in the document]
AArnott: Done.

src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -0,0 +1,1414 @@
...
+            private readonly JsonRpcTargetOptions targetOptions;
+
+            public Converter(SystemTextJsonFormatter formatter, JsonRpcProxyOptions proxyOptions, JsonRpcTargetOptions targetOptions)
+            {
+                this.formatter = formatter;
+                this.proxyOptions = proxyOptions;
+                this.targetOptions = targetOptions;
+            }
+
+            public override T? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
```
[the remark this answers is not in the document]
AArnott: Very interesting. Thank you for pointing that out. I see now that the code coverage from all the tests confirm what you say. I'll remove the dead code.

src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -0,0 +1,1414 @@
...
+                    typeInfo = this.fallbackResolver.GetTypeInfo(type, options);
+                }
+
+                this.typeInfoCache.TryAdd(type, typeInfo);
+            }
+
+            return typeInfo;
+        }
+
+        private static void PopulateMembersInfos(Type type, JsonTypeInfo jsonTypeInfo, DataContractAttribute? dataContractAttribute)
```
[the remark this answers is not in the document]
AArnott: when `onlyRecognizeDecoratedTypes == false`.
Granted, I never pass that value in today,

src/StreamJsonRpc/SystemTextJsonFormatter.cs
AArnott: Nothing we can do about `CreateObject`, since it is typed as `Func<object>`

AArnott: In fact, `ShouldSerialize` is also in the same boat: the func signature forces the value to have *already* been boxed. We're not boxing it, System.Text.Json is. We can only hope they have a similar optimization to what newtonsoft recently picked up: recognizing common, small values for value-types and reuse the same boxes.

src/StreamJsonRpc/SystemTextJsonFormatter.cs
AArnott: Yes. Excellent.

matteo-prosperi (approved):

test/StreamJsonRpc.Tests/FormatterTestBase.cs
```diff
@@ -0,0 +1,118 @@
...
+        var actual = (T)this.Formatter.Deserialize(sequence);
+        return actual;
+    }
+
+    [DataContract]
+    public class CustomType
+    {
+        [DataMember]
+        public int Age { get; set; }
+    }
```
BertanAygun, line 117: now that we have custom behavior, should we test IgnoreDataMember usage as well? (Btw is it still the case that newtonsoft doesn't correctly handle it?)

AArnott: IgnoreDataMember is only useful when you _don't_ have a DataContract attribute, since if you _do_ have the DataContract attribute applied, only DataMember members will be serialized anyway.
And since the SystemTextJsonFormatter only has special DCS attribute handling when DataContractAttribute is present, there's nothing really to test for IgnoreDataMember as it should have no effect.

PR #909, opened 2023-04-28
Remove DataContract attribute support from the STJ formatter

AArnott:
It probably will meet more customer expectations (at least in the long term) if the default behavior of our System.Text.Json formatter matches the default behavior of the System.Text.Json serializer.

Also, @matteo-prosperi pointed out we do a lot of boxing in this implementation, which is something we should try to avoid in our official implementation.
Finally, it seems like if folks are looking for DCS compatibility, they may want to match behavior of unattributed types as well, but it turns out that in that mode, our class breaks on the simplest cases. This should probably also be fixed. But in the meantime, I've removed that option from what is now dead code.

We might bring it back, but it should be opt-in instead of opt-out, and it should ideally be more thoroughly tested and perf-optimized to avoid boxing as far as possible.
For now, I'm leaning toward leaving the 'dead code' that provides the DCS compatibility in place till we're more settled on what to do with it.

### DIFF
--- a/src/StreamJsonRpc/SystemTextJsonFormatter.cs
+++ b/src/StreamJsonRpc/SystemTextJsonFormatter.cs
@@ -64,9 +64,6 @@ public class SystemTextJsonFormatter : FormatterBase, IJsonRpcMessageFormatter, 
         {
             // Fields are important because anonymous types are emitted with fields, not properties.
             IncludeFields = true,
-
-            // Provide compatibility with DataContractSerializer attributes by default.
-            TypeInfoResolver = new DataContractResolver(onlyRecognizeDecoratedTypes: true),
         });
     }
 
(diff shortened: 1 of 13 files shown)


src/StreamJsonRpc/SystemTextJsonFormatter.cs
```diff
@@ -1203,34 +1200,55 @@ public object Convert(object value, TypeCode typeCode)
...
+    /// var formatter = new SystemTextJsonFormatter
+    /// {
+    ///     JsonSerializerOptions =
+    ///     {
+    ///         TypeInfoResolver = new SystemTextJsonFormatter.DataContractResolver(),
+    ///     },
+    /// };
+    /// ]]></code>
+    /// </example>
     private class DataContractResolver : IJsonTypeInfoResolver
```
matteo-prosperi, line 1218: Shouldn't this be public for users to be able to use it?

matteo-prosperi (approved):

src/StreamJsonRpc/SystemTextJsonFormatter.cs
AArnott: Per the PR description, I'm ripping it out.

PR #926, opened 2023-06-12
Fix unescaped newline in bad rpc header message

zZeck:
This string had \r\n in it, which just caused a newline to print, leading to a very confusing error message. I believe the intent was to print \r\n literally rather than a newline.

Discovered while working on [a pr for another project](https://github.com/ionide/LanguageServerProtocol/pull/51)


AArnott (approved):
Good call. Thank you.

PR #936, opened 2023-06-20
Fix string id detection for NormalizeId

kroimon:
Currently, `observedTransmittedRequestWithStringId` is not set when sending an instance of `OutboundJsonRpcRequest` because the condition checks for the wrong `JsonRpcRequest` class.

### DIFF
--- a/src/StreamJsonRpc/JsonMessageFormatter.cs
+++ b/src/StreamJsonRpc/JsonMessageFormatter.cs
@@ -303,7 +303,7 @@ public class JsonMessageFormatter : FormatterBase, IJsonRpcAsyncMessageTextForma
 
         try
         {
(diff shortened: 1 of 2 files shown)


kroimon:
I am using the `JsonMessageFormatter` which gives me an `OutboundJsonRpcMessage` which derives from `JsonMessageFormatter.JsonRpcRequestBase`, but not from `JsonMessageFormatter.JsonRpcRequest` which the current code checks for.

AArnott:
Oh dear. I see.

AArnott:
@kroimon It looks like I pushed enough changed lines to the PR that the system is attributing to you that it triggered the CLA requirement. If you can sign it, great. If not, let me know and we'll create a new PR internally since the net changes ironically do not include your original 1-line code change.

kroimon:
@microsoft-github-policy-service agree

AArnott (approved):

PR #982, opened 2023-11-28
docs fix NotifyClientOfEvents option name

kingosticks:
Was "NotifyClientsOfEvents" in docs, should be "NotifyClientOfEvents".


AArnott (approved):
Thanks!

PR #998, opened 2024-02-06
Fix invalid task for 1ES Templates

matteo-prosperi:
Note. This hasn't been tested yet

### DIFF
--- a/azure-pipelines/microbuild.after.yml
+++ b/azure-pipelines/microbuild.after.yml
@@ -26,11 +26,10 @@ steps:
   condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
   continueOnError: true
 
-- task: PublishBuildArtifacts@1
+- task: 1ES.PublishPipelineArtifact@1
   inputs:
-    PathtoPublish: $(Build.ArtifactStagingDirectory)/InsertionOutputs
-    ArtifactName: InsertionOutputs
-    ArtifactType: Container
+    targetPath: $(Build.ArtifactStagingDirectory)/InsertionOutputs
+    artifactName: InsertionOutputs
   displayName: 📢 Publish InsertionOutputs as Azure DevOps artifacts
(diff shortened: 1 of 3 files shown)


azure-pipelines/microbuild.after.yml
```diff
@@ -26,11 +26,10 @@ steps:
...
-- task: PublishBuildArtifacts@1
+- task: 1ES.PublishPipelineArtifact@1
   inputs:
-    PathtoPublish: $(Build.ArtifactStagingDirectory)/InsertionOutputs
-    ArtifactName: InsertionOutputs
-    ArtifactType: Container
+    targetPath: $(Build.ArtifactStagingDirectory)/InsertionOutputs
+    artifactName: InsertionOutputs
   displayName: 📢 Publish InsertionOutputs as Azure DevOps artifacts
   condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
```
AArnott, line 34: As an alternative fix (that is likely to be the form that Library.Template takes when it adds optprof support built-in), [check this out](https://github.com/microsoft/vs-servicehub/blob/6e305cfcad0e911311b510ec9c7f96677770b7e4/azure-pipelines/build.yml#L82-L86).

AArnott (changes requested):

azure-pipelines/microbuild.after.yml
AArnott: Well, never mind on _that_, at least for now. Since this is required for optprof and optprof doesn't work in 1ES PT yet, we need this to be a standard publish task. You can see the latest of this PR for my efforts to get optprof to actually work: https://github.com/microsoft/vs-servicehub/pull/179/files#diff-126a782b536ff59ad80a6b7989298655348d7c9010dc44a9e51827ee840c92c7

AArnott:
I'm fixing this by merging again from Library.Template, and making other fixes.

PR #1008, opened 2024-02-20
Docs: add more detail about marshaled object lifetimes

vivlimmsft:
Adding some extra details to the marshaled object document about lifetimes.

I wrote this based on my understanding of a PR review comment from Andrew, so it might not be totally accurate.


AArnott (approved):
Looks great. Thanks!

PR #1013, opened 2024-03-05
Add test for marshalling one object twice

### LINKED ISSUE #1012: We should document and test the case of sending the same RpcMarshalable object across several top-level RPC methods
Eric Maupin suggested that he might be passing the same RpcMarshalable object across the same RPC connection multiple times. That is, the creator of the real object sends it as a method argument to several methods. 

This pattern concerns me, because _if the methods run concurrently_ I _think_ JsonRpc will see the same object reference and store it only once in our marshalable object handle table, and reuse the same handle each time the object appears. On the receiving side, it will see the same handle each time and use the same proxy. Given the docs require the receiver to dispose of a marshalable object, each of these several methods will presumably dispose the object when they are done. But the _first_ method to do this may end up removing the handle from the marshalable tables kept by both sides, zombifying the other proxies.

If so, this behavior is by design. But we should verify that our docs make abundantly clear that when an RpcMarshalable object is sent over RPC, the sender should truly consider that they don't own the object any more and should certainly _not_ send it again.

Let's review the docs and possibly write a test for this scenario.

AArnott:
Closes #1012

### DIFF
--- a/test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
+++ b/test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
@@ -580,6 +580,31 @@ public abstract class MarshalableProxyTests : TestBase
     }
 
     [Fact]
+    public async Task OneObjectMarshalledTwiceHasIndependentLifetimes()
+    {
+        // The method we call twice returns the same object each time,
+        // but JsonRpc doesn't recognize this and assigns a unique token and proxy each time.
+        IMarshalable proxy1 = (await this.client.GetMarshalableAsync().WithCancellation(this.TimeoutToken))!;
+        IMarshalable proxy2 = (await this.client.GetMarshalableAsync().WithCancellation(this.TimeoutToken))!;
+
+        // Verify that although the original object is the same, the proxies are different.
+        Assert.NotSame(proxy1, proxy2);
+
+        // Verify that the proxies both work.
+        await proxy1.DoSomethingAsync();
+        await proxy2.DoSomethingAsync();
+
+        // Verify that disposing one proxy doesn't break the connection the other proxy has with the original object.
+        // For our purposes, the original object's Dispose isn't so self-destructive that it would break proxy2's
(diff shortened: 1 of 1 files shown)


test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -579,6 +579,24 @@ public async Task IMarshalable_MarshaledBackAndForth()
...
+        IMarshalable proxy1 = (await this.client.GetMarshalableAsync().WithCancellation(this.TimeoutToken))!;
+        IMarshalable proxy2 = (await this.client.GetMarshalableAsync().WithCancellation(this.TimeoutToken))!;
+
+        await proxy1.DoSomethingAsync();
+        await proxy2.DoSomethingAsync();
+
+        Assert.NotSame(proxy1, proxy2);
+        proxy1.Dispose();
+
+        await proxy2.DoSomethingAsync();
```
matteo-prosperi, line 596: Isn't this calling `DoSomethingAsync()` on a disposed `ReturnedMarshalable` (not a disposed proxy) since `proxy1.Dispose()` will cause the disposal of `ReturnedMarshalable`?
I find this line confusing. If my assumption above is correct, I would add a comment before this call clarifying that while `proxy2` is not disposed, the remote object is, so the call may or may not work.

AArnott: The original object's Dispose method doesn't break the object, which is intentional because the point of the test isn't to verify the object, but the RPC connection.
I've added code comments to clarify this.

BertanAygun (approved):

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: I also added a block to ensure the dispose happened end-to-end before proceeding to test the second proxy.

PR #1015, opened 2024-03-12
Add support for call-scoped marshalable objects

### LINKED ISSUE #1009: Add support for call-scoped marshaled objects
Per the protocol: https://github.com/microsoft/vs-streamjsonrpc/blob/main/doc/general_marshaled_objects.md#protocol

Can we replace this 'not supported' scenario with working code?
https://github.com/microsoft/vs-streamjsonrpc/blob/d6f369cd64cf6be25cc4eb965e38a91fa8c8ca03/src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs#L266-L269

AArnott:
Closes #1009

### DIFF
--- a/doc/rpc_marshalable_objects.md
+++ b/doc/rpc_marshalable_objects.md
@@ -7,11 +7,41 @@ StreamJsonRpc allows transmitting marshalable objects (i.e., objects implementin
 
 Marshalable interfaces must:
 
-1. Extend `IDisposable`.
+1. Extend `IDisposable` (unless interface is call-scoped).
 1. Not include any properties.
 1. Not include any events.
 
 The object that implements a marshalable interface may include properties and events as well as other additional members but only the methods defined by the marshalable interface will be available on the proxy, and the data will not be serialized.
+
+The `RpcMarshalableAttribute` must be applied directly to the interface used as the return type, parameter type, or member type within a return type or parameter type's object graph.
+The attribute is not inherited.
+In fact different interfaces in a type hierarchy can have this attribute applied with distinct settings, and only the settings on the attribute applied directly to the interface used will apply.
+
+## Call-scoped vs. explicitly scoped
+
+### Explicit lifetime
+
+An RPC marshalable interface has an explicit lifetime by default.
+This means that the receiver of a marshaled object owns its lifetime, which may extend beyond an individual RPC call.
+Memory for the marshaled object and its proxy are not released until the receiver either disposes of its proxy or the JSON-RPC connection is closed.
+
+### Call-scoped lifetime
+
+A call-scoped interface produces a proxy that is valid only during the RPC call that delivered it.
+It may only be used as part of a method request as or within an argument.
+Using it as or within a return value or exception will result in an error.
+
+This is the preferred model when an interface is expected to only be used within request arguments because it mitigates the risk of a memory leak due to the receiver failing to dispose of the proxy.
+This model also allows the sender to retain control over the lifetime of the marshaled object.
+
+Opt into call-scoped lifetimes by setting the `CallScopedLifetime` property to `true` on the attribute applied to the interface:
+
+```css
+[RpcMarshalable(CallScopedLifetime = true)]
+```
+
+It is not possible to customize the lifetime of an RPC marshaled object except on its own interface.
+For example, applying this attribute to the parameter that uses the interface is not allowed.
 
 ## Use cases
 
@@ -104,6 +134,8 @@ class RpcServer : IRpcServer
 }
 ```
 
+Call-scoped marshalable interfaces may not be used as a return type or member of its object graph.
+
 ### Method argument
 
 In this use case the RPC *client* provides the marshalable object to the server:
@@ -118,6 +150,8 @@ IRpcContract client = jsonRpc.Attach<IRpcContract>();
 var counter = new Counter();
 await client.ProvideCounterAsync(counter);
 ```
+
+Call-scoped marshalable interfaces may only appear as a method parameter or a part of its object graph.
 
 ### Value within a single argument's object graph
 
@@ -144,6 +178,7 @@ await client.ProvideClassAsync(arg);
 ```
 
 ⚠️ While this use case is supported, be very wary of this pattern because it becomes less obvious to the receiver that an `IDisposable` value is tucked into the object tree of an argument somewhere that *must* be disposed to avoid a resource leak.
+This risk can be mitigated by using call-scoped marshalable interfaces.
 
 ### As an argument without a proxy for an RPC interface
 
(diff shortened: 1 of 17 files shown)


test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -93,6 +93,11 @@ public interface IMarshalable : INonMarshalable
     {
     }
 
+    [RpcMarshalable(CallScopedLifetime = true)]
+    public interface IMarshalableWithCallScopedLifetime : IMarshalable
```
matteo-prosperi, line 97: Am I right that, if I had
```cs
    [RpcMarshalable]
    public interface IMarshalableWithNonCallScopedLifetime : IMarshalableWithCallScopedLifetime
```
it would not be call-scoped?

matteo-prosperi (approved):

doc/rpc_marshalable_objects.md
```diff
@@ -7,12 +7,38 @@ StreamJsonRpc allows transmitting marshalable objects (i.e., objects implementin
...
 
 The object that implements a marshalable interface may include properties and events as well as other additional members but only the methods defined by the marshalable interface will be available on the proxy, and the data will not be serialized.
 
+## Call-scoped vs. explicitly scoped
+
+### Explicit lifetime
+
+An RPC marshalable interface has an explicit lifetime by default.
+This means that the receiver of a marshaled object owns its lifetime, which may extend beyond an individual RPC call.
+Memory for the marshaled object and its proxy are not released until the receiver either disposes of its proxy or the JSON-RPC connection is closed.
```
BertanAygun, line 22: It would be good to mention that JsonRpc connection gets closed if original service requested is disposed since that could be a common pitfall.

doc/rpc_marshalable_objects.md
```diff
@@ -7,12 +7,38 @@ StreamJsonRpc allows transmitting marshalable objects (i.e., objects implementin
...
+This is the preferred model when an interface is expected to only be used within request arguments because it mitigates the risk of a memory leak due to the receiver failing to dispose of the proxy.
+This model also allows the sender to retain control over the lifetime of the marshaled object.
+
+Opt into call-scoped lifetimes by setting the `CallScopedLifetime` property to `true` on the attribute applied to the interface:
+
+```css
+[RpcMarshalable(CallScopedLifetime = true)]
+```
+
+It is not possible to customize the lifetime of an RPC marshaled object except on its own interface.
```
BertanAygun, line 39: How does this work with inheritance? Can I override it on an inherited interface if I just use RpcMarshalable attribute on the inheriting interface.

test/StreamJsonRpc.Tests/AsyncEnumerableTests.cs
```diff
@@ -456,6 +456,7 @@ public async Task NotifyAsync_ThrowsIfAsyncEnumerableSent()
     public async Task ArgumentEnumerable_ReleasedOnErrorResponse()
```
BertanAygun, line 456: How does call scoped marshalable objects work with async enumerators? Since the call is technically completed when AsyncEnumerable is returned but I assume in most cases, the call scoped marshaled argument would be used while enumerator is active.

BertanAygun (commented):
Changes look good to me but I had one open question at https://github.com/microsoft/vs-streamjsonrpc/pull/1015#discussion_r1522201076

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: correct.

doc/rpc_marshalable_objects.md
AArnott: The concept of *services* isn't in JSON-RPC at all. That's a brokered service concept which is defined in the vs-servicehub repo. Most proxy-generating APIs in JsonRpc itself do _not_ close the connection. But the `static JsonRpc.Attach<T>` method _does_ return a proxy whose disposal closes the connection, and that is documented on that particular API.

doc/rpc_marshalable_objects.md
AArnott: There is no inheritance of this attribute. Here is what I'm adding to the doc:

> The `RpcMarshalableAttribute` must be applied directly to the interface used as the return type, parameter type, or member type within a return type or parameter type's object graph.
The attribute is not inherited.
In fact different interfaces in a type hierarchy can have this attribute applied with distinct settings, and only the settings on the attribute applied directly to the interface used will apply.

test/StreamJsonRpc.Tests/AsyncEnumerableTests.cs
AArnott: Hmmmmm.... A very good question that's worth documenting, testing, and ... finding the answer to because I don't know. 😲

AArnott: It didn't work as you hoped. But I update it to support it.

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
```diff
@@ -1040,6 +1141,57 @@ public Task AcceptGenericProxyContainerAsync(ProxyContainer<IGenericMarshalable<
...
+            return Task.FromResult<IMarshalableWithCallScopedLifetime>(new MarshalableAndSerializable());
+        }
+
+        public async IAsyncEnumerable<int> CallScopedMarshalableReturnsAsyncEnumerable(IMarshalableWithCallScopedLifetime marshalable)
+        {
+            // Yield before using the marshalable since we want to test that the call-scoped argument
+            // is available for the whole async enumerable state machine.
+            await Task.Yield();
+
+            await marshalable.DoSomethingAsync();
```
BertanAygun, line 1169: Does "await Task.Yield" behave same as "yield return <somevalue>"? 
I was going to suggest doing "yield return .." twice with calling DoSomethingAsync in between for completeness sake.

BertanAygun (approved):

BertanAygun (approved):

test/StreamJsonRpc.Tests/MarshalableProxyTests.cs
AArnott: No, `await Task.Yield();` does not actually produce a value, where `yield return` does. 
The test was effective without yielding two values, and it keeps the unit test method a bit simpler to produce just one.  TBH I don't even know if I needed the yield first, because async enumerables aren't executed at all until they are actually enumerated, which in the case of RPC doesn't happen by default until the receiver starts the foreach.

PR #1033, opened 2024-04-19
Add .NET 6 and 8 targets and drop dependencies

### LINKED ISSUE #1032: Add TFM to enable reduction in explicit dependencies
### Is your feature request related to a problem? Please describe.
I want to minimise dependencies in my project by utilising framework dependencies wherever possible

### Describe the solution you'd like
I want the package to not have a dependencies on packages which can be provided by the framework (net 6/net core 3) ie System.Text.Json

### Describe alternatives you've considered
Accept the additional dependency

### Additional context

_No response_

thompson-tomo:
Add in net 7 & 8 due to additional framework dependencies available hence facilitating dependency reduction

Closes #1032


AArnott (changes requested):
Thanks for authoring the change.
However instead of adding net7 and net8 targets, I think we should just add one net6 target.
Similar to another PR we worked on together, we can roll back the System.Text.Json package dependency version to 6.0.0 for netstandard targets so that they don't require something newer than the net6.0 runtime provides.

nit: Also, please use the canonical `netstandard` capitalization rather than `netStandard` in your msbuild conditional expressions.

thompson-tomo:
@AArnott i have taken a look at this and unfortunately  this project has a requirement of STJ v7, so the option of replacing 7 & 8 with just 8 can't work. So what i have done is dropped the Polyfil packages i could to 6 and made the additional TFM's 6&8 with the difference between 6 & 8 being STJ + System.IO.Pipelines which would be addressed via https://github.com/dotnet/Nerdbank.Streams/issues/743 so would suggest updating that one which i will do & then bumping Nerdbank.Streams here

AArnott:
I'm working through the compiler warnings from this change. I've also published a new Nerdbank.Streams so we can update that.

AArnott (approved):

PR #1037, opened 2024-05-06
Trim and update dependencies

AArnott:
This gets StreamJsonRpc nuget dependencies down to a minimum when targeting .NET 6 and 8.


AArnott:
The build break is due to the holding period for nuget.org package updates in Azure Artifact feeds. It will resolve itself in 6 hours and I'll re-queue the build.

RyanToth3 (approved):

Issue #1038, opened 2024-05-06
SystemTextJsonFormatter throws exceptions when ETW logging is enabled


dibarbet:
I've been working on converting the Roslyn LSP server to STJ in https://github.com/dotnet/roslyn/pull/73207

When validating the changes in VS, I hit a huge number of exception count regressions (thousands in some scenarios), see [this internal link](https://dev.azure.com/devdiv/DevDiv/_git/VS/pullrequest/548623).  These regressions were not reproducible in my changes locally, nor in the build from the VS insertion, until I either
1.  set the env var `StreamJsonRpc_TestWithEventSource`
2.  or ran VS under perfview using the streamjsonrpc ETW provider (F0CFFE53-BCC5-53F5-8873-995568808A3B)

The exception stacks in the RPS run pretty much all look like the following:
![image](https://github.com/microsoft/vs-streamjsonrpc/assets/5749229/0a8b3f06-3149-408d-81d6-459fb6e38d48)

It appears as though we're hitting this throw in `TryGetArgumentByNameOrIndex` - https://github.com/microsoft/vs-streamjsonrpc/blob/f6a3b480bc02f79d73eac256ae3768f51146225a/src/StreamJsonRpc/SystemTextJsonFormatter.cs#L532
because the JsonRpcEventSource is calling to get a string representation of the message with `name: null` and where `JsonArguments` is `JsonValueKind.Object` - https://github.com/microsoft/vs-streamjsonrpc/blob/f6a3b480bc02f79d73eac256ae3768f51146225a/src/StreamJsonRpc/JsonRpcEventSource.cs#L297

PR #1039, opened 2024-05-06
Fix exceptions being thrown when ETW attempts to log messages with System.Text.Json

dibarbet:
I took a stab at fixing https://github.com/microsoft/vs-streamjsonrpc/issues/1038

Before my fix, the test I added fails with the same `System.Text.Json.JsonException : Unexpected value kind for arguments: Object` I'm seeing in the roslyn insertion.

I haven't figured out how to test the change in VS however


AArnott:
Thanks for the investigation and taking a stab at the fix. However, your fix just avoids an exception without solving the underlying problem. Closing in favor of #1040.

PR #1070, opened 2024-08-13
Fix proxies with derived interfaces

AArnott:
### Fix proxy generation for interfaces A, B when B : A

When multiple interfaces are expressly implemented, and one derives from another, it led to over-subscription of the events on the base interface, causing failures at runtime.
The fix is to ensure we only visit each member once, by visiting each *type* once, regardless of the number of times it is referenced directly or indirectly in the list of types to implement.

###  Allow non-unique interfaces to be requested

Now that we can handle interfaces being referenced twice *indirectly*, we might as well not throw when interfaces are referenced twice *directly*.

### Tips for reviewers

Use a diff view that doesn't call out whitespace changes, since some indentation-only changes have been made that clutters up the diff.

### DIFF
--- a/src/StreamJsonRpc/ProxyGeneration.cs
+++ b/src/StreamJsonRpc/ProxyGeneration.cs
@@ -133,66 +133,67 @@ internal static class ProxyGeneration
             FieldBuilder calledMethodField = proxyTypeBuilder.DefineField("calledMethod", typeof(EventHandler<string>), FieldAttributes.Private);
             FieldBuilder marshaledObjectHandleField = proxyTypeBuilder.DefineField("marshaledObjectHandle", typeof(long?), fieldAttributes);
 
-            var ctorActions = new List<Action<ILGenerator>>();
-            foreach (TypeInfo contract in contractInterfaces)
+            if (FindAllOnThisAndOtherInterfaces(contractInterfaces, i => i.DeclaredProperties).FirstOrDefault() is PropertyInfo unsupportedProperty)
             {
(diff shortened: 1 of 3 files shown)


src/StreamJsonRpc/ProxyGeneration.cs
```diff
@@ -133,66 +123,67 @@ internal static TypeInfo Get(Type contractInterface, ReadOnlySpan<Type> addition
             FieldBuilder calledMethodField = proxyTypeBuilder.DefineField("calledMethod", typeof(EventHandler<string>), FieldAttributes.Private);
             FieldBuilder marshaledObjectHandleField = proxyTypeBuilder.DefineField("marshaledObjectHandle", typeof(long?), fieldAttributes);
 
-            var ctorActions = new List<Action<ILGenerator>>();
-            foreach (TypeInfo contract in contractInterfaces)
+            if (FindAllOnThisAndOtherInterfaces(contractInterfaces, i => i.DeclaredProperties).FirstOrDefault() is PropertyInfo unsupportedProperty)
```
RyanToth3, line 126: Is this supposed to be `!FindAllOnThisAndOtherInterfaces` like it was before?

Right it now it looks like its just saying "if there are any properties on any of the contract interfaces then fail", unless I'm misinterpreting it?

AArnott: You interpret it correctly. It's supposed to fail if any interface has a property on it.

And that's what it did before. VerifySupported throws if you pass false to it. And the expression would evaluate to false if any properties were declared.

RyanToth3: Ah got it, I forgot StreamJsonRpc doesn't support properties

RyanToth3 (approved):

Issue #1093, opened 2024-11-13
NuGet vulnerability warnings with Visual Studio 2022 17.12


ContentsMayVary:
Having upgraded to Visual Studio 2022 17.12 today, we are now getting NuGet vulnerability warnings for this package.

**To reproduce:**

Add the following NuGet package reference to a `netstandard2.0` project:
``` 
<PackageReference Include="StreamJsonRpc" Version="2.20.17" />
```

When you compile that project with Visual Studio 2022 17.12 you get the following warnings:

```
warning NU1902: Package 'MessagePack' 2.5.108 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-4qm4-8hg2-g2xm
warning NU1903: Package 'System.Text.Json' 8.0.3 has a known high severity vulnerability, https://github.com/advisories/GHSA-8g4q-xg66-9fp4
warning NU1903: Package 'System.Text.Json' 8.0.3 has a known high severity vulnerability, https://github.com/advisories/GHSA-hh2w-p6rv-4g7w
```

Andrey-2021:
Additionally.
After update Microsoft Visual Studio Community 2022 to version 17.12.0

I can't setup nuget packages.
When I trying setup packege,
I have error:
CACHE https://api.nuget.org/v3-vulnerabilities/2024.11.12.12.02.33/2024.11.14.05.25.25/ vulnerability.update.json
And the NuGet window freezes

![Image](https://github.com/user-attachments/assets/be2a35f2-7c84-4334-822b-1972d9d7fcec)

dotnetjunkie:
This new warning is a problem. Although developers should of course strive to upgrade packages to later, safe packages, this isn't always possible. I find myself, for instance, regularly switching to older branches in order to reproduce and analyze reported bugs. But now I find myself in the situation where those older branches stopped compiling, because we enabled 'warnings as errors' in our projects, while these older branches don't reference the newest NuGet packages, as they didn't exist at that time.

Although the specific NuGet rules 1901, 1902, and 1903 can be excluded in the config file from the 'warnings as errors' list, this is quite some work in a solution with almost 70 projects. Likewise is it quite some work to upgrade the NuGet vulnerable packages in those older branches, and it makes no sense to do that just for the sake of bug analysis.

There might be an easy way to disable this new feature, but unfortunately, the error message isn't clear about this.

[...]

PR #1094, opened 2024-11-18
security: bumps STJ to 8.0.5 due to security advisories

baywet:
partial #1093


RyanToth3 (approved):

RyanToth3:
Thank you for making this change!

PR #1098, opened 2024-12-03
Only use credscan suppressions for Windows build

matteo-prosperi:
This avoids a build error when running credscan for the package publishing, since the suppression file is not available during that job.

### DIFF
--- a/azure-pipelines/official.yml
+++ b/azure-pipelines/official.yml
@@ -71,9 +71,6 @@ extends:
           suppressionFile: $(System.DefaultWorkingDirectory)\azure-pipelines\falsepositives.gdnsuppress
         sbom:
           enabled: true
-        #credscan:
-        #  suppressionsFile: $(Build.SourcesDirectory)/azure-pipelines/CredScanSuppressions.json
-        #  debugMode: true # required for whole directory suppressions
       stages:
       - stage: Build
         variables:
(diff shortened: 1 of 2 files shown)


azure-pipelines/official.yml
```diff
@@ -71,9 +71,6 @@ extends:
           suppressionFile: $(System.DefaultWorkingDirectory)\azure-pipelines\falsepositives.gdnsuppress
```
BertanAygun, line 71: Is the issue that repo is not cloned at all or $(Build.SourcesDirectory) variable is empty? The rest of the tasks also seems to use files from azure-pipelines so I am wondering if the change needed is to replace Build.SourcesDirectory with System.DefaultWorkingDirectory instead?

BertanAygun (changes requested):

azure-pipelines/official.yml
matteo-prosperi: The release jobs don't clone the repo, so `Build.SourceDirectory` is not available. Since we are only using the suppression file to exclude some files generated during the build, we can configure using the suppression file only during the build

BertanAygun: Sounds good but I am curious of the above tasks worked then so it may be worth experimenting later.

BertanAygun (approved):

PR #1100, opened 2024-12-20
Add NerdbankMessagePackFormatter

trippwill:
Add an `IJsonRpcMessageFormatter` implementation that uses [Nerdbank.MessagePack](https://github.com/AArnott/Nerdbank.MessagePack). This serializer has better startup time and is AOT-safe. This alone won't make StreamJsonRpc as a whole AOT safe, but it's a step in that direction.


src/StreamJsonRpc/Protocol/JsonRpcError.cs
```diff
@@ -5,6 +5,7 @@
 using System.Runtime.Serialization;
 using StreamJsonRpc.Reflection;
 using JsonNET = Newtonsoft.Json.Linq;
+using PT = PolyType;
```
AArnott, line 8: I would avoid aliasing namespaces where they don't present type name conflicts or other confusion. The pre-existing aliases here are because two namespaces relate to JSON and present conflicting types (by name or concept), so keeping them distinct is useful. But PolyType doesn't present this problem.

src/StreamJsonRpc/Protocol/JsonRpcMessage.cs
```diff
@@ -14,6 +16,11 @@ namespace StreamJsonRpc.Protocol;
 [KnownType(typeof(JsonRpcRequest))]
 [KnownType(typeof(JsonRpcResult))]
 [KnownType(typeof(JsonRpcError))]
+#pragma warning disable CS0618 //'KnownSubTypeAttribute.KnownSubTypeAttribute(Type)' is obsolete: 'Use the generic version of this attribute instead.'
```
AArnott, line 19: I was afraid that multitargeting libraries would suffer from this. I wonder if I should remove the `Obsolete` attribute and use a custom analyzer instead, with a custom code that you can suppress in a multitargeting library like this.
https://github.com/AArnott/Nerdbank.MessagePack/issues/185

src/StreamJsonRpc/Protocol/JsonRpcRequest.cs
```diff
@@ -61,6 +64,7 @@ public enum ArgumentMatchResult
     /// </value>
     [DataMember(Name = "params", Order = 3, IsRequired = false, EmitDefaultValue = false)]
     [STJ.JsonPropertyName("params"), STJ.JsonPropertyOrder(3), STJ.JsonIgnore(Condition = STJ.JsonIgnoreCondition.WhenWritingNull)]
+    [PT.PropertyShape(Name = "params", Order = 1)]
```
AArnott, line 67: Why is Order = 1 here but 3 for the pre-existing attributes?

src/StreamJsonRpc/NerdbankMessagePackFormatter.ISerializationContextBuilder.cs
```diff
@@ -0,0 +1,241 @@
...
+/// </summary>
+/// <remarks>
+/// The MessagePack implementation used here comes from https://github.com/AArnott/Nerdbank.MessagePack.
+/// </remarks>
+public sealed partial class NerdbankMessagePackFormatter
+{
+    /// <summary>
+    /// Provides a builder interface for configuring the serialization context.
+    /// </summary>
+    public interface IFormatterContextBuilder
```
AArnott, line 27: Considering the set of exotic types grows over time, using an *interface* with a method for each exotic type is a recipe for unpleasant APIs for binary breaking changes in the future because we can't freely add methods to an interface after it ships.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
```diff
@@ -0,0 +1,2301 @@
...
+            SerializeDefaultValues = false,
+        };
+
+        serializer.RegisterConverter(new RequestIdConverter());
+        serializer.RegisterConverter(new JsonRpcMessageConverter(this));
+        serializer.RegisterConverter(new JsonRpcRequestConverter(this));
+        serializer.RegisterConverter(new JsonRpcResultConverter(this));
+        serializer.RegisterConverter(new JsonRpcErrorConverter(this));
+        serializer.RegisterConverter(new JsonRpcErrorDetailConverter(this));
+        serializer.RegisterConverter(new TraceParentConverter());
```
AArnott, line 132: This requirement for state within each converter that is associated with each connection is going to cause extra startup cost for each connection/brokered service because the 'plan' for serialization will have to be recreated each time that the set of converters change.
I'm curious whether the `MessagePackSerializationContext` can help here, to provide some kind of reference to state associated with the connection that you control so that you don't have to recreate the converters each time.
To make a difference though, we'll need to not only be able to reuse converters but reuse the `MessagePackSerializer` object itself across connections. Do you think we could achieve that?

Tracking with https://github.com/AArnott/Nerdbank.MessagePack/issues/186

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
```diff
@@ -0,0 +1,2301 @@
...
+        /// </remarks>
+        ReadOnlySequence<byte> OriginalMessagePack { get; }
+    }
+
+    /// <inheritdoc cref="FormatterBase.MultiplexingStream"/>
+    public new MultiplexingStream? MultiplexingStream
+    {
+        get => base.MultiplexingStream;
+        set => base.MultiplexingStream = value;
+    }
```
AArnott, line 172: This leaves me wondering why it's here.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
```diff
@@ -0,0 +1,2301 @@
...
+            public override JsonObject? GetJsonSchema(JsonSchemaContext context, ITypeShape typeShape)
+            {
+                return CreateUndocumentedSchema(typeof(ProgressClientConverter<TClass>));
+            }
+        }
+
+        /// <summary>
+        /// Converts a progress token to an <see cref="IProgress{T}"/> or an <see cref="IProgress{T}"/> into a token.
+        /// </summary>
+        private class PreciseTypeConverter<TClass> : MessagePackConverter<TClass>
```
AArnott, line 683: Should "Precise" be "Progress"?

AArnott (commented):
Thanks for doing all this work. You're providing invaluable feedback for the Nerdbank.MessagePack library's feature set as well.

src/StreamJsonRpc/NerdbankMessagePackFormatter.ISerializationContextBuilder.cs
trippwill: I got the interface vs class breaking change compat backwards in my head.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: I think the envelope serializer should be re-usable across connections. The set of envelope types is the same for all connections I think?

src/StreamJsonRpc/NerdbankMessagePackFormatter.ICompositeTypeShapeProviderBuilder.cs
```diff
@@ -0,0 +1,42 @@
...
+/// </summary>
+/// <remarks>
+/// The MessagePack implementation used here comes from https://github.com/AArnott/Nerdbank.MessagePack.
+/// </remarks>
+public sealed partial class NerdbankMessagePackFormatter
+{
+    /// <summary>
+    /// Provides a builder interface for adding type shape providers.
+    /// </summary>
+    public interface ICompositeTypeShapeProviderBuilder
```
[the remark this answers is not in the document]
trippwill: I simplified this further by removing this interface and merging the methods in the FormatterContectBuilder. But yes, it would be nice to rely on built-in support from PolyType for composing type shape providers.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
AArnott: Yes, that sounds like a good start.
Although the set of types to be serialized is probably larger on the user-defined side, so the startup hit will still be present.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: I really can't see a reason for it. Copied it directly from MessagePackFormatter. Can't seem any harm in removing it.

src/StreamJsonRpc/NerdbankMessagePackFormatter.ISerializationContextBuilder.cs
trippwill: Removed the interface, using the class in the configuration lambda

src/StreamJsonRpc/Protocol/JsonRpcMessage.cs
trippwill: I wasn't sure if I should #ifdef based on framework, or just use the netstandard2.0 API and silence the obsolete warning.

src/StreamJsonRpc/NerdbankMessagePackFormatter.RawMessagePack.cs
```diff
@@ -0,0 +1,88 @@
...
+
+/// <summary>
+/// Serializes JSON-RPC messages using MessagePack (a fast, compact binary format).
+/// </summary>
+/// <remarks>
+/// The MessagePack implementation used here comes from https://github.com/AArnott/Nerdbank.MessagePack.
+/// </remarks>
+public partial class NerdbankMessagePackFormatter
+{
+    private struct RawMessagePack
```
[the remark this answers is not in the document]
trippwill: Changed over to use the NB.MP one.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
AArnott: The StreamJsonRpc design is that every connection must have a unique instance of an IMessageFormatter. Given that, and that StreamJsonRpc has no insight into how callers may specially configure that formatter before adding it, I think your "FormatterContextBuilder" idea may be useful here. Suppose the StreamJsonRpc user create one FormatterContext which is then reused within each instance of NerdbankMessagePackFormatter. This reused context has a fully-configured MessagePackSerializer within it, so that that instance can be reused across connections.
So all you'd need at that point is a way to tunnel per-connection state into the custom converters you've registered within it (which themselves must then be stateless). Would that work?

trippwill:
Just FYI: Currently all NerdbankMessagePack unit tests fail.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: The per-connection state is the instance of NerdbankMessagePackFormatter?

I can expose the FormatterContextBuilder and FormatterContext as public types, and alter SetFormatterContext so that it takes a FormatterContext instance instead of a configuration lambda.

Would the per-connection state mostly take care of itself, since the FormatterContext would be set on each new instance of NerdbankMessagePackFormatter created by StreamJsonRpc?

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: I copied the name from MessagePackFormatter and I'm not sure what Precise refers to. The other type of converter created by this resolver is ProgressClient.

src/StreamJsonRpc/NerdbankMessagePackFormatter.FormatterContextBuilder.cs
```diff
@@ -0,0 +1,219 @@
...
+/// The MessagePack implementation used here comes from https://github.com/AArnott/Nerdbank.MessagePack.
+/// </remarks>
+public sealed partial class NerdbankMessagePackFormatter
+{
+    /// <summary>
+    /// Provides methods to build a serialization context for the <see cref="NerdbankMessagePackFormatter"/>.
+    /// </summary>
+    public class FormatterContextBuilder
+    {
+        private readonly NerdbankMessagePackFormatter formatter;
```
trippwill, line 26: This is per-connection state. The dependency here needs to be broken to support reusability of the FormatterContext.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: The per-connection state dependency is currently encapsulated in FormatterContextBuilder. We'll need to break that direct dependency, but I'm pretty confident we can do it.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
```diff
@@ -0,0 +1,2109 @@
...
+            MessagePackReader readAhead = reader.CreatePeekReader();
+            int propertyCount = readAhead.ReadMapHeader();
+            for (int i = 0; i < propertyCount; i++)
+            {
+                if (MethodPropertyName.TryRead(ref readAhead))
+                {
+                    return context.GetConverter<Protocol.JsonRpcRequest>(context.TypeShapeProvider).Read(ref reader, context);
+                }
+                else if (ResultPropertyName.TryRead(ref readAhead))
+                {
```
trippwill, line 1114: @AArnott - After adopting MessagePackString, this converter always fails throwing the UnrecognizedJsonRpcMessageException exception at the end of the method.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: I can WriteRaw(MyMPString.MsgPck.Span) and then ReadString() just fine.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
trippwill: I can WriteRaw(MyMPString.MsgPck.Span) and then ReadString() just fine.

src/StreamJsonRpc/Protocol/JsonRpcMessage.cs
AArnott: Using `#if` will help ensure your code is more correct since .NET 8+ APIs are more constrained and will report errors at compile-time that the ns2.0 APIs will not report until runtime. But the choice is yours.

trippwill:
@AArnott : Current test stats:

   2157 Passed
   195 Failed
   13 Skipped

trippwill:
Current test results:
   2182 Passed
   170 Failed
   13 Skipped

@AArnott - Currently seeing a lot of test failures around exceptions.
While adding additional formatter configuration for the tests to pass, I'm seeing some
opportunities to make the builder API a bit nicer if that's the pattern we want to go with.

The need to pre-register converters in the serializer without the ability to resolve and cache at runtime like legacy MessagePack-CSharp resolvers, is currently the cause of the most repeated builder calls. I suppose we could make a converter act similar to a resolver, or make some converters more open to supporting a wider set of types?

trippwill:
Benchmarks:

```

BenchmarkDotNet v0.13.10, Windows 11 (10.0.26100.2605) (Hyper-V)
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.402
  [Host]     : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  Job-GLSONT : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  Job-VLZGJF : .NET Framework 4.8.1 (4.8.9282.0), X64 RyuJIT VectorSize=256


```
| Method             | Runtime              | Formatter           | Mean     | Error    | StdDev   | Ratio | RatioSD | Gen0   | Gen1   | Allocated | Alloc Ratio |
|------------------- |--------------------- |-------------------- |---------:|---------:|---------:|------:|--------:|-------:|-------:|----------:|------------:|
| **InvokeAsync_NoArgs** | **.NET 8.0**             | **JSON**                | **77.92 μs** | **1.471 μs** | **1.913 μs** |  **0.88** |    **0.03** | **0.4883** |      **-** |  **13.09 KB** |        **0.84** |
| InvokeAsync_NoArgs | .NET Framework 4.7.2 | JSON                | 88.86 μs | 0.743 μs | 0.659 μs |  1.00 |    0.00 | 2.4414 | 0.1221 |  15.67 KB |        1.00 |
|                    |                      |                     |          |          |          |       |         |        |        |           |             |
| **InvokeAsync_NoArgs** | **.NET 8.0**             | **MessagePack**         | **46.29 μs** | **0.631 μs** | **0.591 μs** |  **1.09** |    **0.02** | **0.1221** |      **-** |   **4.07 KB** |        **0.63** |
| InvokeAsync_NoArgs | .NET Framework 4.7.2 | MessagePack         | 42.43 μs | 0.611 μs | 0.571 μs |  1.00 |    0.00 | 1.0376 | 0.0610 |   6.43 KB |        1.00 |
|                    |                      |                     |          |          |          |       |         |        |        |           |             |
| **InvokeAsync_NoArgs** | **.NET 8.0**             | **NerdbankMessagePack** | **51.66 μs** | **0.589 μs** | **0.551 μs** |  **0.88** |    **0.01** |      **-** |      **-** |   **4.06 KB** |        **0.63** |
| InvokeAsync_NoArgs | .NET Framework 4.7.2 | NerdbankMessagePack | 58.85 μs | 0.218 μs | 0.170 μs |  1.00 |    0.00 | 1.0376 | 0.0610 |   6.44 KB |        1.00 |

```

BenchmarkDotNet v0.13.10, Windows 11 (10.0.26100.2605) (Hyper-V)
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.402
  [Host]     : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  Job-GLSONT : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  Job-VLZGJF : .NET Framework 4.8.1 (4.8.9282.0), X64 RyuJIT VectorSize=256


```
| Method             | Runtime              | Formatter           | Mean     | Error     | StdDev    | Ratio | Gen0   | Gen1   | Allocated | Alloc Ratio |
|------------------- |--------------------- |-------------------- |---------:|----------:|----------:|------:|-------:|-------:|----------:|------------:|
| **NotifyAsync_NoArgs** | **.NET 8.0**             | **JSON**                | **2.821 μs** | **0.0177 μs** | **0.0157 μs** |  **0.44** | **0.0877** |      **-** |    **2216 B** |        **0.94** |
| NotifyAsync_NoArgs | .NET Framework 4.7.2 | JSON                | 6.347 μs | 0.0327 μs | 0.0306 μs |  1.00 | 0.3738 | 0.0076 |    2359 B |        1.00 |
|                    |                      |                     |          |           |           |       |        |        |           |             |
| **NotifyAsync_NoArgs** | **.NET 8.0**             | **MessagePack**         | **1.877 μs** | **0.0122 μs** | **0.0114 μs** |  **0.63** | **0.0038** |      **-** |     **168 B** |        **0.68** |
| NotifyAsync_NoArgs | .NET Framework 4.7.2 | MessagePack         | 2.961 μs | 0.0345 μs | 0.0306 μs |  1.00 | 0.0381 | 0.0038 |     248 B |        1.00 |
|                    |                      |                     |          |           |           |       |        |        |           |             |
| **NotifyAsync_NoArgs** | **.NET 8.0**             | **NerdbankMessagePack** | **1.891 μs** | **0.0134 μs** | **0.0125 μs** |  **0.56** | **0.0057** |      **-** |     **168 B** |        **0.67** |
| NotifyAsync_NoArgs | .NET Framework 4.7.2 | NerdbankMessagePack | 3.347 μs | 0.0272 μs | 0.0255 μs |  1.00 | 0.0381 | 0.0038 |     250 B |        1.00 |

```

BenchmarkDotNet v0.13.10, Windows 11 (10.0.26100.2605) (Hyper-V)
Intel Xeon Platinum 8370C CPU 2.80GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.402
  [Host]     : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  Job-GLSONT : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  Job-VLZGJF : .NET Framework 4.8.1 (4.8.9282.0), X64 RyuJIT VectorSize=256


```
| Method                        | Runtime              | Formatter           | Mean      | Error     | StdDev    | Ratio | RatioSD | Gen0    | Gen1   | Allocated | Alloc Ratio |
|------------------------------ |--------------------- |-------------------- |----------:|----------:|----------:|------:|--------:|--------:|-------:|----------:|------------:|
| **CreateConnectionAndInvokeOnce** | **.NET 8.0**             | **JSON**                |  **68.88 μs** |  **0.650 μs** |  **0.576 μs** |  **0.39** |    **0.00** |  **2.0000** |      **-** |  **49.85 KB** |        **0.86** |
| CreateConnectionAndInvokeOnce | .NET Framework 4.7.2 | JSON                | 175.96 μs |  2.712 μs |  2.537 μs |  1.00 |    0.00 |  9.3333 |      - |  58.21 KB |        1.00 |
|                               |                      |                     |           |           |           |       |         |         |        |           |             |
| **CreateConnectionAndInvokeOnce** | **.NET 8.0**             | **MessagePack**         |  **49.62 μs** |  **0.977 μs** |  **1.711 μs** |  **0.37** |    **0.01** |  **1.5000** |      **-** |  **36.55 KB** |        **0.83** |
| CreateConnectionAndInvokeOnce | .NET Framework 4.7.2 | MessagePack         | 133.71 μs |  2.591 μs |  2.983 μs |  1.00 |    0.00 |  7.0000 |      - |  44.21 KB |        1.00 |
|                               |                      |                     |           |           |           |       |         |         |        |           |             |
| **CreateConnectionAndInvokeOnce** | **.NET 8.0**             | **NerdbankMessagePack** | **184.48 μs** |  **2.625 μs** |  **2.455 μs** |  **0.40** |    **0.02** |  **2.0000** |      **-** |  **54.96 KB** |        **0.88** |
| CreateConnectionAndInvokeOnce | .NET Framework 4.7.2 | NerdbankMessagePack | 484.65 μs | 14.761 μs | 43.523 μs |  1.00 |    0.00 | 10.0000 | 3.0000 |  62.52 KB |        1.00 |

AArnott:
> Currently seeing a lot of test failures around exceptions and having difficulty tracing the issues. Many errors are caused by an NB.MP throwing b/c an exception is not serializable.

Seeing the exception type, message and stack trace usually helps me understand and suggest fixes for issues like this.
I don't think either msgpack library includes built-in support for exception serialization, so I'm not sure what NB.MP failure would be to blame here.

>  I'm seeing some opportunities to make the builder API a bit nicer if that's the pattern we want to go with.

FYI I haven't looked closely at the core of your pull request yet, since I'm trying to stay 'OOF' :wink: so I can't make a firm recommendation at this point.

> The need to pre-register converters in the serializer without the ability to resolve and cache at runtime like legacy MessagePack-CSharp resolvers, is currently the cause of the most repeated builder calls. I suppose we could make a converter act similar to a resolver, or make some converters more open to supporting a wider set of types?

I recently removed the exception that is thrown when you register more converters after serialization has happened. But it still resets the converter cache so it's a runtime hit that should be avoided if possible.

Is your concern mostly around generics? I'm still working on a better solution for that from NB.MP that I hope will make this simpler for you.

> Unrelated thought: Do we really need two different serializer instances? Maybe it would better to register all the envelope and support type shapes and converters in the default user data configuration, and let the user add additional converters and shapes via the builder. Instead of switching back and forth between an "rpc" configuration and a "userData" configuration.

Historically, separating envelope from user data has been useful in protecting the integrity of the JSON-RPC protocol and our extensions, while allowing the user to have flexibility in how types are serialized. If we have just one serializer object, it seems the user will either be disallowed to customize how types are serialized that we internally must serialize, or the user's customizations could invalidate the objects we serialize for the envelope. And this set of sensitive types may change over time as we change what or how we serialize objects as part of the core protocol or the exotic types. 

If I'm missing the mark of what you're thinking of, please correct me. 
Otherwise, I'm curious what cost (complexity, run time, etc.) you're thinking would benefit from consolidating the serializer instances.

> ... benchmarks...

Most of those look pretty good. The last one seems to mix in making the connection itself, which seems likely to be noisy. But I have no idea why NB.MP would make the process of connecting so much slower than the other libraries. Do you?

trippwill:
> Historically, separating envelope from user data has been useful in protecting the integrity of the JSON-RPC protocol and our extensions, while allowing the user to have flexibility in how types are serialized. 

Totally makes sense.

> Is your concern mostly around generics?

Mostly, yes. With MP-CS resolvers, the right formatter is found by working through an ordered list of resolvers for any given type. This allows for behaviors like a resolver intercepting intrinsic rpc-marhalables before they matched to a more conventional formatter. Same for things like Pipes and Streams as arguments or return types. And IAsyncEnumerable<T> as a property on a class. With NB.MP, a type like `MyClass : IObserver<int>` has to have an RpcMarhsalable converter registered explicitly for it. Currently, I have this done as part of builder. Maybe the better solution would be to expose an `ObserverConverter<T>` from StreamJsonRpc that the user can apply to a custom type like MyClass?

Basically, not being able to intercept the exotic types as they are seen (for best perf) means the StreamJsonRpc consumer needs to be very explicit up-front. 

> But I have no idea why NB.MP would make the process of connecting so much slower than the other libraries. Do you?

Not yet :-)

> FYI I haven't looked closely at the core of your pull request yet, since I'm trying to stay 'OOF' 😉

Totally understandable. I should probably step away from this myself for awhile.

trippwill:
Latest Test Results:
   2205 Passed
   147 Failed
   13 Skipped

trippwill:
Latest Test Results:
   2270 Passed
   82 Failed
   13 Skipped

trippwill:
Test Results:

   2275 Passed
   77 Failed
   13 Skipped

trippwill:
| Method                        | Runtime              | Formatter           | Mean      | Error    | StdDev    | Ratio | RatioSD | Gen0    | Gen1   | Allocated | Alloc Ratio |
|------------------------------ |--------------------- |-------------------- |----------:|---------:|----------:|------:|--------:|--------:|-------:|----------:|------------:|
| **CreateConnectionAndInvokeOnce** | **.NET 8.0**             | **JSON**                |  **75.15 μs** | **1.493 μs** |  **2.767 μs** |  **0.41** |    **0.02** |  **2.0000** |      **-** |  **49.83 KB** |        **0.86** |
| CreateConnectionAndInvokeOnce | .NET Framework 4.7.2 | JSON                | 182.33 μs | 3.626 μs |  7.805 μs |  1.00 |    0.00 |  9.3333 |      - |  58.28 KB |        1.00 |
|                               |                      |                     |           |          |           |       |         |         |        |           |             |
| **CreateConnectionAndInvokeOnce** | **.NET 8.0**             | **MessagePack**         |  **49.67 μs** | **0.737 μs** |  **1.126 μs** |  **0.38** |    **0.01** |  **1.3333** |      **-** |  **36.58 KB** |        **0.83** |
| CreateConnectionAndInvokeOnce | .NET Framework 4.7.2 | MessagePack         | 130.67 μs | 2.109 μs |  1.973 μs |  1.00 |    0.00 |  7.0000 |      - |  44.25 KB |        1.00 |
|                               |                      |                     |           |          |           |       |         |         |        |           |             |
| **CreateConnectionAndInvokeOnce** | **.NET 8.0**             | **NerdbankMessagePack** | **243.71 μs** | **4.817 μs** |  **5.735 μs** |  **0.50** |    **0.02** |  **2.0000** |      **-** |  **66.71 KB** |        **0.79** |
| CreateConnectionAndInvokeOnce | .NET Framework 4.7.2 | NerdbankMessagePack | 492.03 μs | 9.718 μs | 15.413 μs |  1.00 |    0.00 | 13.0000 | 3.0000 |  84.54 KB |        1.00 |

Latest benchmarks hint that there is a correlation between time and number of serializer runtime register calls. In the latest commit, many more converters and new sub type maps are added in the ctor of NerdbankMessagePackFormatter. The time could also be a result of the calls to converter resolvers.

AArnott:
Does this benchmark include the RegisterConverter calls themselves? If so, we should probably isolate that in a "connection creation" benchmark, leaving a benchmark that measures only the incremental cost of an RPC call.

src/StreamJsonRpc/NerdbankMessagePackFormatter.RequestIdConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+            else
+            {
+                writer.Write(value.String);
+            }
+        }
+
+        public override JsonObject? GetJsonSchema(JsonSchemaContext context, ITypeShape typeShape) => JsonNode.Parse("""
+        {
+            "type": ["string", { "type": "integer", "format": "int64" }]
```
AArnott, line 54: I should probably find a way to make this more discoverable, but this isn't a valid JSON schema.
image

A valid form for this would be:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema",
	"oneOf": [
		{
			"type": "string"
		},
		{ "type": "integer", "format": "int64" }
	]
}
```

You don't need the `$schema` property of course, but including it while you write the schema out in VS Code gives you the validation on your json schema.

In JSON though, integers default to 64-bit I believe, so you could probably get away with just this:

```json
{
	"type": ["string", "integer"]
}
```

src/StreamJsonRpc/Reflection/MessageFormatterRpcMarshaledContextTracker.cs
```diff
@@ -396,7 +397,7 @@ internal MarshalToken GetToken(object marshaledObject, JsonRpcTargetOptions opti
     private static void ValidateMarshalableInterface(Type type, RpcMarshalableAttribute attribute)
     {
         // We only require marshalable interfaces to derive from IDisposable when they are not call-scoped.
-        if (!attribute.CallScopedLifetime && !typeof(IDisposable).IsAssignableFrom(type))
+        if (attribute.CallScopedLifetime && !typeof(IDisposable).IsAssignableFrom(type))
```
AArnott, line 400: This seems incorrect to me. And it seems unrelated to Nerdbank.MessagePack formatter support. Am I missing something?

trippwill: I don't think that was meant to be checked-in

AArnott:
FYI I've been building more on your branch with my own branch name. I'll push a bunch of commits at once soon, since I have only one last test failure. 🎈

AArnott (commented):
I've got all tests passing now. I haven't thoroughly reviewed the root formatter file, but the rest of them look good. And the test coverage looks great. Thank you, @trippwill.

src/StreamJsonRpc/NerdbankMessagePackFormatter.cs
```diff
@@ -0,0 +1,1425 @@
...
+    [return: NotNullIfNotNull(nameof(paramsObject))]
+    private static (IReadOnlyDictionary<string, object?> ArgumentValues, IReadOnlyDictionary<string, Type> ArgumentTypes)? GetParamsObjectDictionary(object? paramsObject)
+    {
+        if (paramsObject is null)
+        {
+            return default;
+        }
+
+        // Look up the argument types dictionary if we saved it before.
+        Type paramsObjectType = paramsObject.GetType();
```
AArnott, line 265: Instead of using reflection, this method should acquire the type shape. And maybe we should just serialize it as a whole instead of breaking it up.

src/StreamJsonRpc/Reflection/ExceptionSerializationHelpers.cs
```diff
@@ -25,7 +25,7 @@ internal static class ExceptionSerializationHelpers
     private static StreamingContext Context => new StreamingContext(StreamingContextStates.Remoting);
 
     internal static T Deserialize<T>(JsonRpc jsonRpc, SerializationInfo info, TraceSource? traceSource)
-        where T : Exception
+    ////where T : Exception
     {
```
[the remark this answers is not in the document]
AArnott: I'm guessing not.

RyanToth3 (approved):

znakeeye:
No nuget package available for this? My project doesn't recognize `NerdbankMessagePackFormatter`. Tried both `2.22.11` and `2.23.32-alpha`.

AArnott:
True. We haven't released this change in a package. In fact, we recently decided to delay shipping this another release, so you can watch #1280 for when we bring it back. Hopefully in about a month.

Issue #1087, opened 2024-10-24
Disposal of marhsallable object followed by disposal of jsonRpc on client leads to InvalidOperationException on server


ni28:
Sometimes I'm getting the following exception on server side on 'await jsonRpc.Completion' when jsonRpc is disposed just after disposal of marshallable object on client side:

      System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
         at System.Collections.Generic.List`1.Enumerator.MoveNext()
         at StreamJsonRpc.Reflection.RpcTargetInfo.DisposeAsync()
         at StreamJsonRpc.JsonRpc.JsonRpcDisconnectedShutdownAsync(JsonRpcDisconnectedEventArgs eventArgs)

I believe the reason is a race condition when accessing the localTargetObjectsToDispose field in the RpcTargetInfo class from different threads. There are other non-concurrent collections in the RpcTargetInfo that probably also need attention.

I created a stress test to reproduce this (attached). It may take a while to get the exception in the server console.

[vs-streamjsonrpc-test.zip](https://github.com/user-attachments/files/17506261/vs-streamjsonrpc-test.zip)

AArnott:
Thank you for your report, investigation and repro! What an awesome bug report.

AArnott:
@matteo-prosperi  I wonder if you could take a first look at this?

drewnoakes:
Should this be closed after #1156?

PR #1156, opened 2025-02-25
Fix disposal of marshalable object proxies

matteo-prosperi:
This addresses https://github.com/microsoft/vs-streamjsonrpc/issues/1087

### DIFF
--- a/src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
+++ b/src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
@@ -45,10 +45,17 @@ internal class RpcTargetInfo : System.IAsyncDisposable
 
     public async ValueTask DisposeAsync()
     {
-        if (this.localTargetObjectsToDispose is object)
+        List<object>? objectsToDispose;
+        lock (this.SyncObject)
+        {
+            objectsToDispose = this.localTargetObjectsToDispose;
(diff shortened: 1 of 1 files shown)


AArnott (approved):
This looks good to me. Is the theory of the bug then that DisposeAsync was being called concurrently?

matteo-prosperi:
@AArnott,
the issue was easily reproducible with the sample provided in https://github.com/microsoft/vs-streamjsonrpc/issues/1087.

The sample has a loop of the following:
```cs
        await using var stream = new NamedPipeClientStream(".", "TestPipe", PipeDirection.InOut, PipeOptions.Asynchronous);
        await stream.ConnectAsync().ConfigureAwait(false);

        using var jsonRpc = JsonRpc.Attach(stream);
        var serverProxy = jsonRpc.Attach<IServer>();

        using var myService = await serverProxy.GetMyService().ConfigureAwait(false);
        await myService.DoSomething().ConfigureAwait(false);
```

Which results in frequent exceptions on the following line, due to the collection having been modified concurrently with the iteration
```cs
foreach (object target in this.localTargetObjectsToDispose)
```

I verified that the issue doesn't repro anymore after my change.

I believe that the cause of the concurrent change of the collection is `RevertAddLocalRpcTarget.Dispose()`.

PR #1187, opened 2025-05-07
Cap the amount of text copied into the stringbuilder in JsonRpcEventSource.Format

ToddGrun:
Speedometer test shows two LOH allocations occurring in this method when opening a large cs file, when the value passed into JsonRpcEventSource.Format is a JsonDocument.

1) The ToString call on the document
2) The StringBuilder.Append call using the ToString result

It would be non-trivial to get rid of the first allocation, however, the second can be removed by just calling the StringBuilder.Append overload that takes in a range from the given string to copy into the StringBuilder.

Here's is the allocation from the speedometer test that this is attempting to improve:

image


AArnott (approved):
Thanks.

I don't anticipate a runtime perf benefit because this code only runs when an ETW trace is collected that specifies the streamjsonrpc ETW provider explicitly. Still, it may make our perf runs a little faster and a little more 'true' to customer perception.

PR #1201, opened 2025-06-13
Remove Dependency on Newtonsoft.Json from JsonRpc

eerhardt:
The call to JsonConvert from JsonRpc.InvokeCoreAsync is unconditionally preserving Newtonsoft.Json in trimmed/aot'd apps.

Inspecting the current code, I am not sure how this error case ever happens. The call to InvokeCoreAsync only returns JsonRpcError or JsonRpcResult objects. The else shouldn't ever be taken.

Instead of serializing the object using Newtonsoft.Json, simply put the object's Type into the exception message to indicate what the unexpected object was.

cc @AArnott

### DIFF
--- a/src/StreamJsonRpc/JsonRpc.cs
+++ b/src/StreamJsonRpc/JsonRpc.cs
@@ -1575,7 +1575,7 @@ public class JsonRpc : IDisposableObservable, IJsonRpcFormatterCallbacks, IJsonR
             }
             else
             {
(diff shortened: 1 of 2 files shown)


AArnott (approved):

AArnott (commented):
/azp run

AArnott (commented):
/azp run

PR #1207, opened 2025-06-18
Refactor code and annotations to fix all warnings in NativeAOTCompatibility.

eerhardt:
There are a couple patterns here:

- Use GetMemberWithSameMetadataDefinitionAs to reflect on generic types/methods in a statically verifiable fashion. This method only exists in .NET 6+.
- Suppress warnings when using GetInterfaces when the Type is annotated as preserve 'All'. https://github.com/dotnet/linker/issues/1731
- When looking for a hard-coded interface like IProgress<T>, any preserved Type that implements the interface won't have the interface trimmed.
- MethodNameMap uses GetInterfaceMap, which may not work in all cases. See https://github.com/dotnet/runtime/issues/89157

cc @AArnott 

FYI - @agocke @MichalStrehovsky - would you mind checking the suppressions I needed to add here? The biggest one is the GetInterfaceMap issue.

### DIFF
--- a/src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
+++ b/src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
@@ -70,8 +70,8 @@ public class MessageFormatterProgressTracker
     /// </summary>
     /// <param name="objectType">The type which may implement <see cref="IProgress{T}"/>.</param>
     /// <returns>The <see cref="IProgress{T}"/> from given <see cref="Type"/> object, or <see langword="null"/>  if no such interface was found in the given <paramref name="objectType" />.</returns>
-    public static Type? FindIProgressOfT([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces)] Type objectType)
-        => TrackerHelpers<IProgress<int>>.FindInterfaceImplementedBy(objectType);
+    public static Type? FindIProgressOfT(Type objectType)
+        => TrackerHelpers.FindIProgressInterfaceImplementedBy(objectType);
 
     /// <summary>
     /// Checks if a given <see cref="Type"/> implements <see cref="IProgress{T}"/>.
@@ -79,21 +79,21 @@ public class MessageFormatterProgressTracker
     /// <param name="objectType">The type which may implement <see cref="IProgress{T}"/>.</param>
     /// <returns>true if given <see cref="Type"/> implements <see cref="IProgress{T}"/>; otherwise, false.</returns>
     [Obsolete($"Use {nameof(CanSerialize)} instead.")]
-    public static bool IsSupportedProgressType([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.Interfaces)] Type objectType) => CanSerialize(objectType);
+    public static bool IsSupportedProgressType(Type objectType) => CanSerialize(objectType);
 
     /// <summary>
     /// Checks if a given <see cref="Type"/> implements <see cref="IProgress{T}"/>.
     /// </summary>
     /// <param name="objectType">The type which may implement <see cref="IProgress{T}"/>.</param>
     /// <returns>true if given <see cref="Type"/> implements <see cref="IProgress{T}"/>; otherwise, false.</returns>
-    public static bool CanSerialize([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.Interfaces)] Type objectType) => TrackerHelpers<IProgress<int>>.CanSerialize(objectType);
+    public static bool CanSerialize(Type objectType) => FindIProgressOfT(objectType) is not null;
 
     /// <summary>
     /// Checks if a given <see cref="Type"/> is a closed generic of <see cref="IProgress{T}"/>.
     /// </summary>
     /// <param name="objectType">The type which may be <see cref="IProgress{T}"/>.</param>
     /// <returns>true if given <see cref="Type"/> is <see cref="IProgress{T}"/>; otherwise, false.</returns>
-    public static bool CanDeserialize(Type objectType) => TrackerHelpers<IProgress<int>>.CanDeserialize(objectType);
+    public static bool CanDeserialize(Type objectType) => TrackerHelpers.IsIProgress(objectType);
 
     /// <summary>
     /// Gets a <see cref="long"/> type token to use as replacement of an <see cref="object"/> implementing <see cref="IProgress{T}"/> in the JSON message.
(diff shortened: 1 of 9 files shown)


src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
```diff
@@ -237,12 +240,19 @@ internal ProgressParamInformation(object progressObject, long token)
         {
             Requires.NotNull(progressObject, nameof(progressObject));
 
-            Type? iProgressOfTType = FindIProgressOfT(progressObject.GetType());
+            [UnconditionalSuppressMessage("Trimming", "IL2072:RequiresUnreferencedCode", Justification = "The 'IProgress<>' Type must exist and so trimmer kept it. In which case it also kept it on any type which implements it.")]
+            static Type? FindIProgress(object progressObject) => FindIProgressOfT(progressObject.GetType());
```
agocke, line 244: Not sure I understand the logic here. How do we know it was kept? And why do we assume that it would be present on a type even if it is kept? And even if it is, how do we know that everything about it hasn't been trimmed such that the only thing left is just a name in metadata?

test/NativeAOTCompatibility/Program.cs
```diff
@@ -24,14 +25,12 @@
 
 // When properly configured, this formatter is safe in Native AOT scenarios for
 // the very limited use case shown in this program.
-#pragma warning disable IL2026 // Members annotated with 'RequiresUnreferencedCodeAttribute' require dynamic access otherwise can break functionality when trimming application code
-#pragma warning disable IL3050 // Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.
+[UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Using the Json source generator.")]
+[UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Using the Json source generator.")]
 IJsonRpcMessageFormatter CreateFormatter() => new SystemTextJsonFormatter()
```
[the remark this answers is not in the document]
eerhardt: I think in order to do this "right" we will need to make some form of SystemTextJsonFormatter (either a whole new class, or a new "trim/aot safe" constructor) that is guaranteed to work consistently if you use the source generator. As it stands, the existing SystemTextJsonFormatter class has methods that will generate IL on the fly, and do other incompatible behaviors.

At least for now the SystemTextJsonFormatter is marked as RUC and RDC, so callers get warnings and can learn about what isn't compatible.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
eerhardt: The method that is called here `FindIProgressOfT`:

https://github.com/microsoft/vs-streamjsonrpc/blob/e07c2c6ec78d56b609a98a5769a1f682d352b5ab/src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs#L73

https://github.com/microsoft/vs-streamjsonrpc/blob/e07c2c6ec78d56b609a98a5769a1f682d352b5ab/src/StreamJsonRpc/Reflection/TrackerHelpers.cs#L30-L34

https://github.com/microsoft/vs-streamjsonrpc/blob/e07c2c6ec78d56b609a98a5769a1f682d352b5ab/src/StreamJsonRpc/Reflection/TrackerHelpers.cs#L23

So the `IProgress<T>` interface can't be trimmed, since `TrackerHelpers<IProgress<int>>` references the type.

I had thought of rewriting `TrackerHelpers` instead to look like what we did in SignalR:

https://github.com/dotnet/aspnetcore/blob/9a6b2eb7ee0eaffcdd3a361b06899b4fe735a4d6/src/SignalR/common/Shared/ReflectionHelper.cs#L50-L71

But I wasn't sure it was worth it. It would remove the need to suppress here since we wouldn't need to use a `DynamicallyAccessedMembers` attribute on the `FindInterfaceImplementedBy`method at all. It may make sense to do this refactoring.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -371,7 +365,18 @@ private static IReadOnlyDictionary<string, List<MethodSignature>> GetRequestMeth
...
+        [UnconditionalSuppressMessage("Trimming", "IL2062:UnrecognizedReflectionPattern", Justification = "exposedMembersOnType is annotated as preserve All members, so any Types returned from GetInterfaces should be preserved as well. See https://github.com/dotnet/linker/issues/1731.")]
+        void ActOnInterfaces([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] Type exposedMembersOnType)
+        {
+            foreach (Type iface in exposedMembersOnType.GetInterfaces())
+            {
+#pragma warning disable IL2072 // The Roslyn analyzer warns a different error code than the ilc compiler.
+                ActOn(iface.GetTypeInfo());
+#pragma warning restore IL2072
+            }
+        }
```
MichalStrehovsky, line 377: Nit: if you rewrite this to use `for (int i = 0; ...)` (to work around a Roslyn trim analyzer limitation around `foreach`), the suppression could be under a NET9_OR_BELOW ifdef and eventually becomes dead code that could be found and cleaned up at some point.

(I really hate suppressions.)

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -488,7 +493,8 @@ void ActOn([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMeth
     /// </summary>
     /// <param name="exposedMembersOnType">Type to reflect over and analyze its events.</param>
     /// <returns>A list of EventInfos found.</returns>
-    [SuppressMessage("Trimming", "IL2075:'this' argument does not satisfy 'DynamicallyAccessedMembersAttribute' in call to target method. The return value of the source method does not have matching annotations.", Justification = "false positive: https://github.com/dotnet/linker/issues/1731")]
+    [UnconditionalSuppressMessage("Trimming", "IL2065:UnrecognizedReflectionPattern", Justification = "false positive: https://github.com/dotnet/linker/issues/1731")]
+    [UnconditionalSuppressMessage("Trimming", "IL2075:UnrecognizedReflectionPattern", Justification = "false positive: https://github.com/dotnet/linker/issues/1731")]
```
MichalStrehovsky, line 497: This could also be under ifdef (same limitation around recognizing foreach applies.)

The one later in this file too.

MichalStrehovsky (commented):
Looks reasonable, I share Andy's concerns in general. I think this is not the first place where we had to suppress warnings about accessing interfaces lists and looking for a generic instantiation of a type we know exists. Maybe another new API candidate so that it can be done without suppressions.

AArnott (approved):
This looks good, pending suggestions for updates.

src/StreamJsonRpc/Reflection/MessageFormatterProgressTracker.cs
eerhardt: I've done this refactoring now. This allowed for a decent amount of RUC attributes to be removed.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
eerhardt: Nice! I didn't realize that was the problem. Fixed.

eerhardt:
Thanks for the feedback. I believe I've addressed it all, if you want to take a relook.

PR #1223, opened 2025-07-07
Add source generation of proxies

AArnott:
This cuts down on startup time and gets us a significant step closer to being NativeAOT ready.

## 🌞 Highlights

Startup performance improves because we don't have to reflect over the RPC interface, Ref.Emit ourselves a type that implements that interface, then JIT compile it. The JIT step was immune to ngen because the type doesn't exist at ngen time. Having a source generated implementation over the RPC interface solves all these problems.

NativeAOT is incapable of Ref.Emit and JIT. It has limited reflection capability. So all these perf improvements also gets us closer to full functionality when run under NativeAOT.

## 📖 Overview

- New attributes
  - `JsonRpcContractAttribute` which users _should_ label their RPC interfaces with. 
  - `JsonRpcProxyMappingAttribute` which is used by the source generator to allow reflection-based discovery of the source generated proxy for a given RPC contract interface.
  - `JsonRpcProxyInterfaceGroupAttribute` which predefines sets of RPC interfaces for which a proxy should be source generated so they can be implemented together.
  - `ExportRpcContractProxiesAttribute` is an assembly-targeted attribute that exposes the generated proxies as public types and optionally forbids (by policy) external assemblies from source generating their own proxies. The forbidding policy will be important for assemblies such as RpcContracts, which reserves the right to add interface members, which would become a binary breaking change in a world of source generated proxies otherwise.
- A new analyzer that reads `[JsonRpcContract]` attributed interfaces and reports diagnostics when they violate our rules for such interfaces (e.g. no properties, methods must use one of a few return types, etc.) as well as uphold the new requirement that such interfaces must be declared with at least `internal` visibility so that the source generator can emit a proxy class that implements it.
  
  This new analyzer also scans `[RpcMarshalable]` interfaces and upholds its variant of the rules.
- A source generator that produces proxy classes that implement those `[JsonRpcContract]` interfaces and the `JsonRpcProxyMappingAttribute` to allow their discovery at runtime. 
  The generated proxy classes are declared as `internal` (or `public`) and discovered via the `[JsonRpcProxyMappingAttribute]` at runtime.
  A project that compiles against the new StreamJsonRpc library may opt-in to C# interceptor behavior that will rewrite (most) calls to `JsonRpc.Attach` to instead call a custom method that directly instantiates the necessary proxy type. This eliminates _all_ reflection and (more importantly) moves the compiled assembly away from the `Attach` methods that are attributed as requiring dynamic and unreferenced code which are incompatible with NativeAOT.
  This interceptor is not on by default because it will cause runtime failures if the requested interfaces have no source generated proxy for them, whereas prior behavior would have generated dynamic proxies for any compatible interface.
  - Named arguments support is done with a `Dictionary<string, object?>` instead of a custom struct. This avoids the reflection required by dynamic proxies.
  - `JsonRpc.Attach` calls are searched for in the consuming project and when the interface type(s) are statically discoverable, we ensure that proxies for those types are source generated (reducing the need for `JsonRpcProxyInterfaceGroupAttribute` to only those non-analyzable combinations).
- A diagnostic suppressor to turn off trim/NativeAOT warnings due to Attach calls that the Interceptor will rewrite anyway.
- Add and update tests so that the source generator and new proxies are tested in a variety of configurations.
- New docs for the analyzers.

### Out of scope

These are important items, but it will take a bit more work and I'm eager to merge this so we can show some incremental progress. 

- Enable use of source generated proxies for RPC marshaled objects.
- Stop declaring empty structs for named arguments of methods that take no parameters.


AArnott:
Actually, I think I'll implement a minimal `IReadOnlyDictionary` in the struct itself (which I should just make a class, because it gets boxed anyway). That will avoid the reflection requirements all around.

src/StreamJsonRpc.Analyzers/GeneratorModels/ProxyModel.cs
```diff
@@ -0,0 +1,242 @@
...
+        {
+            // This proxy has no source to emit. It represents a pre-existing external proxy.
+            return;
+        }
+
+        // TODO: consider declaring the proxy type with equivalent visibility as the interface,
+        //       since a public interface needs a publicly accessible proxy.
+        //       Otherwise Reflection is required to access the type.
+        SourceWriter writer = new();
+        writer.WriteLine("""
```
RyanToth3, line 69: Same question about using `StringBuilder`

src/StreamJsonRpc.Analyzers/GeneratorModels/Container.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            throw new InvalidOperationException($"The {this.FullName} container must be declared as partial.");
+        }
+
+        this.Parent?.VerifyPartiality();
+    }
+
+    private void WriteWithinCore(SourceWriter writer, Action<SourceWriter> writeContent)
+    {
+        writer.WriteLine($"{this.Keywords} {this.Name}");
```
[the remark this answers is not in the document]
RyanToth3: Oh I see, `SourceWriter` is your thing that uses `StringBuilder` under the hood. Never mind!

src/StreamJsonRpc.Analyzers/GeneratorModels/ProxyModel.cs
RyanToth3: Disregard

src/StreamJsonRpc.Analyzers/SourceWriter.cs
```diff
@@ -0,0 +1,191 @@
...
+        this.builder.Append(NewLine);
+    }
+
+    /// <summary>
+    /// Appends a new line with the specified text.
+    /// </summary>
+    /// <param name="text">The text to append.</param>
+    /// <param name="disableIndentation">Append text without preserving the current indentation.</param>
+    public void WriteLine(
+        [StringSyntax("c#-test")] string text,
```
RyanToth3, line 105: @matteo-prosperi We should add this syntax to our Gladstone configuration object unit tests to get colorization there

src/StreamJsonRpc/JsonRpcProxyOptions.cs
```diff
@@ -62,22 +125,64 @@ public Func<string, string> MethodNameTransform
...
+            Verify.Operation(!this.IsFrozen, Resources.CannotMutateFrozenInstance);
+            this.acceptProxyWithExtraInterfaces = value;
+        }
+    }
+
+    /// <summary>
+    /// Gets or sets the implementation of the proxy to use when generating a client proxy.
+    /// </summary>
+    /// <value>The default value is <see cref="ProxyImplementation.PreferSourceGenerated"/>.</value>
+    public ProxyImplementation ProxyStyle
```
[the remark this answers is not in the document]
RyanToth3: Maybe `ProxySource`? I don't have a strong opinion either way though, I don't mind `ProxyStyle`.

test/StreamJsonRpc.Analyzer.Tests/Resources/EmptyInterface/IFoo.g.cs
```diff
@@ -0,0 +1,24 @@
...
+#pragma warning disable CS0436 // prefer local types to imported ones
+
+[global::StreamJsonRpc.Reflection.JsonRpcProxyMappingAttribute(typeof(StreamJsonRpc.Generated.IFoo_Proxy))]
+partial interface IFoo
+{
+}
+
+namespace StreamJsonRpc.Generated
+{
+
```
RyanToth3, line 13: The formatting of these generated classes are a bit upsetting but I'm guessing there's a good reason for it?

RyanToth3 (approved):

src/StreamJsonRpc.Analyzers/GeneratorModels/Container.cs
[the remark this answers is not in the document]
AArnott: The thing to avoid is `SyntaxFactory`.

RyanToth3 (approved):

Issue #1230, opened 2025-07-22
Using a struct in an IAsyncEnumerable parameter with native AOT doesn't work


eerhardt:
Using the following code:

```C#
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT license. See LICENSE file in the project root for full license information.

#pragma warning disable SA1402 // File may only contain a single type
#pragma warning disable SA1649 // File name should match first type name

using System.Diagnostics.CodeAnalysis;
using System.Text.Json;
using System.Text.Json.Serialization;
using Nerdbank.Streams;
using StreamJsonRpc;
using StreamJsonRpc.Reflection;

Console.WriteLine("This test is run by \"dotnet publish -r [RID]-x64\" rather than by executing the program.");

// That said, this "program" can run select scenarios to verify that they work in a Native AOT environment.
// When TUnit fixes https://github.com/thomhurst/TUnit/issues/2458, we can move this part of the program to unit tests.
(Stream clientPipe, Stream serverPipe) = FullDuplexStream.CreatePair();
JsonRpc serverRpc = new JsonRpc(new HeaderDelimitedMessageHandler(serverPipe, CreateFormatter()));
JsonRpc clientRpc = new JsonRpc(new HeaderDelimitedMessageHandler(clientPipe, CreateFormatter()));
serverRpc.AddLocalRpcMethod("Add", new Server().Add);
serverRpc.AddLocalRpcMethod("GetOutputsAsync", new Server().GetOutputsAsync);
serverRpc.StartListening();
clientRpc.StartListening();

int sum = await clientRpc.InvokeAsync<int>(nameof(Server.Add), 2, 5);
Console.WriteLine($"2 + 5 = {sum}");

IAsyncEnumerable<CommandOutput> output = await clientRpc.InvokeAsync<IAsyncEnumerable<CommandOutput>>(nameof(Server.GetOutputsAsync));
await foreach (CommandOutput item in output)
{
    Console.WriteLine(item.Text);
}

// When properly configured, this formatter is safe in Native AOT scenarios for
// the very limited use case shown in this program.
[UnconditionalSuppressMessage("Trimming", "IL2026", Justification = "Using the Json source generator.")]
[UnconditionalSuppressMessage("AOT", "IL3050", Justification = "Using the Json source generator.")]
IJsonRpcMessageFormatter CreateFormatter() => new SystemTextJsonFormatter()
{
[...]

PR #1233, opened 2025-07-23
Reduce trimming demands for RPC target objects

AArnott:
I don't think this nets to any actual win, because the overall target object still has an `All` demand placed on it, but I'm working to weed this down and ultimately offer an alternative API for building this up, so this is just a step.

### DIFF
--- a/src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
+++ b/src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
@@ -94,7 +94,7 @@ internal class RpcTargetInfo : System.IAsyncDisposable
         }
     }
 
-    internal static MethodNameMap GetMethodNameMap([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TypeInfo type)
+    internal static MethodNameMap GetMethodNameMap([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces)] TypeInfo type)
     {
         MethodNameMap? map;
         lock (MethodNameMaps)
@@ -377,7 +377,7 @@ internal class RpcTargetInfo : System.IAsyncDisposable
             }
         }
 
-        void ActOn([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TypeInfo t)
+        void ActOn([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods)] TypeInfo t)
         {
             // As we enumerate methods, skip accessor methods
             foreach (MethodInfo method in t.DeclaredMethods.Where(m => !m.IsSpecialName))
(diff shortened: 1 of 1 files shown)


src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -497,18 +497,15 @@ void ActOn([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] Type
 #if !NET10_0_OR_GREATER
     [UnconditionalSuppressMessage("Trimming", "IL2065:UnrecognizedReflectionPattern", Justification = "false positive: https://github.com/dotnet/linker/issues/1731")]
 #endif
-    private static IReadOnlyList<EventInfo> GetEventInfos([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TypeInfo exposedMembersOnType)
+    private static IReadOnlyList<EventInfo> GetEventInfos([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicEvents | DynamicallyAccessedMemberTypes.NonPublicEvents | DynamicallyAccessedMemberTypes.Interfaces)] TypeInfo exposedMembersOnType)
```
eerhardt, line 500: Can you try building this targeting `net10.0` and publishing the AOT compatibility test against `net10.0`? I'm concerned that this will now produce trim warnings since it is no longer annotated with `.All`.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
```diff
@@ -591,7 +588,7 @@ internal class MethodNameMap
         private readonly Dictionary<MethodInfo, JsonRpcIgnoreAttribute?> ignoreAttributes = new Dictionary<MethodInfo, JsonRpcIgnoreAttribute?>();
 
         [UnconditionalSuppressMessage("Trimming", "IL2111:UnrecognizedReflectionPattern", Justification = "typeInfo is annotated as preserve All members, so any Types returned from ImplementedInterfaces should be preserved as well. See https://github.com/dotnet/linker/issues/1731.")]
-        internal MethodNameMap([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] TypeInfo typeInfo)
+        internal MethodNameMap([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces)] TypeInfo typeInfo)
```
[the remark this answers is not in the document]
AArnott: Oops. I guess All _is_ required per that justification message, and I didn't realize it because the suppression was turning off the warning.

src/StreamJsonRpc/Reflection/RpcTargetInfo.cs
AArnott: You're right about that. .NET 10 also flagged two more IL2060 warnings against JsonMessageFormatter around (wait for it...) the failure you already experienced at runtime #1230. So apparently there's an analyzer hole that .NET 10 fills.

RyanToth3 (approved):

PR #1329, opened 2025-10-21
Fix perf regression in testing for RPC marshalable objects

AArnott:
Some recent refactoring inadvertently caused RPC marshalable object testing to not break out early when the type is not an interface.
In this change, I correct that mistake. I also cache redundant attribute checks for interface types.

As reported in devdiv-2575202


AArnott:
This is being validated for effectiveness with [this validation insertion](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=12624415&view=results).

PR #1344, opened 2025-11-07
Add `PolyTypeJsonFormatter` as an experimental API

AArnott:
This is a formatter intended for use in the vs-servicehub repo to support FrameworkServices so that we can have a full stack brokered service system that is NativeAOT safe.
The formatter is hard to use, and will likely evolve into a fully PolyType-based JSON serializer. In the meantime, I have `[Experimental]` slapped onto it to avoid having to maintain the public API for any user outside of vs-servicehub.


etvorun (approved):
:shipit:

PR #1427, opened 2026-05-12
security: upgrades nerdbank message pack version

### LINKED ISSUE #1426: NerdBank Security Issue?
I noticed warnings in my build that this package relies on Nerdbank.MessagePack 1.0.2 that has a high severity vulnerability.

When you get a chance, can you release an update with current dependencies?

Thanks so much for your work!

baywet:
closes #1415
fixes #1426 
due to this advisory, restores are failing on a bunch of projects. https://github.com/AArnott/Nerdbank.MessagePack/security/advisories/GHSA-2cwq-pwfr-wcw3
This pull request upgrade the dependency in an attempt to release and fix the issue.
However I was not able to test things out locally since https://dev.azure.com/azure-public/vside/_artifacts/feed/msft_consumption_public does not have that specific version, and I couldn't figure out how to get the permissions to do so.
Nuking the nuget config locally results in quite a few unit tests failing. Not sure how to proceed from there?

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -28,7 +28,7 @@
     <PackageVersion Include="Microsoft.VisualStudio.Threading.Only" Version="$(VisualStudioThreadingVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.Threading" Version="$(VisualStudioThreadingVersion)" />
     <PackageVersion Include="Microsoft.VisualStudio.Validation" Version="17.13.22" />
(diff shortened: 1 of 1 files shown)


AArnott:
Thanks for volunteering this. We'll go with #1430 instead of this PR because this PR targets an unreleased version and you probably want a release sooner rather than later.

That said, you can workaround this locally by adding a direct reference to the Nerdbank.MessagePack package yourself, or if you're using CPVM, just add a PackageVersion item to Directory.Packages.props to the same dependency (and turn on transitive pinning).

PR #1445, opened 2026-06-01
Support W3C string serialization of traceparent in msgpack formatters

### LINKED ISSUE #1444: NerdbankMessagePackFormatter should accept plain W3C strings for traceparent/tracestate
## Problem

When a non-.NET client (e.g. TypeScript via `@microsoft/servicehub-framework`) sends a JSON-RPC request over MessagePack with W3C Trace Context fields (`traceparent`/`tracestate`) encoded as **plain strings**, the .NET `NerdbankMessagePackFormatter` fails to deserialize them:

`
Unexpected msgpack code 217 (str 8)
`

This is because `TraceParentConverter.Read` unconditionally calls `reader.ReadArrayHeader()`, expecting the compact binary format:

- `traceparent`: `[version, [traceIdBytes, parentIdBytes, flags]]`
- `tracestate`: `["k1", "v1", "k2", "v2"]`

However, this binary format is an internal optimization that only .NET StreamJsonRpc knows how to produce. Non-.NET runtimes naturally encode the standard W3C string representation (e.g. `"00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"`), which is also what the JSON formatter uses.

## Proposed fix

Make `TraceParentConverter.Read` and `ReadTraceState` tolerant of **both** formats:

1. Peek at the next MessagePack token type
2. If it's an **array** → read the compact binary format (existing behavior, used by .NET↔.NET)
3. If it's a **string** → parse the W3C string directly (the `TraceParent(string)` constructor already does this; `tracestate` is already a plain `key=value,key=value` string)

This would be fully backward compatible for .NET↔.NET communication while enabling interop with non-.NET clients that encode trace context as standard W3C strings.

## Context

The `@microsoft/servicehub-framework` npm package recently added general-purpose message filters (`outgoingMessageFilter`/`incomingMessageFilter`) that allow callers to inject metadata like `traceparent` into JSON-RPC envelopes. This works correctly with the UTF-8/JSON formatter but fails with MessagePack because of the binary format requirement.

The current workaround requires TypeScript consumers to reverse-engineer StreamJsonRpc's internal MessagePack encoding format, which is fragile and couples external consumers to an undocumented internal detail.

### Additionally: expose a public encoding utility

As a secondary improvement, it would be useful to expose a public static utility for encoding/decoding trace context in StreamJsonRpc's MessagePack format, so that callers who want to produce the compact representation can do so without reverse-engineering the internal format. But accepting plain strings would make this optional rather than required.

## References

- [`TraceParentConverter`](https://github.com/microsoft/vs-streamjsonrpc/blob/main/src/StreamJsonRpc/NerdbankMessagePackFormatter.TraceParentConverter.cs) — current binary-only reader
- [`TraceParent(string)`](https://github.com/microsoft/vs-streamjsonrpc/blob/main/src/StreamJsonRpc/Protocol/TraceParent.cs) — existing W3C string parser
- [vs-servicehub message filters PR](https://github.com/microsoft/vs-servicehub/pull/487) — the JS-side change that exposed this interop gap

AArnott:
Closes #1444


drewnoakes (approved):
Thanks for fixing this so quickly Andrew.

LGTM. Could add some tests covering parsing failures.

RyanToth3 (approved):

PR #1485, opened 2026-07-15
Stabilize timeout disconnect test

AArnott:
The timeout/disconnect test relied on a fixed delay, which allowed disposal to race ahead of the outbound timeout on busy CI runners.

Recreate the client with a test-only `JsonRpc` subclass that observes cancellation of the original outbound send token. The test now waits for that deterministic signal before disconnecting, preserving the intended timeout-before-disconnect ordering.

### DIFF
--- a/test/StreamJsonRpc.Tests/JsonRpcTests.cs
+++ b/test/StreamJsonRpc.Tests/JsonRpcTests.cs
@@ -1061,12 +1061,20 @@ public abstract partial class JsonRpcTests : TestBase
     [Fact]
     public async Task InvokeWithCancellationAsync_TimeoutAndDisconnectDuringSend_ThrowsTimeoutException()
     {
+        this.clientRpc.Dispose();
+        this.serverRpc.Dispose();
+        this.ReinitializeRpcWithoutListening(observeClientOutboundCancellation: true);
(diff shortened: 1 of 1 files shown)


test/StreamJsonRpc.Tests/JsonRpcTests.cs
```diff
@@ -1061,36 +1061,30 @@ public async Task InvokeWithCancellationAsync_UsesOutboundRequestTimeout()
     [Fact]
     public async Task InvokeWithCancellationAsync_TimeoutAndDisconnectDuringSend_ThrowsTimeoutException()
     {
+        this.clientRpc.Dispose();
+        this.serverRpc.Dispose();
```
sandyarmstrong, line 1065: Should other calls to the Reinit method have these Dispose calls preceding them?

sandyarmstrong (approved):

test/StreamJsonRpc.Tests/JsonRpcTests.cs
AArnott: What should _probably_ be happening is that `ReinitializeRpcWithoutListening` should itself dispose of those fields before reinitializing them.

